PR #404, opened 2026-04-28
Add 'split-file' to llvm_cache.cmake

jumerckx:
circt depends on split-file, this change should hopefully make it possible to use the wheel to build circt against.


makslevental:
i'm gonna force merge this - wasm bindings are broken but that doesn't matter

makslevental (approved):

PR #496, opened 2026-08-01
[ci] Fix jupyterlite page being wiped by a racing Pages deploy

### LINKED ISSUE #473: ModuleNotFoundError occured while running the Jupyterlite Notebook
I see the following 

Image

makslevental:
Fixes #473.

## What's wrong

`piplite.install('mlir-python-bindings')` fails in the JupyterLite notebook because https://llvm.github.io/eudsl/jupyter/ isn't on the live site at all. Every path under it 404s, so the `import mlir` in the next cell raises `ModuleNotFoundError`. The `%%capture` on the install cell hides the piplite failure, which is why it surfaces as a confusing import error one cell later.

```
$ curl -o /dev/null -w '%{http_code}\n' https://llvm.github.io/eudsl/jupyter/
404
$ curl -o /dev/null -w '%{http_code}\n' https://llvm.github.io/eudsl/console/
404
$ curl -o /dev/null -w '%{http_code}\n' https://llvm.github.io/eudsl/
200
```

## Root cause

Five workflows call `deploy_pip_page.yml`, and two of them can run concurrently on a push to main. On the last push, [run 30673686943](https://github.com/llvm/eudsl/actions/runs/30673686943) built the full 96MB page and deployed it, then a second deploy from `build_test_release_eudsl_python_extras` overwrote it ~30 seconds later with a tree containing **only `index.html`**. Pages serves whatever the last deployment contained, so `/console` and `/jupyter` were deleted from the live site.

Confirmed by downloading both artifacts and diffing: the live `index.html` is byte-identical to the 116KB stub, not the 96MB full page.

```
2f70bb6055505a4ac2e8908ec766569a  live index.html
2f70bb6055505a4ac2e8908ec766569a  stub artifact  (index.html only)
50b12ec9a76101297bde322c351a6817  full artifact  (console/ + jupyter/)
```

The second deploy produced a stub because its `Download latest WASM wheel` step looks for `mlir_python_bindings_ubuntu_wasm32_artifact` in a run that never publishes it:

```
==> (not found) Artifact: mlir_python_bindings_ubuntu_wasm32_artifact
==> Found the following artifacts instead:
	==> (found) Artifact: eudsl_python_extras_sdist_artifact
##[error]no artifacts found
```

Everything downstream then quietly no-op'd, `tar` archived a single file, and `if: (!cancelled())` on the upload/deploy steps let the failed job publish anyway. `(!cancelled())` is **true** for a failed job; only cancellation makes it false. The job reported `success` while its Pages deployment reported `failure`.

That step was also unreachable as written: it keys off `inputs.workflow_call`, but no caller passes any inputs, and `inputs.workflow_caller_run_id` was never declared in this workflow, so it always evaluated to empty.

## The fix

- Resolve both wheels from the pip index this job already builds from the releases, dropping the cross-run artifact dependency and the unused input.
- Add a **Verify page contents** gate so a partial `page/` can never be uploaded. It checks the console and jupyter trees exist and that the piplite index really lists both wheels.
- Replace `if: (!cancelled())` with `if: success()` so a failed build stops before publishing. Also skip publishing on `pull_request`, so the existing PR trigger validates the build without deploying.
- Serialize the nested deploys behind their child chains so two runs of the same commit stop racing for the single `pages` concurrency slot.
- Add `set -euxo pipefail` and quote the wheel globs so failures in these steps actually fail the job.

### DIFF
--- a/.github/workflows/build_llvm.yml
+++ b/.github/workflows/build_llvm.yml
@@ -391,11 +391,14 @@ jobs:
       workflow_call: true
       workflow_caller_run_id: ${{ github.run_id }}
 
+  # build_mlir_python_bindings_wheel.yml and build_test_release_eudsl.yml each end in
+  # their own deploy, so deploying off [build] alone would race them for the single
+  # "pages" concurrency slot. Wait for those chains so this deploy lands last.
   call-deploy-pip-page:
 
     if: (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
 
-    needs: [build]
+    needs: [build, call-build-python-bindings, call-build-eudsl]
 
     permissions:
       contents: read
--- a/.github/workflows/build_mlir_python_bindings_wheel.yml
+++ b/.github/workflows/build_mlir_python_bindings_wheel.yml
@@ -478,11 +478,13 @@ jobs:
       workflow_call: true
       workflow_caller_run_id: ${{ github.run_id }}
 
+  # build_test_release_eudsl_python_extras.yml ends in its own deploy; wait for it so the
+  # two don't race for the single "pages" concurrency slot.
   call-deploy-pip-page:
 
     if: (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
 
-    needs: [release-mlir-python-bindings]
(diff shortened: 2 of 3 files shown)


makslevental:
cc @anutosh491

Issue #502, opened 2026-08-02
[wasm] 20 side modules each duplicate LLVM's cl::opt registry (~35 MB, and an abort on load)


makslevental:
Loading the wasm wheel in Pyodide logs an LLVM abort before continuing:

```
Loading mlir_python_bindings
: CommandLine Error: Option 'dump-thin-cg-sccs' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
Failed to load mlir_python_bindings
The following error occurred while loading mlir_python_bindings:
Aborted(). Build with -sASSERTIONS for more info.
```

Loading then proceeds and everything works -- all 25 modules import and MLIR -> SPIR-V -> WGSL runs -- so this is currently noise rather than breakage. But it is the same defect as #499, at a larger scale, and it is worth fixing before it stops being harmless.

## Cause

LLVM registers command-line options through file-scope static `cl::opt` objects, and `cl::Option::addArgument` aborts on a duplicate name. **20 side modules in the wheel each carry their own copy of that machinery**, so whichever loads second re-registers over the first.

Every extension declares `libMLIRPythonWasmCAPI.so` as a dependency, yet each also statically links LLVM's support code:

```
20 libraries containing "inconsistency in registered CommandLine options", 109.2 MB total

  _mlir                        2,032,006  needs [libMLIRPythonSupport-mlir.so, libnanobind-mlir.so, libMLIRPythonWasmCAPI.so]
  _mlirAsyncPasses             1,819,372  needs [ ...same... ]
  _mlirDialectsAMDGPU          1,838,526  needs [ ...same... ]
  _mlirDialectsGPU             1,844,947  needs [ ...same... ]
  ... 16 more, all ~1.8 MB
  libMLIRPythonWasmCAPI.so    73,541,091
```

The option string itself (`dump-thin-cg-sccs`) appears only in `libMLIRPythonWasmCAPI.so`. The abort fires when a second module re-registers an option the CAPI dylib already owns, and the message prints the name from the existing registry -- the same misleading shape as #499, where the reported option was not in the library at fault.

## Why the size numbers are the tell

The two extensions this repo declares itself are tiny; the 19 that come from upstream MLIR's `declare_mlir_python_extension` calls are ~1.9 MB each:

```
_mlirWasmExecutionEngine        4,912 bytes
_mlirSPIRVToWGSL                5,284 bytes

19 upstream extensions      ~35.3 MB combined
```

[...]

PR #507, opened 2026-08-02
[jupyter] Point the install-cell comment at the loader, not LLVMSupport

makslevental:
Follow-up to #506, which attributed the abort in the install cells to the wheel linking LLVMSupport
into every extension module. That was wrong.

llvm/llvm-project#213509 removed those copies and it did work — measured on the wheel from the #504
bump (`20260803+2450b7e17`):

| | before | after |
|---|---|---|
| libraries defining `cl::Option::addArgument` | 21 | **1** (`libMLIRPythonWasmCAPI.so`) |
| wheel size | 38 MB | **23.5 MB** |
| `_mlirAsyncPasses` | 1.73 MB | 0.00 MB |

And the abort is still printed, so LLVMSupport duplication was never the cause.

The cause is the Pyodide 0.27 loader. `get_dynlibs` returns the `.so` paths in zip namelist order,
unsorted, which puts `libMLIRPythonWasmCAPI.so` at entry 24 of 25 — after all 21 extensions that
name it by bare basename in their `dylink.0` NEEDED list. The first extension loads it under that
basename; when the loop later reaches the aggregate's own entry it passes an **absolute path**, which
misses the basename key in `LDSO.loadedLibsByName` and instantiates the library a second time. Its
`cl::opt` constructors run twice and the second registration aborts. 0.27.7's `loadDynlib` aliases
path → basename *after* loading but never checks basename → path *before*, so the guard is
one-directional.

Fixed upstream by pyodide/pyodide#5610 ("replace `loadDynamicLibrary` with `emscripten_dlopen`"),
which is in Pyodide 0.28.0. We pin 0.27.7. Full analysis in #502.

Comment only — the code in #506 is unchanged and still correct, since the symptom is real regardless
of which layer causes it.

### DIFF
--- a/pages/jupyter/contents/mlir-python-starter.ipynb
+++ b/pages/jupyter/contents/mlir-python-starter.ipynb
@@ -24,7 +24,7 @@
     {
       "id": "37e7b57b-06eb-4249-a88b-dfbb7dba510e",
       "cell_type": "code",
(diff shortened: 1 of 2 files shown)


makslevental:
Superseded by #510, which deletes the code this comment described. #508 bumped Pyodide to 0.29.4, so the abort no longer happens and the whole workaround goes away rather than getting a corrected comment.

PR #508, opened 2026-08-02
[wasm] Bump Pyodide to 0.29.4 so the CAPI dylib loads once

makslevental:
Fixes the abort in #502 at its actual cause.

## Why

Pyodide 0.27's loader keys loaded libraries by path but resolves `dylink.0` dependencies by
basename. `get_dynlibs` returns a wheel's `.so` paths in **zip namelist order**, unsorted, which
puts `libMLIRPythonWasmCAPI.so` at entry 24 of 25 — behind all 21 extensions that name it by bare
basename. The first extension loads the aggregate under the basename key; when the loop reaches the
aggregate's own entry it passes an absolute path, misses, and instantiates it a **second time**.
0.27.7's `loadDynlib` aliases path → basename *after* loading and never checks basename → path
*before*, so the guard is one-directional.

Confirmed in the console on the deployed site — two keys, two different DSOs, two different export
tables:

```js
const L = pyodide._module.LDSO.loadedLibsByName;
const a = L["libMLIRPythonWasmCAPI.so"];
const b = L["/lib/python3.12/site-packages/mlir/_mlir_libs/libMLIRPythonWasmCAPI.so"];
console.log(a === b, a.exports === b.exports);   // false false
```

So the 70 MB aggregate is fetched, compiled, and instantiated twice on every page load, and LLVM's
`cl::opt` constructors run twice:

```
CommandLine Error: Option 'dump-thin-cg-sccs' registered more than once!
```

Fixed upstream by **pyodide/pyodide#5610**, which drops the path/basename bookkeeping in favour of
`emscripten_dlopen`; present from 0.28.0.

### DIFF
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -57,6 +57,29 @@ runs:
         timezoneMacos: "Asia/Singapore"
         timezoneWindows: "Singapore Standard Time"
 
+    # Zip stores naive local time with no timezone, and the step above puts the
+    # runners in Asia/Singapore. So a wheel built at 13:09 UTC records 21:09, and
+    # unzipping it anywhere west of Singapore yields files dated hours in the
+    # future -- which makes ninja re-run cmake until it gives up:
+    #   ninja: error: manifest 'build.ninja' still dirty after 100 tries,
+    #   perhaps system time is not set
+    # That bites anyone unpacking a CI artifact to build against locally.
+    # SOURCE_DATE_EPOCH is honoured by scikit-build-core and by pyodide build's
+    # repack, so setting it here covers both scripts/llvm_wasm and
+    # projects/mlir-python-bindings-wasm.
+    #
+    # Midnight UTC *yesterday*, not today: with SOURCE_DATE_EPOCH set the stamp
+    # is rendered in UTC, which removes the Singapore skew, but a naive stamp is
+    # still read back as local time. Midnight-today would still look like the
+    # future to a consumer in UTC-12 unpacking a build made before noon UTC.
+    # Backing off a day beats any real timezone offset while keeping the
+    # timestamp recent and plausible for everything else that reads this.
+    - name: "Set SOURCE_DATE_EPOCH"
+      shell: bash
+      run: |
+
+        echo "SOURCE_DATE_EPOCH=$(( ($(date -u +%s) / 86400 - 1) * 86400 ))" >> $GITHUB_ENV
+
     - name: "Clean ubuntu"
       if: (inputs.container && startsWith(inputs.runs-on, 'ubuntu'))
       shell: bash
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -250,6 +250,15 @@ runs:
           echo CMAKE_OSX_DEPLOYMENT_TARGET=11.0 >> $GITHUB_ENV
         fi
 
+    # pyodide-build compares the HOST python's major.minor against the cross-build
+    # environment's and refuses a mismatch, so PYODIDE_VERSION below dictates the
+    # host interpreter too -- not inputs.python-version, which defaults to 3.12.
+    - name: Install host Python for the cross-build environment
+      if: inputs.target-arch == 'wasm32'
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.13'
+
     - name: Install pyodide
       if: inputs.target-arch == 'wasm32'
       shell: bash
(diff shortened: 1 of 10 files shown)


.github/actions/setup_base/action.yml
```diff
@@ -259,24 +259,38 @@ runs:
...
-        PYODIDE_VERSION=0.27.7
+        #   scripts/llvm_wasm/Dockerfile
+        #   projects/mlir-python-bindings-wasm/build.sh
+        #
+        # 0.29.4 rather than 0.27.7 for pyodide/pyodide#5610: 0.27's loader keys
+        # loaded libraries by path but resolves dylink dependencies by basename, so
+        # it instantiates libMLIRPythonWasmCAPI.so twice and LLVM's cl::opt
+        # constructors run twice, which aborts (llvm/eudsl#502). 0.29.4 is also the
+        # first release whose micropip (0.11.1) accepts pyemscripten_* tags, which
+        # is what lets deploy_pip_page.yml skip retagging the wheel.
```
makslevental: remove this part of the comment

.github/workflows/build_mlir_python_bindings_wheel.yml
```diff
@@ -132,7 +132,12 @@ jobs:
         run: |
           
           if [[ "${{ matrix.target-arch }}" == "wasm32" ]]; then
-            pip download mlir-wheel --plat pyemscripten_2024_0_wasm32 --no-deps --python-version 3.12 -f https://llvm.github.io/eudsl
+            # Tag and python version come from the cross-build env setup_base
+            # pinned, so this cannot drift from the wheel being built. Note a
+            # standalone run of this workflow needs a released mlir-wheel already
+            # carrying that tag; right after a Pyodide bump only the full
+            # build_llvm.yml chain works, since it builds MLIR in the same run.
```
makslevental: remove this part of the comment

.github/actions/setup_base/action.yml
```diff
@@ -259,24 +259,38 @@ runs:
...
         case "$PY" in
-          3.12*) ;;
-          *) echo "::error::expected python 3.12 from the $PYODIDE_VERSION cross-build env, got $PY"; exit 1 ;;
+          3.13*) ;;
+          *) echo "::error::expected python 3.13 from the $PYODIDE_VERSION cross-build env, got $PY"; exit 1 ;;
         esac
 
+        # Publish the wheel's ABI tag so nothing downstream has to hardcode it.
+        # PYODIDE_PLATFORM_TAG must match the value in deploy_pip_page.yml, which
+        # cannot read it from here (that job does not install the xbuildenv).
```
makslevental: remove this part of the comment

.github/workflows/deploy_pip_page.yml
```diff
@@ -61,28 +66,22 @@ jobs:
...
-      # only learned `pyemscripten_*` in 0.11.1. Under 0.9.0 the piplite index resolves the
-      # package fine and then rejects every wheel in it, so `piplite.install` fails with
-      # "Can't find a pure Python 3 wheel". Download by the tag the release actually
-      # carries, then retag to the one this Pyodide accepts. Drop the retag once the
-      # kernel pin moves past micropip 0.11.1.
+      # No retag needed: pyodide-build tags wasm wheels `pyemscripten_*_wasm32`
+      # (PEP 783), and micropip understands that from 0.11.1, which is what
+      # Pyodide 0.29.4 ships. Under the 0.27.x line (micropip 0.9.0) the piplite
+      # index resolved the package and then rejected every wheel in it, so the
+      # wheel had to be retagged to `pyodide_*` first.
```
makslevental: remove this part of the comment

.github/workflows/deploy_pip_page.yml
```diff
@@ -197,6 +196,9 @@ jobs:
           import pathlib
           import re
 
+          # The ABI the pinned Pyodide accepts. micropip only installs wheels whose
+          # platform tag it recognises, and from 0.11.1 (Pyodide 0.29.4) that includes
+          # the pyemscripten_* tags pyodide-build emits, so no retag is needed.
```
makslevental: remove this part of the comment

.github/actions/setup_base/action.yml
makslevental: Removed.

.github/workflows/build_mlir_python_bindings_wheel.yml
makslevental: Removed.

.github/actions/setup_base/action.yml
makslevental: Removed.

.github/workflows/deploy_pip_page.yml
makslevental: Removed.

.github/workflows/deploy_pip_page.yml
makslevental: Removed.

PR #517, opened 2026-08-05
[eudsl-llvmpy] DSL: module globals, README, and coverage

makslevental:
Finishes the DSL frontend on top of the value, AST, control-flow, and `@function` PRs.

- Module globals with initializers and address spaces.
- A README documenting both the object layer and the DSL, with their limitations.
- Coverage wiring for the DSL plus behavioral-gap tests (`set_successor` and related) that the coverage gate would otherwise leave unexercised.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -79,13 +79,6 @@ void populate_context(nb::module_ &m) {
             return self.get().getFunction(name);
           },
           "name"_a, nb::rv_policy::reference)
-      .def_prop_ro("globals",
-                   [](eudsl::Module &self) {
-                     std::vector<llvm::GlobalVariable *> out;
-                     for (llvm::GlobalVariable &g : self.get().globals())
-                       out.push_back(&g);
-                     return out;
-                   })
       .def(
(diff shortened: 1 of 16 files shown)


projects/eudsl-llvmpy/src/IR/Context.cpp
```diff
@@ -122,7 +123,31 @@ void populate_context(nb::module_ &m) {
...
+          "target_machine"_a)
+      .def(
+          "add_global",
+          [](eudsl::Module &self, llvm::Type *ty, const std::string &name,
+             llvm::Constant *init, bool isConstant,
+             unsigned addressSpace) -> llvm::GlobalVariable * {
+            return new llvm::GlobalVariable(
+                self.get(), ty, isConstant,
+                llvm::GlobalValue::ExternalLinkage, init, name, nullptr,
+                llvm::GlobalValue::NotThreadLocal, addressSpace);
```
makslevental, line 141: don't assume hardcoding for any params - expose whatever `llvm::GlobalValue::ExternalLinkage`, `llvm::GlobalValue::NotThreadLocal`, and `nullptr` correspond to in the ctor and use default args at the python level instead (possibly with `none` handling)

PR #520, opened 2026-08-06
[eudsl-llvmpy] MLIR parity: isinstance over type().__name__ in tests

makslevental:
Convert object-class assertions to isinstance(x, llvm.Foo), matching the
MLIR python bindings' convention (builtin_types.py). The exact-type checks
in the downcast sweeps (test_downcast_kinds.py, the valueTypeInfo sweep in
test_cpp_coverage.py) and dynamic kind filters stay type().__name__, since
they verify the dispatch produces exactly that leaf class.

### DIFF
--- a/projects/eudsl-llvmpy/tests/test_constants.py
+++ b/projects/eudsl-llvmpy/tests/test_constants.py
@@ -31,7 +31,7 @@ _AGGREGATE_SRC = dedent(
 def test_const_int():
     with llvm.Context() as ctx:
         c = llvm.const_int(llvm.types.i32(ctx), 42)
-        assert type(c).__name__ == "ConstantInt"
+        assert isinstance(c, llvm.ConstantInt)
         assert c.value == 42
         assert str(c) == "i32 42"
         neg = llvm.const_int(llvm.types.i32(ctx), -1, signed=True)
@@ -58,10 +58,10 @@ def test_const_int_out_of_range_raises():
 def test_const_bool_and_fp():
     with llvm.Context() as ctx:
         t = llvm.const_bool(ctx, True)
-        assert type(t).__name__ == "ConstantInt"
+        assert isinstance(t, llvm.ConstantInt)
         assert str(t) == "i1 true"
         f = llvm.const_fp(llvm.types.f64(ctx), 1.5)
-        assert type(f).__name__ == "ConstantFP"
+        assert isinstance(f, llvm.ConstantFP)
         assert f.double_value == 1.5
     assert_no_leaks()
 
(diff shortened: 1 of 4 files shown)


makslevental:

This PR converts 7 assertions from `type(x).__name__ == "Foo"` to `isinstance(x, llvm.Foo)` for MLIR parity. Tests only; no production change. The `conftest.py` leak gate is present and active.

### 1. The `BinaryOperator` conversion is strictly weaker; `FPBinaryOperator` now slips through (Important)

`test_values.py:64` and `:77`. `BinaryOperator` is the one converted target with a subclass in the bindings: `FPBinaryOperator` (MRO `FPBinaryOperator → BinaryOperator → Instruction → User → Value`). The old exact-name check pinned the concrete class; `isinstance(x, llvm.BinaryOperator)` also passes for an `FPBinaryOperator`:
```
parse "fadd float %a, %b" -> first inst concrete type: FPBinaryOperator
isinstance(inst, BinaryOperator): True     old "== 'BinaryOperator'" would be: False
BinaryOperator subclasses: ['FPBinaryOperator']
```
The test comment says the intent is "the Value type_hook downcasts the add to its concrete class" — the exact-name form was pinning exactly that. After conversion, a regression where the type_hook mis-downcasts an integer `add` to `FPBinaryOperator` would no longer be caught. Both current tests use an integer `add`, so they pass today; the loss is future regression-catching strength, and it contradicts the assertion's stated purpose. Keep these two as exact-name checks (or `type(x) is llvm.BinaryOperator`).

### Smaller points

- **Confirm the flat `llvm.<ClassName>` names resolve on the branch build.** This PR is the first to reference classes as `llvm.ConstantInt`/`llvm.ConstantFP`/`llvm.GlobalVariable`/`llvm.BinaryOperator`. On the branch's flat `__init__` (`from .eudslllvm_ext import *`) they should resolve like the already-used flat functions, but I couldn't confirm against the installed build (which is the later submodule layout where these live at `llvm.ir.*`). A quick local `pytest` of `test_constants.py`/`test_values.py` on the branch would surface a typo/un-exported class as `AttributeError`.
- **Refactor is incomplete relative to its parity goal.** `test_cpp_coverage.py:296-297` (and `:154`, `:203-204`) still use `type(...).__name__ == "ConstantInt"/"ConstantFP"/"LoadInst"/...` in the same touched file. Not a correctness issue, but the conversion hits only 7 of many identical sites, leaving the suite mixed.

### What's good

- The five conversions to `ConstantInt`, `ConstantFP`, `GlobalVariable` are faithful — each is a leaf with no subclasses (`subclasses: []`), so `isinstance` is equivalent to the exact-name check.
- The PR correctly **left** `UndefValue`/`PoisonValue`/`ConstantPointerNull` as exact-name checks. Converting those would have been a real bug: `issubclass(PoisonValue, UndefValue)` is True, so `isinstance(poison, UndefValue)` passes — the exact-name form is required there. Avoiding that trap is the right call.
- No assertions were dropped or altered beyond the type-check form; surrounding assertions (`.value`, `str(...)`, `.initializer`) are preserved.

PR #524, opened 2026-08-06
[eudsl-llvmpy] MLIR parity: constant/global accessors

makslevental:
Flesh out the bare constant/global registrations:
- ConstantDataSequential: num_elements, get_element_as_int/double, is_string,
  as_string (data is stored inline, so User's operand access gives nothing).
- ConstantExpr.opcode_name; BlockAddress.function/basic_block.
- GlobalAlias.aliasee; GlobalIFunc.resolver.

ConstantArray/ConstantStruct already get element access by inheriting the
User __len__/__getitem__ added in the iteration PR, so they need nothing more.
Note: is_string is wrapped in a lambda because isString has a default CharSize
argument a method pointer can't bind. A fuller Attribute/AttrBuilder surface
remains a candidate follow-up.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Constants.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Constants.cpp
@@ -45,13 +45,28 @@ void populate_constants(nb::module_ &m) {
       m, "ConstantDataSequential")
       .def_prop_ro("num_elements", &llvm::ConstantDataSequential::getNumElements)
       .def("get_element_as_int",
-           [](llvm::ConstantDataSequential &self, unsigned i) {
-             return self.getElementAsInteger(i);
+           [](llvm::ConstantDataSequential &self, Py_ssize_t i) {
+             // getElementAsInteger asserts (aborts the process) on a bad index
+             // or a non-integer element type; surface both as catchable Python
+             // errors instead.
+             Py_ssize_t n = static_cast<Py_ssize_t>(self.getNumElements());
+             if (i < 0 || i >= n)
+               throw nb::index_error("element index out of range");
+             if (!self.getElementType()->isIntegerTy())
+               throw nb::value_error("element type is not an integer");
+             return self.getElementAsInteger(static_cast<unsigned>(i));
            },
            "index"_a)
       .def("get_element_as_double",
-           [](llvm::ConstantDataSequential &self, unsigned i) {
-             return self.getElementAsDouble(i);
+           [](llvm::ConstantDataSequential &self, Py_ssize_t i) {
+             // getElementAsDouble likewise asserts on a bad index or a
+             // non-double element type.
+             Py_ssize_t n = static_cast<Py_ssize_t>(self.getNumElements());
+             if (i < 0 || i >= n)
+               throw nb::index_error("element index out of range");
+             if (!self.getElementType()->isDoubleTy())
+               throw nb::value_error("element type is not double");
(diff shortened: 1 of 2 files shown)


makslevental:

The real diff touches only `src/IR/Constants.cpp` and adds `tests/test_const_accessors.py`. New accessors: `ConstantDataSequential.{num_elements, get_element_as_int, get_element_as_double, is_string, as_string}`, `ConstantExpr.opcode_name`, `BlockAddress.{function, basic_block}`, `GlobalAlias.aliasee`, `GlobalIFunc.resolver`. (The `ConstantInt/FP` value accessors and `GlobalVariable.initializer/is_constant` mentioned around this area are pre-existing, covered elsewhere, and correctly not re-tested here.) Probed against a build.

### 1. Out-of-range index on element accessors aborts the interpreter; untested (Important)

`test_const_accessors.py:56,62,65` only pass in-range indices. The bindings (`Constants.cpp:47,52`) forward the raw `unsigned` to `getElementAsInteger`/`getElementAsDouble` with no bounds check — a bad index kills the whole process:
```
get_element_as_int(4) on a 4-elt array:
Assertion failed: (Elt < getNumElements() && "Invalid Elt"), Constants.cpp:3039  -> SIGABRT (exit 134)
```
Add a boundary test asserting `IndexError` (not abort). This likely also warrants a guard in the binding — a library that aborts the interpreter on a bad arg is a robustness defect — but at minimum pin the failure mode.

### 2. Wrong element type aborts the interpreter; untested (Important)

`get_element_as_double` on an integer vector, and `get_element_as_int` on a double vector, both abort:
```
get_element_as_double on <4 x i32>:  Assertion failed: (getElementType()->isDoubleTy() ...) -> SIGABRT
get_element_as_int on <2 x double>:  Assertion failed: (isa<IntegerType>... ) -> SIGABRT
```
The test uses each accessor only on a matching element type. Notably `as_string` is the one element accessor that *does* guard + is tested (`:66-68`), which highlights the inconsistency — the other three CDS accessors have neither guard nor negative test.

### Smaller points

- **`get_element_as_int` signedness is unpinned.** It returns raw `uint64_t`; the test only checks positive small values, so it never documents that a negative element reads unsigned (`i8 0xFF -> 255`). Unlike `ConstantInt` (which has both `value`/sext and `zext_value`), the CDS accessor has only the unsigned form, and nothing pins it. One assertion on a negative element locks it.
- **Monolithic test hides later coverage** (`:44`). All nine accessors are in one function, so an early `assert` failure short-circuits the rest — a regression in `aliasee`/`resolver` could be masked by an unrelated earlier failure. Split per kind. Also `ce.opcode_name == "ptrtoint"` (`:71`) couples to LLVM's opcode spelling (low risk, stable API string).

### What's good

- Every accessor asserts a concrete value, not just non-None: `num_elements==4`, `get_element_as_int(0)==ord("a")`, `get_element_as_double(0)==1.5`, `as_string=="abcd"`, `opcode_name=="ptrtoint"`, `.name` on `function`/`basic_block`/`aliasee`/`resolver`. No tautologies.
- The one C++ branch (`as_string` throwing on non-string) has both polarities tested: `is_string`/`not is_string` and `pytest.raises(ValueError, match="not a string")`.
- `1.5`/`2.5` are exactly representable, so the float equality is safe, not flaky.
- Leak gate holds — the `reference_internal` accessors (`function`/`basic_block`/`aliasee`/`resolver`) leave zero live Context/Module.

PR #526, opened 2026-08-06
[eudsl-llvmpy] MLIR parity: lazy sequence views for traversal

makslevental:
Replace the eager std::vector materialization behind .functions/.basic_blocks/
.args/.instructions/.operands/.users/.uses with a lazy Sequence<T> view: it
holds closures computing length and the i-th element on demand, so len() and
integer indexing don't build a list (a slice materializes just its window).
Iteration uses Python's sequence protocol; elements downcast via type_hook.
Each accessor keep_alive<0,1>s its view to the parent, which owns the storage.
One templated bindSequence registers the seven element types (Function,
BasicBlock, Argument, Instruction, Value, User, Use).

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -112,6 +112,7 @@ nanobind_add_module(eudslllvm_ext
   src/IR/Types.cpp
   src/IR/Kinds.cpp
   src/IR/Values.cpp
+  src/IR/Sequence.cpp
   src/IR/Instructions.cpp
   src/IR/Constants.cpp
   src/IR/Builder.cpp
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -5,6 +5,7 @@
 #include "IR/Common.h"
 #include "IR/Errors.h"
 #include "IR/Ownership.h"
+#include "IR/Sequence.h"
 #include "IR/TargetInit.h"
 
 #include <llvm/AsmParser/Parser.h>
@@ -89,13 +90,23 @@ void populate_context(nb::module_ &m) {
           "context",
           [](eudsl::Module &self) -> eudsl::Context & { return self.context(); },
           nb::rv_policy::reference_internal)
-      .def_prop_ro("functions",
-                   [](eudsl::Module &self) {
-                     std::vector<llvm::Function *> out;
-                     for (llvm::Function &f : self.get().functions())
-                       out.push_back(&f);
-                     return out;
-                   })
+      .def_prop_ro(
+          "functions",
+          [](eudsl::Module &self) {
+            llvm::Module *mod = &self.get();
+            eudsl::Sequence<llvm::Function> seq;
+            seq.length = [mod] {
+              return static_cast<std::size_t>(
+                  std::distance(mod->begin(), mod->end()));
+            };
+            seq.at = [mod](std::size_t i) {
+              auto it = mod->begin();
+              std::advance(it, i);
+              return &*it;
+            };
+            return seq;
+          },
+          nb::keep_alive<0, 1>())
       .def("__len__",
            [](eudsl::Module &self) {
              return static_cast<Py_ssize_t>(std::distance(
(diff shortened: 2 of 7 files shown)


makslevental:

Reviewed `tests/test_lazy_views.py` against the `Sequence<T>` views the PR adds. Probed against a build.

### 1. Nested views (`operands`/`users`/`uses`) dangle when the module dies — segfault / silent UAF, and lifetime is never tested (Critical)

`test_lazy_views.py:62-77` exercises these views only while the whole module graph is alive. The implementation attaches `keep_alive<0,1>` to the *immediate* `Value`/`User` wrapper (`src/IR/Values.cpp`), and `Sequence.h:12` claims "the parent, whose storage owns the elements." But for these three the parent is a non-owning `Value`/instruction wrapper — it does **not** keep the owning `Module`/`LLVMContext` alive. So a live view can outlive its storage:
```
add = f.entry_block.instructions[0]; ops = add.operands
del mod, f, add; gc.collect()      # live_module_count == 0, module destroyed
len(ops)                           # SEGFAULT (exit 139)
# arg.users in the same shape doesn't crash but returns len 0 — silent use-after-free (true count is 1)
```
Contrast `Module.functions`, where the parent *is* the module: that scenario is sound (module stays alive while the view lives). So the `keep_alive` target is right for the container views and wrong for the sub-object views — an asymmetry the test can't see because every view is read while `mod` is a live local. A test holding only a nested view, dropping the module, then touching the view would have caught this. This looks like a real binding bug (the sub-object views need to keep the module/context alive), not just a test gap.

### 2. "Lazy" is asserted structurally, never behaviorally (Important)

The docstring and comments promise laziness, but the only related checks are `type(fns).__name__ == "FunctionSequence"` and `not isinstance(fns, list)` (`:35-36`) — both structural, and both would pass for an eager snapshot wrapped in a custom class. The defining behavior is never asserted:
```
insts = eb.instructions   # view created first
len(insts)                # 2
# ...emit one more instruction into eb...
len(insts)                # 3  -> view reflects the post-creation mutation
```
One such assertion is what separates real coverage of "lazy" from an unverified claim.

### 3. Leak gate only ever observes the empty post-teardown state (Important)

`test_lazy_views.py:79-80` deletes every view before `assert_no_leaks()`, and the autouse fixture checks counts only after the frame is gone. So nothing proves anything about a **live** view: neither that holding one keeps its module alive (over-release → finding 1) nor that dropping one releases it (over-retain → leak). A test asserting `live_module_count == 1` while a view is the sole reference, then `== 0` after dropping it, would exercise the `keep_alive` contract directly.

### Smaller points

- **Index-protocol edges under-covered.** Only positive out-of-range is tested (`fns[9]`, `uses[5]`). Untested though implemented (`Sequence.h:44-51`): negative out-of-range (`fns[-100]` raises `IndexError`, works), empty sequence (`g`'s args), and non-integer/non-slice index.
- **`ops[0] == args[0]` is a near-tautology** (`:66,71`) — nanobind returns the identical object for the same pointer (`ops[0] is args[0]`), so `==` succeeds by identity regardless of `Value.__eq__`. Comparing two views obtained via different paths that should be unequal would be stronger.

### What's good

- Covers all seven `Sequence<T>` instantiations (Function, BasicBlock, Argument, Instruction, Value, User, Use).
- Element downcast is genuinely verified: `isinstance(insts[0], BinaryOperator)` (`:60`) and `type(ops[0]).__name__ == "Argument"` both hold, so the `pick()` silent-degrade concern doesn't apply here.
- Slice paths including negative step are real and pass: `fns[0:1]`, `fns[::-1] == ["g","f"]` (`:41-42`).
- No brittle `str()` coupling and no `list(view) == list(view)` tautologies.

Highest-value additions, in order: (1) a nested-view-outlives-module test (surfaces the finding 1 crash), (2) a mutation-reflected-in-live-view assertion, (3) a live-view keep_alive count assertion.

makslevental:
Re: finding 1 (sub-object views dangling past the module). Addressed. This PR now adds the review's other asks (behavioral-laziness assertion, live-view keep_alive count, index-edge coverage, and a `TypeError` on non-integer index instead of a bare `std::bad_cast`).

The lifetime bug itself is fixed by a commit on the **top of the stack** (`eudsl-llvmpy-types-generic`, PR #564), following MLIR's owner-reference model (its `PyObjectRef` chain rooted at the storage owner): each `Sequence` view now holds a strong reference to the owning `Module`, and every element yielded by `__getitem__` is kept alive to that owner. Owner resolution from a bare `llvm::Value` uses a new `llvm::Module* -> eudsl::Module*` registry + `nb::find`. New test `test_nested_view_outlives_module` (previously a segfault) plus per-drop `live_module_count` assertions confirm a held view/element keeps the module alive and dropping it releases the module.

makslevental:
Correction: the lifetime fix now lives in its own PR, **#565** (`eudsl-llvmpy-view-lifetime`), on top of the stack — split out of #564 since it's an ownership fix unrelated to the is_generic types port.

PR #527, opened 2026-08-06
[eudsl-llvmpy] MLIR parity: implicit current context

makslevental:
Add a thread-local current-context stack: Context.__enter__ pushes, __exit__
pops, and Context.current() exposes the innermost active one (or None). The
single-argument primitive type factories (void_t/label_t/i1..f64) take an
optional context that defaults to the current one, so inside 'with Context():'
you can write llvm.i32() instead of llvm.i32(ctx). Explicit context still
works, so the DSL (which threads context explicitly) is unaffected; calling a
factory with neither raises a clear error.

Multi-argument factories (int_t/ptr_t/struct_t, const_bool, md_string, Module,
IRBuilder) keep context as a required leading argument -- making it optional
there would reorder the positional API, a breaking change left for later.


makslevental:

Reviewed `tests/test_current_context.py` and the vector-downcast/factory-signature changes in `tests/test_types.py`. Updated after weekend refactors.

---

### 1. Thread-locality is a documented guarantee with zero test coverage (Important)

`Ownership.cpp` implements the stack as `static thread_local std::vector<Context*> gContextStack;`. Nothing exercises more than one thread. Drop `thread_local` (making it a plain global) and every test passes, while concurrent use silently corrupts each thread's current context. Assert a child thread sees `Context.current() is None` while the parent holds a context.

### 2. Implicit-context path partially covered (Suggestion)

Originally only 2 factories tested; now 9 factory calls covered: `i32()`, `f32()`, `void()`, `i64()`, `f64()`, `int(7)`, `ptr()`, `struct([...])`, `md_string("hi")`. Still untested with implicit context: `Module('m')`, `IRBuilder()`, `parse_assembly(...)`, `const_bool(True)`, `md_node([...])`, `BasicBlock.create('entry')`.

### Smaller points

- **No test that `__exit__` restores on exception.** Only clean LIFO nesting tested. An exception propagating out of the `with` body currently behaves, but nothing guards a regression.
- **Vector downcast: accessor split not negatively asserted.** No `assert not hasattr(sv, "num_elements")` for `ScalableVectorType`.
- **Same-Context reentry (`with a: with a:`) untested.** Works today (stack pops cleanly, live count returns to 0), but unguarded.

### What's good

- Nesting test is precise: `current() is None` before, innermost-wins during, restore-to-parent after inner exit, `None` after outer exit.
- `assert types.i32() == types.i32(ctx)` is meaningful: types are uniqued per-context, so it proves implicit factory resolved to the *same* context.
- The no-context test asserts the real error type and a substring of the real message.
- `test_types.py` correctly migrated to keyword `context=` for the reordered factories.
- Leak hygiene sound: autouse fixture checks both counts post-teardown.

projects/eudsl-llvmpy/tests/test_dsl_values.py
makslevental, line 1: isn't the whole point of this PR that you don't need to pass ctx anymore? shouldn't these all be removed now? and in the rest of the tests?

makslevental: Still relevant. The tests in `test_dsl_values.py` still pass `ctx` explicitly everywhere (`llvm.Context() as ctx`, `llvm.Module("m", ctx)`, `llvm.types.i32(ctx)`, `llvm.IRBuilder(ctx)`). The whole point of this PR is the implicit current-context fallback, so these should exercise the implicit path — drop the explicit `ctx` arg and rely on the `with llvm.Context()` block setting the current context.

PR #528, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: Context and Module lifetimes

makslevental:
First PR of the eudsl-llvmpy rewrite. Replaces the litgen-generated LLVM C-API bindings with hand-written nanobind bindings over the LLVM C++ API, and lays the Context/Module foundation the rest of the stack builds on.

- Removes litgen, `eudsl-llvmpy-generate.py`, the generated sources, and the litgen/eudsl-tblgen build deps.
- Binds `Context` and `Module` directly over the C++ types. The `LLVMContext` is held by `shared_ptr` and each `Module` keeps it alive, so a Module that outlives its `with Context()` block stays safe.
- Live-instance counting for Context and Module plus moved-from tracking, so leaks and use-after-consume are observable from Python.
- `EUDSL_LLVMPY_TARGETS` picks which LLVM targets to link and initialize, defaulting to the host arch so the extension builds against a host-only LLVM distro.


projects/eudsl-llvmpy/src/IR/Context.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  nb::class_<eudsl::Module>(m, "Module")
+      .def(nb::init<const std::string &, eudsl::Context &>(), "name"_a,
+           "context"_a, nb::keep_alive<1, 3>())
+      .def_prop_rw(
+          "name",
+          [](eudsl::Module &self) { return self.get().getModuleIdentifier(); },
+          [](eudsl::Module &self, const std::string &name) {
+            self.get().setModuleIdentifier(name);
+            self.get().setSourceFileName(name);
```
makslevental: doesn't make sense these are both set here - these should be individual `source_filename` and `module_identifier`. in general don't take liberties renaming methods/fields/etc even if you believe they're semantically equivalent.

projects/eudsl-llvmpy/src/IR/Context.cpp
```diff
@@ -0,0 +1,91 @@
...
+      .def("__str__", [](eudsl::Module &self) {
+        std::string s;
+        llvm::raw_string_ostream os(s);
+        self.get().print(os, nullptr);
+        return s;
+      });
+
+  m.def(
+      "parse_assembly",
+      [](const std::string &ir, eudsl::Context &ctx, const std::string &name) {
```
makslevental: same thing here: not one `name` but separate `module_identifier` and `source_filename`. after fixing this problem here (specious renames) go through the rest of the PR stack and fix anywhere you've done this as well.

projects/eudsl-llvmpy/src/IR/Ownership.cpp
```diff
@@ -0,0 +1,53 @@
...
+    : mod(std::make_unique<llvm::Module>(name, ctx.get())), owner(&ctx) {}
+
+Module::Module(std::unique_ptr<llvm::Module> m, Context &ctx)
+    : mod(std::move(m)), owner(&ctx) {}
+
+llvm::Module &Module::get() const {
+  if (!mod)
+    throw std::runtime_error(
+        "module has been consumed (moved into the JIT) and can no longer be "
+        "used");
```
makslevental: LLVM coding style dictates that multiline conditionals and conditionals with multiline conditions need braces:
```suggestion
  if (!mod) {
    throw std::runtime_error(
        "module has been consumed (moved into the JIT) and can no longer be "
        "used");
  }
```

projects/eudsl-llvmpy/tests/test_bindings.py
```diff
@@ -1,123 +1,48 @@
...
 
 
 def test_symbol_collision():
-    # noinspection PyUnresolvedReferences
-    import eudsl_tblgen
+    # eudsl-tblgen is a separate extension in a different nanobind domain;
+    # importing both must not clash.
+    import eudsl_tblgen  # noqa: F401
+
+    import llvm  # noqa: F401
```
makslevental: you don't need to do local imports **ever**. fix in this PR and then go through the rest of the stack and hoist local imports

projects/eudsl-llvmpy/tests/test_context.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_module_rename():
+    with llvm.Context() as ctx:
+        mod = llvm.Module("before", ctx)
+        mod.name = "after"
+        assert mod.name == "after"
+        assert "ModuleID = 'after'" in str(mod)
+        del mod
+    gc.collect()
+    assert_no_leaks()
```
makslevental: seems the `gc.collect()` should just be in `assert_no_leaks` since they appear almost everywhere together

PR #530, opened 2026-08-07
[eudsl-llvmpy] DSL value layer: ArithValue operators and pointer/aggregate sugar

makslevental:
Starts the DSL value layer on top of the object layer.

- `register_value_caster` / `maybe_downcast`, backed by a stateless `nb::inst_reference` primitive. This is MLIR-style caster registration rather than a base-class monkeypatch.
- `ArithValue`, a registered Value subclass carrying arithmetic and comparison operators with int/float dispatch, scalar coercion, and typed chaining.
- Pointer, GEP, load, and store sugar with explicit element types, plus aggregate extract/insert.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Casters.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Casters.cpp
@@ -1,0 +1,33 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Stateless support for the Python value-caster layer (see llvm/dsl/casters.py),
+// which is analogous to MLIR's register_value_caster / PyValue::maybeDownCast.
+//
+// The C++ type_hook (Kinds.h) already downcasts a returned llvm::Value* to its
+// concrete *bound C++* class (Instruction, ConstantInt, ...). The Python caster
+// layer adds a second, user-extensible step: re-wrapping the same Value* as a
+// Python subclass (e.g. the DSL's ArithValue).
+//
+// This file deliberately holds NO Python references at static scope -- the
+// caster registry lives in Python -- so nothing leaks at interpreter shutdown.
+// The one primitive C++ must provide is nb::inst_reference, which binds an
+// existing C++ pointer into a chosen Python type with its lifetime tied to a
+// parent object (nanobind offers no Python-level equivalent).
+
+#include "IR/Common.h"
+
+#include <llvm/IR/Value.h>
+
+void populate_casters(nb::module_ &m) {
+  m.def(
+      "_wrap_value_as",
+      [](llvm::Value *v, nb::handle pyType, nb::handle parent) -> nb::object {
+        return nb::steal(nb::detail::nb_inst_reference(
+            (PyTypeObject *)pyType.ptr(), (void *)v, parent.ptr()));
+      },
+      "value"_a, "py_type"_a, "parent"_a = nb::none(),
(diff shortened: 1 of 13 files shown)


projects/eudsl-llvmpy/src/llvm/dsl/casters.py
```diff
@@ -0,0 +1,49 @@
...
+    """Re-wrap `value` as its registered caster subclass, if any.
+
+    `parent` (a Context/Module/Value) ties the wrapper's lifetime to the owner
+    of the underlying pointer, matching the reference_internal convention used
+    throughout the bindings.
+    """
+    caster = _casters.get(value.type.type_id)
+    if caster is None:
+        return value
+    return _wrap_value_as(value, caster, parent)
```
makslevental, line 49: why did you put this stuff in Python when the mlir python bindings puts it all in cpp

projects/eudsl-llvmpy/src/IR/Casters.cpp
```diff
@@ -0,0 +1,33 @@
...
+  m.def(
+      "_wrap_value_as",
+      [](llvm::Value *v, nb::handle pyType, nb::handle parent) -> nb::object {
+        return nb::steal(nb::detail::nb_inst_reference(
+            (PyTypeObject *)pyType.ptr(), (void *)v, parent.ptr()));
+      },
+      "value"_a, "py_type"_a, "parent"_a = nb::none(),
+      "Re-wrap an existing Value* as the given Python (sub)type, tying its "
+      "lifetime to `parent`. Backs llvm.dsl.casters.maybe_downcast.");
+}
```
makslevental, line 33: look below where i comment that this probably doesn't need to be exposed

projects/eudsl-llvmpy/src/llvm/dsl/casters.py
makslevental: Still relevant. The caster registry and `maybe_downcast` logic lives in Python (`casters.py`) while MLIR puts the equivalent (`register_value_caster`, `PyValue::maybeDownCast`) in C++. The C++ side here (`Casters.cpp`) only exposes `_wrap_value_as` (a thin `nb_inst_reference` call). The trade-off: Python-side is more extensible for users, but adds per-call overhead and makes the caster invisible to C++ code paths. Worth reconsidering if performance matters for hot loops.

projects/eudsl-llvmpy/src/IR/Casters.cpp
makslevental: Still relevant. `_wrap_value_as` is the only thing in `Casters.cpp` — it is a single `nb_inst_reference` call. If the Python caster layer were moved to C++ (per the comment above), this file would be absorbed into that implementation and would not need a separate public binding.

projects/eudsl-llvmpy/src/llvm/dsl/casters.py
makslevental: Fixed. Moved the caster registry to C++ — `register_value_caster` and `maybe_downcast` are now C++ functions backed by a static `unordered_map<unsigned, nb::object>`, mirroring MLIR's `PyGlobals::valueCasterMap` pattern. Python `casters.py` is now a thin decorator wrapper over the C++ implementation. `_wrap_value_as` is gone.

projects/eudsl-llvmpy/src/IR/Casters.cpp
makslevental: Fixed. `_wrap_value_as` no longer exists. The C++ now exposes `register_value_caster(type_id, caster)` and `maybe_downcast(value, parent)` directly — the registry and the wrap logic are both internal to `Casters.cpp`.

PR #532, opened 2026-08-07
[eudsl-llvmpy] Attributes, metadata, and error handling

makslevental:
Binds the attribute and diagnostic surface on top of the core object layer.

- Function/parameter attributes plus the linkage and calling-convention enums.
- Metadata: `MDString`, `MDNode`, and attachment onto instructions and globals.
- Error handling that turns LLVM diagnostics (parse and verify failures) into Python exceptions instead of aborts.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Attributes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Attributes.cpp
@@ -4,21 +4,16 @@
 
 #include "IR/Common.h"
 
-#include <llvm/IR/CallingConv.h>
 #include <llvm/IR/GlobalValue.h>
 
 void populate_attributes(nb::module_ &m) {
-  // Linkage (GlobalValue::LinkageTypes) is bound with the GlobalValue hierarchy
-  // in Values.cpp, next to the Function/global classes that use it.
   nb::enum_<llvm::GlobalValue::VisibilityTypes>(m, "Visibility")
       .value("DEFAULT", llvm::GlobalValue::DefaultVisibility)
       .value("HIDDEN", llvm::GlobalValue::HiddenVisibility)
       .value("PROTECTED", llvm::GlobalValue::ProtectedVisibility);
 
-  // CallingConv::ID is a namespace of unsigned constants, not an enum class;
-  // expose the common ones as module-level ints under a submodule.
-  nb::module_ cc = m.def_submodule("CallingConv");
-  cc.attr("C") = (unsigned)llvm::CallingConv::C;
-  cc.attr("FAST") = (unsigned)llvm::CallingConv::Fast;
-  cc.attr("COLD") = (unsigned)llvm::CallingConv::Cold;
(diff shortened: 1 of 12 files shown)


projects/eudsl-llvmpy/src/IR/Attributes.cpp
```diff
@@ -0,0 +1,24 @@
...
+      .value("DEFAULT", llvm::GlobalValue::DefaultVisibility)
+      .value("HIDDEN", llvm::GlobalValue::HiddenVisibility)
+      .value("PROTECTED", llvm::GlobalValue::ProtectedVisibility);
+
+  // CallingConv::ID is a namespace of unsigned constants, not an enum class;
+  // expose the common ones as module-level ints under a submodule.
+  nb::module_ cc = m.def_submodule("CallingConv");
+  cc.attr("C") = (unsigned)llvm::CallingConv::C;
+  cc.attr("FAST") = (unsigned)llvm::CallingConv::Fast;
+  cc.attr("COLD") = (unsigned)llvm::CallingConv::Cold;
```
makslevental: instead of doing this just create an enum class whose values match the ints

projects/eudsl-llvmpy/src/IR/Values.cpp
```diff
@@ -166,5 +168,32 @@ void populate_values(nb::module_ &m) {
...
-          "name"_a = "", nb::rv_policy::reference_internal);
+          "name"_a = "", nb::rv_policy::reference_internal)
+      .def_prop_rw("linkage", &llvm::Function::getLinkage,
+                   &llvm::Function::setLinkage)
+      .def_prop_rw("visibility", &llvm::Function::getVisibility,
+                   &llvm::Function::setVisibility)
+      .def_prop_rw(
+          "calling_conv",
+          [](llvm::Function &self) { return (unsigned)self.getCallingConv(); },
+          [](llvm::Function &self, unsigned cc) {
```
makslevental: see the comments above where i say you should refactor to use an enum class instead of ints

projects/eudsl-llvmpy/src/IR/Attributes.cpp
makslevental: Still relevant. `CallingConv` is still exposed as module-level unsigned ints on a submodule (`cc.attr("C") = (unsigned)llvm::CallingConv::C`). Should use `nb::enum_` to get a proper enum class with type safety and repr, same as `Visibility` and `Linkage` are already done with `nb::enum_<>` in this same file.

projects/eudsl-llvmpy/src/IR/Values.cpp
makslevental: Still relevant. `calling_conv` getter/setter in `Values.cpp` still works with raw `unsigned` values. Once `CallingConv` is bound as an `nb::enum_<>` (per the comment on Attributes.cpp), this property should accept/return that enum type instead of bare ints.

PR #534, opened 2026-08-07
[eudsl-llvmpy] DSL: vendor the AST canonicalizer and CF transformers

makslevental:
Vendors the AST canonicalizer and control-flow transformers from eudsl-python-extras with the MLIR dependencies stripped. These rewrite a decorated Python function's AST into the shape the control-flow lowering consumes, and stand alone here so the lowering PR that uses them stays reviewable.


projects/eudsl-llvmpy/tests/test_ast.py
makslevental, line 1: you need to test linenos just like i do in eudsl-python-extras

makslevental: Still relevant. The AST transformers manipulate line numbers extensively (`set_lineno`, `increment_lineno`, `end_lineno` adjacency checks for elif detection) but no test asserts that output nodes have correct line numbers. The `eudsl-python-extras` pattern of asserting `node.lineno` / `node.end_lineno` after transformation would catch regressions in the line-number arithmetic.

projects/eudsl-llvmpy/src/llvm/ast/canonicalize.py
```diff
@@ -0,0 +1,140 @@
...
+        self.context = context
+        self.first_lineno = first_lineno
+
+
+class StrictTransformer(Transformer):
+    def visit_FunctionDef(self, node: ast.FunctionDef):
+        return node
+
+
+def transform_func(f, *transformer_ctors: type(Transformer)):
```
makslevental: is this type annotation incorrect?

projects/eudsl-llvmpy/src/llvm/ast/canonicalize.py
```diff
@@ -0,0 +1,140 @@
...
+        logger.debug("[transformed code diff]\n%s", "\n" + "\n".join(diff))
+        logger.debug("[transformed code]\n%s", new_code)
+        module.body[0] = func_node
+
+    logger.debug("[final transformed code]\n\n%s", new_code)
+
+    return module
+
+
+# TODO(max): unify with `replace_closure` in ast/utils.py
```
makslevental: this comment doesn't make sense (`replace_closure` doesn't exist in the current ast/util.py)

projects/eudsl-llvmpy/src/llvm/ast/canonicalize.py
makslevental, line 1: what is the point of using `type(Transformer)` instead of `Transformer`?

projects/eudsl-llvmpy/src/llvm/ast/canonicalize.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @abstractmethod
+    def patch_function(self, original_f):
+        pass  # pragma: no cover
+
+
+def patch_function(f, patchers: List[type(FunctionPatcher)]):
+    context = types.SimpleNamespace()
+    for patcher in patchers:
+        new_f = patcher(context).patch_function(f)
```
makslevental: this is completely wrong - you need to apply all of the patchers.
```suggestion
    new_f = f
    for patcher in patchers:
        new_f = patcher(context).patch_function(new_f)
```

add a test for this

PR #535, opened 2026-08-07
[eudsl-llvmpy] DSL control flow: if/elif/else, while, for lowered to blocks and phi nodes

makslevental:
Lowers Python control flow to basic blocks and phi nodes.

- `if` / `elif` / `else`.
- `while`, and `for` over `range_`, with loop-carried phis. A phi-based loop runtime is fed by AST transforms that lift each loop body into nested cond/body functions parameterized by the carried variables.
- `break`, `continue`, and early `return` raise `NotImplementedError` for now.

### DIFF
--- a/projects/eudsl-llvmpy/src/llvm/dsl/cf.py
+++ b/projects/eudsl-llvmpy/src/llvm/dsl/cf.py
@@ -1,0 +1,160 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""Runtime for lowered control flow: real basic blocks and phi nodes.
+
+The AST canonicalizer (llvm.ast.cf_transformers) rewrites
+
+    if c:
+        r = yield a + 1
+    else:
+        r = yield b
+    return r
+
+into
+
+    with if_ctx_manager(c, (placeholder_opaque_t(),)) as __if_op__:
+        r = yield_(a + 1)
+    with else_ctx_manager(__if_op__):
+        r = yield_(b)
+    return r
+
+The MLIR analogue uses regions whose results become block/region results. Here
+we emit explicit LLVM blocks: `if_ctx_manager` creates then/merge blocks and a
+provisional conditional branch (false edge -> merge); `else_ctx_manager`
+repoints that false edge to a fresh else block; and the values passed to
+`yield_` become phi nodes at the merge block. Because Python's `r = yield_(...)`
+executes in the else branch last, the else-branch `yield_` is the one that
+builds the phis (both incoming edges are then known) and returns them, so `r`
+binds to the phi that `return r` uses.
+"""
+
+from contextlib import contextmanager
+
+from ..ast.canonicalize import Canonicalizer, FunctionPatcher
+from ..ast import cf_transformers as _T
+from .casters import maybe_downcast
+from .context import current_builder, current_function
+
+
+def placeholder_opaque_t():
+    # Marks a phi-result slot in the rewritten AST. The real value is the phi
+    # built by yield_; the placeholder itself is never inspected here.
+    return None
+
+
+class _IfOp:
+    """Bookkeeping for one lowered if/else."""
+
+    def __init__(self, cond):
+        b = current_builder()
+        fn = current_function()
+        self.builder = b
+        self.cond = cond
+        self.entry_block = b.insert_block
+        self.then_block = fn.append_basic_block("if.then")
+        self.merge_block = fn.append_basic_block("if.end")
+        self.else_block = None
+        # Provisional branch; else_ctx_manager repoints successor 1 (false edge).
+        self.cond_br = b.cond_br(cond, self.then_block, self.merge_block)
+        self.then_vals = None
+        self.else_vals = None
+        self.active = "then"
+
+    def _terminate_current(self):
+        b = self.builder
+        if b.insert_block.terminator is None:
+            b.br(self.merge_block)
+
+    def record_and_maybe_phi(self, values):
+        """Called by yield_. Record this branch's values; on the else branch
+        (both edges known) build the phis and return them."""
+        b = self.builder
+        if self.active == "then":
+            self.then_vals = list(values)
+            self._terminate_current()
+            # No phis yet; the value bound here is overwritten by the else
+            # branch's assignment (or unused for a side-effecting if). Match the
+            # else-branch return shape (scalar for a single value).
+            vals = list(values)
+            if len(vals) == 1:
+                return vals[0]
+            return tuple(vals)
+        # else branch
+        self.else_vals = list(values)
+        self._terminate_current()
+        b.set_insert_point(self.merge_block)
+        phis = []
+        for i, tv in enumerate(self.then_vals):
+            phi = b.phi(tv.type, f"if.phi.{i}")
+            phi.add_incoming(tv, self.then_block)
+            phi.add_incoming(self.else_vals[i], self.else_block)
+            phis.append(maybe_downcast(phi, current_function()))
+        self.phis = phis
+        if len(phis) == 1:
+            return phis[0]
+        return tuple(phis)
+
+
+_if_stack = []
+
+
+@contextmanager
+def if_ctx_manager(cond, results=()):
+    op = _IfOp(cond)
+    _if_stack.append(op)
+    op.builder.set_insert_point(op.then_block)
+    try:
+        yield op
+    finally:
+        # Each branch body is terminated by its trailing yield_ (the
+        # canonicalizer guarantees one); do not terminate here or the merge
+        # block would get a self-branch. Just leave the builder at merge for
+        # whatever follows (a no-else if continues here; an else block enters
+        # next and repositions).
+        _if_stack.pop()
+        op.builder.set_insert_point(op.merge_block)
+
+
+@contextmanager
+def else_ctx_manager(op):
+    b = op.builder
+    fn = current_function()
+    op.else_block = fn.append_basic_block("if.else")
+    # Repoint the entry conditional branch's false edge to the else block.
+    op.cond_br.set_successor(1, op.else_block)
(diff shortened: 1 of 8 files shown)


makslevental:

Reviewed `tests/test_dsl_cf.py` and `tests/test_dsl_cf_reject.py` against `src/llvm/dsl/cf.py` and the CF transformers. Updated after weekend refactors.

---

### 1. `range_` negative step silently miscompiles (Important)

`for_loop`'s exit condition is still hardcoded to `iv < stop_v` regardless of step sign. Step-2 ascending is now tested (`test_for_range_single_and_stepped_args`, `test_for_with_step_and_two_carried`), but no descending/negative-step test exists. This is a genuine correctness bug:
```
countdown(5) = 0     # python sum(range(5,0,-1)) = 15; loop body never runs (5 < 0 false)
```

### ~~2. `range_` argument-count variants unguarded~~ — FIXED

1-arg (`range_(n)`), 3-arg (`range_(0, n, 2)`), and error case (`test_for_range_wrong_arg_count_raises`) all now covered.

### 3. Reject/error paths — remaining gap (Suggestion)

Most reject paths now covered: `_reject_nested_control_flow` via two dedicated tests, `_carried_from_yield` guards, "body must end with yield," "for target single name," and "range_ 1-3 args." The sole remaining gap: early return inside `while`/`for` specifically (vs. `if`) — same handler, different visitor entry.

### Smaller points

- **CFG structure / phi wiring never asserted** — no test inspects phi incomings (value, predecessor pairs). Correctness rests solely on JIT execution.
- **`mod.verify()` never called** — `test_if_else_produces_phi` is pure substring-matching; a malformed CFG could pass. Add `assert mod.verify() is None`.
- **`printed.count("phi i32") == 2`** couples to phi count/type; brittle under harmless refactors (low risk given JIT checks).

### What's good

- JIT tests execute real machine code on edge inputs: `sum_to(0)`, `total(1)`, and the full 3-way `classify`.
- Reject tests use `pytest.raises(..., match=...)` asserting *which* rejection fired.
- `conftest.py` leak gate is real and load-bearing.

PragmaTwice (commented):
Ahh so Python `yield` is used to indicate a new phi node here?

makslevental:
> Ahh so Python `yield` is used to indicate a new phi node here?

It's something like this - I'm not 100% sure how this works (Claude copied/adapted what I did for mlir-python-extras for loops and conditionals but there there are no phis). Reading the docs you're right I guess but then it moves the phis themselves around because obv the phis don't belong in the actual conditonal bbs...

I plan to review more closely when I get to this PR in the stack.

projects/eudsl-llvmpy/src/llvm/dsl/func.py
```diff
@@ -0,0 +1,54 @@
...
+        param_types = [_resolve(p.annotation, ctx) for p in sig.parameters.values()]
+        ret_type = _resolve(sig.return_annotation, ctx)
+        fn_name = name or f.__name__
+
+        fn = Function.create(function_t(ret_type, param_types), fn_name, module)
+        entry = fn.append_basic_block("entry")
+        builder = IRBuilder(ctx)
+
+        # Rewrite Python control flow into the cf context-manager calls.
+        f = canonicalize(using=LLVMCanonicalizer())(f)
```
makslevental: this should not be automatically applied and instead should be used as a decorator just as in eudsl-python-extras

projects/eudsl-llvmpy/src/llvm/dsl/cf.py
```diff
@@ -0,0 +1,284 @@
...
+        exit:      (phis are the loop results)
+    """
+    b = current_builder()
+    fn = current_function()
+    inits = list(inits)
+
+    preheader = b.insert_block
+    header = fn.append_basic_block("while.header")
+    body_bb = fn.append_basic_block("while.body")
+    exit_bb = fn.append_basic_block("while.end")
```
makslevental, line 207: add a test with multiple while loops to be sure these labels are uniquefied by LLVM itself

projects/eudsl-llvmpy/src/llvm/dsl/cf.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+@contextmanager
+def else_ctx_manager(op):
+    b = op.builder
+    fn = current_function()
+    op.else_block = fn.append_basic_block("if.else")
+    # Repoint the entry conditional branch's false edge to the else block.
+    op.cond_br.set_successor(1, op.else_block)
+    op.active = "else"
```
makslevental, line 138: what does this do?

projects/eudsl-llvmpy/src/llvm/dsl/func.py
makslevental: Done — removed the auto-apply from `@function`. Users now stack `@canonicalize(using=LLVMCanonicalizer())` explicitly, same as eudsl-python-extras.

projects/eudsl-llvmpy/src/llvm/dsl/cf.py
makslevental: Added `test_multiple_while_loops_have_unique_labels` — two while loops in the same function, asserts both `while.header:` and `while.header1:` appear in the IR, plus JIT-executes to confirm correctness.

projects/eudsl-llvmpy/src/llvm/dsl/cf.py
makslevental: Added a comment: `op.active = "else"` switches `yield_` into phi-building mode — when `yield_` sees `active == "else"` it knows both edges are known and builds the phi nodes, vs when `active == "then"` it just records the then-branch values for later.

makslevental:
## How Python loops and conditionals map to phi nodes

Lowering happens in two stages. First a **source-level AST rewrite**
(`llvm/ast/cf_transformers.py`, driven by `LLVMCanonicalizer`) turns Python
`if`/`while`/`for` into calls to a small set of runtime helpers. Then, when the
rewritten function runs under the `@function` builder, those **runtime helpers**
(`llvm/dsl/cf.py`) emit real LLVM blocks and phi nodes. The bridge between the
two stages is a `yield` protocol: whatever a branch or loop body `yield`s is the
set of SSA values that live *out* of it, and those become the phi incomings.

### Conditionals (`if` / `elif` / `else`)

The AST pass rewrites

```python
if c:
    r = yield a + 1
else:
    r = yield b
return r
```

into a pair of context managers (elif is first canonicalized into a nested
`if` in the `else`; a branch with no explicit `yield` gets an empty one
appended so every branch ends by yielding its live-out values):

```python
with if_ctx_manager(c, (placeholder_opaque_t(),)) as __if_op__:
    r = yield_(a + 1)
with else_ctx_manager(__if_op__):
    r = yield_(b)
return r
```

At runtime this emits:

```
entry:    %c = ...
          br i1 %c, label %if.then, label %if.end   ; false edge provisional
if.then:  %t = add ...
          br label %if.end
if.else:  ; false edge repointed here by else_ctx_manager
          br label %if.end
if.end:   %r = phi [ %t, %if.then ], [ %b, %if.else ]
```

- `if_ctx_manager` creates the `then` and `merge` blocks and emits a provisional
  conditional branch whose false edge goes straight to `merge`.
- `else_ctx_manager` appends an `else` block and repoints that false edge to it.
- Each branch's trailing `yield_(...)` records that branch's live-out values and
  terminates the branch with a branch to `merge`. The **else branch's `yield_`
  runs last** — with both predecessors now known — so it is the one that builds
  one `phi` per carried value at `merge` and returns them. That is why `r` binds
  to the phi that `return r` consumes.
- Incomings are added from the *actual* predecessor block captured at `yield_`
  time (not the branch's entry block), so a nested `if` inside a branch — whose
  real predecessor is an inner merge block — wires up correctly.
- An `if` with no `else` has no value-merging phi; its false edge already targets
  `merge`, so it is used only for its side effects.

### Loops (`while` / `for`)

The loop-carried variables are named by the loop body's trailing
`yield a, b`. The AST pass **lifts the body into nested functions parameterized
by those carried variables** and replaces the loop with a call:

```python
while COND:            def __wcond_L__(acc, i): return COND
    BODY         ->    def __wbody_L__(acc, i): BODY; return (acc, i)
    yield acc, i       (acc, i) = while_loop(__wcond_L__, __wbody_L__, (acc, i))
```

Parameterizing `cond`/`body` by the carried values (rather than closing over
them) is what makes the phi wiring clean: the runtime can pass the header phis
straight in as arguments and thread the body's results back as phi incomings,
with no Python closure rebinding. `while_loop` emits:

```
preheader: br label %header
header:    %acc = phi [ %acc.init, %preheader ], [ %acc.next, %body ]
           %i   = phi [ %i.init,   %preheader ], [ %i.next,   %body ]
           br i1 cond_fn(%acc, %i), label %body, label %while.end
body:      (%acc.next, %i.next) = body_fn(%acc, %i)
           br label %header      ; second phi incoming wired from the *end* block
while.end: ; the header phis are the loop results
```

- One header `phi` is created per carried value, seeded with `[init, preheader]`.
- `cond_fn(*phis)` drives the header's conditional branch (body vs. exit).
- `body_fn(*phis)` computes the next-carried tuple; after the body is emitted, a
  branch back to `header` is added and the phi's second incoming is wired from
  the body's *actual* end block (which may differ from the body entry if the
  body itself introduced control flow).
- At `while.end` the header phis are the loop's results, returned to the call
  site and tuple-unpacked back into the carried variable names.

`for i in range_(start, stop, step)` is lowered on top of `while_loop` by adding
the induction variable `i` as an extra carried value: the condition is
`i < stop` (or `i > stop` for a negative step) and the body prepends `i + step`
to the carried tuple. The induction variable is internal and dropped from the
returned results.

### Not modeled

`break`, `continue`, and early `return` are rejected up front (they would need
edge duplication / predecessor bookkeeping the yield protocol does not do), as
is control flow nested inside a loop body (the lifted body functions are not
revisited by the if/else transformers).

makslevental:
@PragmaTwice does https://github.com/llvm/eudsl/pull/535#issuecomment-5299033589 sound reasonable to you?

PragmaTwice:
Ya looks good to me.

PR #536, opened 2026-08-07
[eudsl-llvmpy] DSL @function: declarations, calls, returns, linkage/cc/attrs/varargs

makslevental:
Adds the `@function` decorator: function definitions, empty-body declarations, calls through `__call__`, returns, and linkage, calling-convention, attribute, and varargs specification.

### DIFF
--- a/projects/eudsl-llvmpy/tests/test_dsl_cf_reject.py
+++ b/projects/eudsl-llvmpy/tests/test_dsl_cf_reject.py
@@ -1,6 +1,7 @@
 #  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 #  See https://llvm.org/LICENSE.txt for license information.
 #  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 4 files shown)


projects/eudsl-llvmpy/tests/test_dsl_cf_reject.py
```diff
@@ -44,23 +59,41 @@ def bad2(c: llvm.types.i1, a: i32) -> i32:
...
+        mod = llvm.Module("m", ctx)
+        i32 = llvm.types.i32(ctx)
+
+        @llvm.function(module=mod)
+        def ok(c: llvm.types.i1, a: i32, b: i32) -> i32:
+            if c:
+                r = yield a
+            else:
+                r = yield b
+            return r
```
makslevental, line 97: doesn't this need a decorator now?

makslevental: Yes, fixed — added `@canonicalize(using=LLVMCanonicalizer())` here. It was missed because the decorator removal was done on `dsl-cf` and this test was added on `dsl-func` above it.

PR #538, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: llvm::Type hierarchy, factories, and type_hook downcasting

makslevental:
Binds the `llvm::Type` hierarchy.

- The `llvm::Type` base and the primitive types (void, label, integer widths, half/bfloat/float/double).
- Derived types with their factories: integer, pointer, struct (literal and named), array, vector, function.
- A `type_hook` that downcasts an `llvm::Type*` to its concrete bound Python class from its `TypeID`, so an accessor returning a base `Type` surfaces as the right subclass.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Kinds.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Kinds.cpp
@@ -8,11 +8,6 @@
 
 namespace eudsl {
 
-const std::type_info *pick(const std::type_info *concrete,
-                           const std::type_info *base) {
-  return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
-}
-
 const std::type_info *typeTypeInfo(llvm::Type::TypeID id) {
   switch (id) {
   case llvm::Type::IntegerTyID:
@@ -26,8 +21,9 @@ const std::type_info *typeTypeInfo(llvm::Type::TypeID id) {
   case llvm::Type::ArrayTyID:
     return &typeid(llvm::ArrayType);
   case llvm::Type::FixedVectorTyID:
+    return &typeid(llvm::FixedVectorType);
   case llvm::Type::ScalableVectorTyID:
-    return &typeid(llvm::VectorType);
+    return &typeid(llvm::ScalableVectorType);
   default:
     // Void, Label, Metadata, Token, the float kinds, TypedPointer and
     // TargetExtType have no bound subclass; they stay llvm::Type.
--- a/projects/eudsl-llvmpy/src/IR/Kinds.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Kinds.cpp
@@ -1,0 +1,38 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "IR/Kinds.h"
+
+#include <llvm/IR/DerivedTypes.h>
+
+namespace eudsl {
+
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base) {
+  return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
+}
+
+const std::type_info *typeTypeInfo(llvm::Type::TypeID id) {
+  switch (id) {
+  case llvm::Type::IntegerTyID:
+    return &typeid(llvm::IntegerType);
+  case llvm::Type::FunctionTyID:
+    return &typeid(llvm::FunctionType);
+  case llvm::Type::PointerTyID:
+    return &typeid(llvm::PointerType);
+  case llvm::Type::StructTyID:
+    return &typeid(llvm::StructType);
+  case llvm::Type::ArrayTyID:
+    return &typeid(llvm::ArrayType);
+  case llvm::Type::FixedVectorTyID:
+  case llvm::Type::ScalableVectorTyID:
+    return &typeid(llvm::VectorType);
+  default:
+    // Void, Label, Metadata, Token, the float kinds, TypedPointer and
+    // TargetExtType have no bound subclass; they stay llvm::Type.
+    return &typeid(llvm::Type);
+  }
+}
+
+} // namespace eudsl
--- a/projects/eudsl-llvmpy/src/IR/Kinds.h
+++ b/projects/eudsl-llvmpy/src/IR/Kinds.h
@@ -1,0 +1,43 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Downcasting support. llvm::Value and llvm::Type are deliberately
+// non-polymorphic (Value.h documents the non-virtual destructor), so
+// nanobind's RTTI-based downcast cannot see through a base pointer.
+// nanobind::detail::type_hook is the documented hook for this case.
+//
+// INVARIANT: every std::type_info returned here must name a class registered
+// with nanobind, otherwise the conversion raises "Unable to convert function
+// return value to a Python type". valueTypeInfo() guards each return with
+// pick() so a not-yet-registered class falls back to base Value, keeping every
+// commit green while later tasks register the leaves.
+
+#pragma once
+
+#include <llvm/IR/Type.h>
+#include <llvm/IR/Value.h>
+
+#include <nanobind/nanobind.h>
+
+#include <typeinfo>
+
+namespace eudsl {
+const std::type_info *typeTypeInfo(llvm::Type::TypeID id);
+const std::type_info *valueTypeInfo(unsigned valueID);
+/// Return `concrete` if a nanobind class is registered for it, else `base`.
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base);
+} // namespace eudsl
+
+template <> struct nanobind::detail::type_hook<llvm::Type> {
+  static const std::type_info *get(llvm::Type *t) {
+    return t ? eudsl::typeTypeInfo(t->getTypeID()) : &typeid(llvm::Type);
+  }
+};
+
+template <> struct nanobind::detail::type_hook<llvm::Value> {
+  static const std::type_info *get(llvm::Value *v) {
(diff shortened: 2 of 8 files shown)


projects/eudsl-llvmpy/src/IR/Kinds.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return &typeid(llvm::FunctionType);
+  case llvm::Type::PointerTyID:
+    return &typeid(llvm::PointerType);
+  case llvm::Type::StructTyID:
+    return &typeid(llvm::StructType);
+  case llvm::Type::ArrayTyID:
+    return &typeid(llvm::ArrayType);
+  case llvm::Type::FixedVectorTyID:
+  case llvm::Type::ScalableVectorTyID:
+    return &typeid(llvm::VectorType);
```
makslevental: why are these the same?

projects/eudsl-llvmpy/src/IR/Kinds.h
```diff
@@ -0,0 +1,43 @@
...
+#include <llvm/IR/Type.h>
+#include <llvm/IR/Value.h>
+
+#include <nanobind/nanobind.h>
+
+#include <typeinfo>
+
+namespace eudsl {
+const std::type_info *typeTypeInfo(llvm::Type::TypeID id);
+const std::type_info *valueTypeInfo(unsigned valueID);
```
makslevental: this is missing in this PR

projects/eudsl-llvmpy/src/IR/Kinds.cpp
makslevental: does this cover all the possible types in LLVM?

projects/eudsl-llvmpy/src/IR/Kinds.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <llvm/IR/DerivedTypes.h>
+
+namespace eudsl {
+
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base) {
+  return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
+}
+
+const std::type_info *typeTypeInfo(llvm::Type::TypeID id) {
```
makslevental, line 11: is there a way to just use `llvm::Type::TypeID` as a `std::type_info *`? maybe something hacky like a `static_cast`? not sure about how `type_info` is actually implemented (I believe for some targets it's just a string but on others it is just a pointer).

projects/eudsl-llvmpy/src/IR/Kinds.cpp
```diff
@@ -0,0 +1,38 @@
...
+  case llvm::Type::StructTyID:
+    return &typeid(llvm::StructType);
+  case llvm::Type::ArrayTyID:
+    return &typeid(llvm::ArrayType);
+  case llvm::Type::FixedVectorTyID:
+  case llvm::Type::ScalableVectorTyID:
+    return &typeid(llvm::VectorType);
+  default:
+    // Void, Label, Metadata, Token, the float kinds, TypedPointer and
+    // TargetExtType have no bound subclass; they stay llvm::Type.
```
makslevental, line 29: why don't these have bound subclasses? we should bind these too...

projects/eudsl-llvmpy/src/IR/Kinds.h
```diff
@@ -0,0 +1,43 @@
...
+/// Return `concrete` if a nanobind class is registered for it, else `base`.
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base);
+} // namespace eudsl
+
+template <> struct nanobind::detail::type_hook<llvm::Type> {
+  static const std::type_info *get(llvm::Type *t) {
+    return t ? eudsl::typeTypeInfo(t->getTypeID()) : &typeid(llvm::Type);
```
makslevental: is this the right thing to do for null pointers???

projects/eudsl-llvmpy/src/IR/Kinds.h
```diff
@@ -0,0 +1,43 @@
...
+  }
+};
+
+template <> struct nanobind::detail::type_hook<llvm::Value> {
+  static const std::type_info *get(llvm::Value *v) {
+    return v ? eudsl::valueTypeInfo(v->getValueID()) : &typeid(llvm::Value);
```
makslevental: is this the right thing to do for null pointers???

projects/eudsl-llvmpy/src/IR/Types.cpp
```diff
@@ -0,0 +1,168 @@
...
+#define EUDSL_PRIMITIVE_TYPE(pyName, getter)                                   \
+  m.def(                                                                       \
+      pyName,                                                                  \
+      [](eudsl::Context &ctx) -> llvm::Type * {                                \
+        return llvm::Type::getter(ctx.get());                                  \
+      },                                                                       \
+      "context"_a, nb::rv_policy::reference, nb::keep_alive<0, 1>())
+
+  EUDSL_PRIMITIVE_TYPE("void_t", getVoidTy);
+  EUDSL_PRIMITIVE_TYPE("label_t", getLabelTy);
```
makslevental: why do these get `_t` suffixes?

projects/eudsl-llvmpy/src/IR/Types.cpp
```diff
@@ -0,0 +1,168 @@
...
+      nb::rv_policy::reference, nb::keep_alive<0, 1>());
+  m.def(
+      "function_t",
+      [](llvm::Type *ret, std::vector<llvm::Type *> params,
+         bool varArg) -> llvm::Type * {
+        return llvm::FunctionType::get(ret, params, varArg);
+      },
+      "return_type"_a, "params"_a, "var_arg"_a = false,
+      nb::rv_policy::reference, nb::keep_alive<0, 1>());
+}
```
makslevental, line 174: instead of using `_t` here and up above just create a submodule `llvm.types` and put everything which is within `populate_types` in that submodule (and then drop the `_t`).

projects/eudsl-llvmpy/src/eudslllvm_ext.cpp
```diff
@@ -7,8 +7,10 @@
 namespace nb = nanobind;
 
 void populate_context(nb::module_ &m);
+void populate_types(nb::module_ &m);
 
 NB_MODULE(eudslllvm_ext, m) {
   m.doc() = "Hand-written nanobind bindings over the LLVM C++ IR API.";
   populate_context(m);
+  populate_types(m);
```
makslevental:
```suggestion
  populate_types(m.def_submodule("types"));
```

makslevental:
## Post-merge test review: correctness / honesty / coverage

Reviewing the tests this PR added (`tests/test_types.py`) against the bindings in `src/IR/Types.cpp` and the `type_hook` in `src/IR/Kinds.h`. Already merged, so this is follow-up material. This is a strong suite: nothing critical, no vacuous always-true assertions, exact `str()` round-trips everywhere, downcasting checked for all seven leaf classes plus the default-to-`Type` fallback, and `__eq__`/`__hash__` exercised with real cross-context uniquing. The findings below are stale comments and a set of bound-but-unexercised members.

### Important

**Stale, misleading comments that name things that don't exist** (`test_types.py:40-42` and `:70-72`). Both describe the type_hook as a future change: "added with the Type type_hook", "until the hook lands the factories return base Type objects", "once the Type type_hook makes `named_struct_t` return a StructType." In this merged PR the hook is already in Kinds.h and the downcast tests sit in the same file. The `:40-42` comment points at `test_types_downcast`, but the function is `test_types_downcast_to_concrete_classes` (`:77`), and the accessors it references actually live in `test_concrete_type_accessors` (`:103`). The `:70-72` comment writes the factory as `named_struct_t`; the binding is `named_struct`. A maintainer following either comment lands on a function that doesn't exist and a claim that's false. (This also trips the CLAUDE.md "no forward-reference comments" rule.)

**`FixedVectorType.num_elements` is never called** (Types.cpp:246). `test_concrete_type_accessors:111-112` builds a fixed vector and reads `.min_num_elements`, which resolves to the parent `VectorType` accessor. The downcast test proves the object is a `FixedVectorType` but never touches its own accessor, so binding it to the wrong method wouldn't be caught.

**`ScalableVectorType.min_num_elements` is never called** (Types.cpp:249). The scalable vector is only ever read through `.is_scalable`; its own `getMinNumElements` binding is never invoked.

**`is_pointer` and `is_label` predicates are never asserted** (Types.cpp:165, :161). `test_type_predicates:21-26` covers `is_void`, `is_sized`, `is_integer`, `is_floating_point` but not these two. `is_pointer` is the notable one, since `ptr` is used all over the suite yet the predicate itself is never checked.

**`Type.__eq__` non-Type branch is never exercised** (Types.cpp:169). Every `==`/`!=` compares two Types. Nothing does `i32 == 5` / `i32 == None`, so the `try_cast`-fails path that returns `False` is uncovered; if it were changed to raise, normal Python equality against unrelated objects would break silently.

### Minor

- `set_body(packed=True)` is untested (`:131` uses the default). The `struct` factory tests packed, but that's `StructType::get`, a different path from `setBody`.
- Single-polarity predicates: `is_void` and `is_integer` assert only True. `is_floating_point`/`is_sized` correctly check both.
- `FunctionType.is_var_arg` True case is only checked via `str()`; the property itself is asserted only False (`:124`).
- `VectorType.element_type` is never read (`ArrayType.element_type` is, at `:109`).
- The `label` and `i16` factories are never constructed anywhere in the suite.

### Well-tested
- `str()` across primitives and every derived form, including packed `<{...}>`, scalable `<vscale x ...>`, and var-arg `(..., ...)`.
- Downcasting to all seven leaf classes plus the `void`/`f64` fallback to `Type`, and an `isinstance(..., VectorType)` check (`:79-96`).
- `__eq__` equal and not-equal on real Types with cross-context uniquing, and `__hash__` via `set` dedup (`:31-34`).
- `IntegerType.bit_width`, `PointerType.address_space`, the full `StructType` surface (`name`, `num_elements`, `element_type`, `is_packed`/`is_opaque` both polarities, `set_body` default path), `ArrayType`, `FunctionType.return_type`/`num_params`/`param_type`/`params`.
- `assert_no_leaks()` on every test, guarding the `keep_alive` contract.

One note I looked into and am *not* flagging: the `assert(t)` in `type_hook<llvm::Type>::get` assumes nanobind never consults the hook with null. Within this PR that holds, since no bound accessor returns a null `Type*`. It only became a real problem later when null-returning `Value` accessors arrived (and was changed to a null guard in #539), so it's not a defect here.

PR #539, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: Value/User spine, Function, Argument, BasicBlock

makslevental:
Binds the `llvm::Value` hierarchy spine.

- The `llvm::Value` base and `llvm::User`.
- The Value class-hierarchy registration and a `reopen<T>` helper for adding methods to an already-registered class.
- `Function`, `Argument`, and `BasicBlock`, with traversal iterators (a function's blocks, a block's instructions, a function's arguments).

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Common.h
+++ b/projects/eudsl-llvmpy/src/IR/Common.h
@@ -43,13 +43,6 @@ inline void unwrap(llvm::Error &&e) {
     throw std::runtime_error(llvm::toString(std::move(e)));
 }
 
-/// Fetch an already-registered nanobind class so a later translation unit can
-/// add methods to it. Used because the Value hierarchy is registered as a spine
-/// in Values.cpp and filled in by Instructions.cpp / Constants.cpp.
-template <typename T> nb::class_<T> reopen() {
-  return nb::borrow<nb::class_<T>>(nb::type<T>());
-}
-
 } // namespace eudsl
 
 // Pulled in here so every translation unit that returns an llvm::Type* or
--- a/projects/eudsl-llvmpy/src/IR/Common.h
+++ b/projects/eudsl-llvmpy/src/IR/Common.h
@@ -43,6 +43,13 @@ inline void unwrap(llvm::Error &&e) {
     throw std::runtime_error(llvm::toString(std::move(e)));
 }
 
+/// Fetch an already-registered nanobind class so a later translation unit can
+/// add methods to it. Used because the Value hierarchy is registered as a spine
+/// in Values.cpp and filled in by Instructions.cpp / Constants.cpp.
+template <typename T> nb::class_<T> reopen() {
+  return nb::borrow<nb::class_<T>>(nb::type<T>());
+}
+
 } // namespace eudsl
 
 // Pulled in here so every translation unit that returns an llvm::Type* or
--- a/projects/eudsl-llvmpy/src/IR/Kinds.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Kinds.cpp
@@ -23,6 +23,10 @@ const std::type_info *pick(const std::type_info *concrete,
   return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
 }
 
+// `id` is an llvm::Value::getValueID() result, which is an `unsigned` rather
+// than the Value::ValueTy enum: instruction IDs are `InstructionVal + opcode`
+// and run past the ValueTy enumerators, so `unsigned` is what the LLVM API
+// hands back here.
 const std::type_info *valueTypeInfo(unsigned id) {
   const std::type_info *base = &typeid(llvm::Value);
 
--- a/projects/eudsl-llvmpy/src/IR/Kinds.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Kinds.cpp
@@ -4,9 +4,53 @@
 
 #include "IR/Kinds.h"
 
+#include <llvm/IR/Argument.h>
+#include <llvm/IR/BasicBlock.h>
+#include <llvm/IR/Constants.h>
 #include <llvm/IR/DerivedTypes.h>
+#include <llvm/IR/Function.h>
+#include <llvm/IR/GlobalAlias.h>
+#include <llvm/IR/GlobalIFunc.h>
+#include <llvm/IR/GlobalVariable.h>
+#include <llvm/IR/InlineAsm.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/IR/Metadata.h>
 
 namespace eudsl {
+
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base) {
+  return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
+}
+
+const std::type_info *valueTypeInfo(unsigned id) {
+  const std::type_info *base = &typeid(llvm::Value);
+
+  if (id >= llvm::Value::InstructionVal) {
+    switch (id - llvm::Value::InstructionVal) {
+#define HANDLE_INST(num, opcode, Class)                                        \
+  case num:                                                                    \
+    return pick(&typeid(llvm::Class), base);
+#include "llvm/IR/Instruction.def"
+    default:
+      return &typeid(llvm::Instruction);
+    }
+  }
+
+  switch (id) {
+// MemoryUse/MemoryDef/MemoryPhi are MemorySSA classes, not IR Value subclasses
+// reachable from a Module; map their enum slots to base Value.
+#define HANDLE_MEMORY_VALUE(Name)                                              \
+  case llvm::Value::Name##Val:                                                 \
+    return base;
+#define HANDLE_VALUE(Name)                                                     \
+  case llvm::Value::Name##Val:                                                 \
+    return pick(&typeid(llvm::Name), base);
+#include "llvm/IR/Value.def"
+  default:
+    return base;
+  }
+}
 
 const std::type_info *typeTypeInfo(llvm::Type::TypeID id) {
   switch (id) {
--- a/projects/eudsl-llvmpy/src/IR/Values.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Values.cpp
@@ -4,6 +4,12 @@
 
 #include "IR/Common.h"
 
+#include <llvm/IR/Constant.h>
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/GlobalObject.h>
+#include <llvm/IR/GlobalValue.h>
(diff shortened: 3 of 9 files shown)


projects/eudsl-llvmpy/src/IR/Common.h
```diff
@@ -43,6 +43,13 @@ inline void unwrap(llvm::Error &&e) {
     throw std::runtime_error(llvm::toString(std::move(e)));
 }
 
+/// Fetch an already-registered nanobind class so a later translation unit can
+/// add methods to it. Used because the Value hierarchy is registered as a spine
+/// in Values.cpp and filled in by Instructions.cpp / Constants.cpp.
+template <typename T> nb::class_<T> reopen() {
+  return nb::borrow<nb::class_<T>>(nb::type<T>());
+}
+
```
makslevental: is this actually used anywhere in the PR stack?

projects/eudsl-llvmpy/src/IR/Values.cpp
```diff
@@ -0,0 +1,155 @@
...
+      .def_prop_ro("operands", [](llvm::User &self) {
+        std::vector<llvm::Value *> out;
+        for (unsigned i = 0, n = self.getNumOperands(); i < n; ++i)
+          out.push_back(self.getOperand(i));
+        return out;
+      });
+
+  // Structural spine of the Value hierarchy. Leaf method bindings are added by
+  // Tasks 9-11, which reopen these classes via reopen<T>(). Registering them
+  // here lets the Value type_hook name them without raising.
```
makslevental: i think this comment is stale (check if reopen is actually used in the PR stack)

projects/eudsl-llvmpy/src/IR/Kinds.cpp
```diff
@@ -4,10 +4,54 @@
...
+#include <llvm/IR/Metadata.h>
 
 namespace eudsl {
 
+const std::type_info *pick(const std::type_info *concrete,
+                           const std::type_info *base) {
+  return nanobind::detail::nb_type_lookup(concrete) ? concrete : base;
+}
+
+const std::type_info *valueTypeInfo(unsigned id) {
```
makslevental, line 30: why is this just an `unsigned`?

projects/eudsl-llvmpy/src/IR/Values.cpp
```diff
@@ -0,0 +1,155 @@
...
+        return out;
+      });
+
+  nb::class_<llvm::Function, llvm::GlobalObject>(m, "Function")
+      .def_static(
+          "create",
+          [](llvm::FunctionType *ft, const std::string &name,
+             eudsl::Module &mod) {
+            return llvm::Function::Create(
+                ft, llvm::GlobalValue::ExternalLinkage, name, mod.get());
```
makslevental: why did you assume `llvm::GlobalValue::ExternalLinkage`? this should be an argument to this function and thus enable the user to select the linkage

makslevental: in general do not hardcode assumptions like this anywhere - the full LLVM APIs should be bound and exposed to the user (add this to CLAUDE.md)

projects/eudsl-llvmpy/tests/test_values.py
```diff
@@ -0,0 +1,97 @@
...
+def test_value_users_and_operands():
+    with llvm.Context() as ctx:
+        mod = llvm.parse_assembly(_SRC, ctx, "m")
+        f = mod.get_function("f")
+        x = f.arg(0)
+        # %x is used by the add.
+        assert x.num_uses == 1
+        add = x.users[0]
+        # Concrete opcode downcast (BinaryOperator) activates in Task 10; here
+        # the add arrives as a registered base (User/Instruction).
```
makslevental: remove comments like this which are forward references - do not create forward reference comments because they might go stale during the merge (add this to CLAUDE.md)

projects/eudsl-llvmpy/tests/test_values.py
```diff
@@ -0,0 +1,97 @@
...
+      %sum = add i32 %x, %y
+      ret i32 %sum
+    }
+    """
+)
+
+
+def test_value_and_user_registered():
+    # Value/User accessors only become reachable once a Value can be obtained
+    # from Python (functions()/traversal in Task 9). This test confirms
```
makslevental: remove the forward task reference here. do not add comments which are forward references

projects/eudsl-llvmpy/src/IR/Common.h
makslevental: Removed — `reopen<T>` is never called anywhere in the stack (only the definition and this stale comment referenced it). Dropped the helper.

projects/eudsl-llvmpy/src/IR/Values.cpp
makslevental: Stale, agreed — there is no `reopen<T>()` call. Rewrote the comment: the spine base classes are registered bare so the concrete subclasses in Instructions.cpp/Constants.cpp can name them as their nanobind base, and so the Value type_hook can name them.

projects/eudsl-llvmpy/src/IR/Values.cpp
makslevental: Fixed. `Function.create` now takes a `linkage` argument (default `ExternalLinkage`); I bound the `Linkage` (GlobalValue::LinkageTypes) enum with the GlobalValue hierarchy in Values.cpp so it's available where Function is bound, and added a test asserting internal linkage lands in the IR. Also added a rule to CLAUDE.md: bind the full API and expose choices as arguments rather than hardcoding them.

projects/eudsl-llvmpy/tests/test_values.py
makslevental: Removed the forward-reference comment. Added a 'no forward-reference comments' rule to CLAUDE.md.

projects/eudsl-llvmpy/tests/test_values.py
makslevental: Removed the forward (Task 9) reference.

projects/eudsl-llvmpy/src/IR/Kinds.cpp
makslevental: `valueTypeInfo` takes the result of `llvm::Value::getValueID()`, which returns `unsigned`. There's no narrower enum that covers it: instruction IDs are `InstructionVal + opcode` and run past the `Value::ValueTy` enumerators, so `unsigned` is the type the LLVM API hands back here. Left as-is; leaving this open for your call.

makslevental:
## Post-merge test review: correctness / honesty / coverage

Reviewing the tests this PR added (`tests/test_values.py`) against the bindings in `src/IR/Values.cpp` and the `type_hook` change in `src/IR/Kinds.h`. This is already merged, so these are notes for follow-up rather than blockers. The traversal and use-def parts are solid. The gap that matters is that the one correctness fix in the PR has no regression test, and a fair number of bound members are never exercised.

### Critical

**The null path of the `type_hook` fix is untested.** Kinds.h changed both `type_hook<llvm::Value>` and `type_hook<llvm::Type>` from an asserting/deref form to a null guard (`return v ? valueTypeInfo(v->getValueID()) : &typeid(llvm::Value)`). That guard exists because null-returning accessors flow back to Python as `None`: `Function.entry_block` returns `nullptr` for an empty function, and `BasicBlock.terminator` uses `getTerminatorOrNull`. Every test only hits the non-null path. `entry_block` is read on non-empty functions (`test_values.py:54`, `:89`) and `terminator` on a block that has one (`test_values.py:60`). Revert the guard to `return valueTypeInfo(v->getValueID());` and the whole suite stays green; the crash (segfault under `-DNDEBUG`, assert-abort otherwise) only shows up at runtime once a `None` is actually produced. `test_append_basic_block` (`test_values.py:84-86`) even parses `declare void @g()` and creates an empty function `h`, so `fn.entry_block is None` was one assertion away and wasn't written. This is the same shape of gap that came up on #540's ownership fix.

### Important

- **`replace_all_uses_with` has zero coverage** (Values.cpp:235). A mutating op with real semantics; a no-op or wrong-arg binding passes. The test module already gives `%x` exactly one use (`test_values.py:74`), so RAUW-then-recount is one step away.
- **`Value.name` setter is untested** (Values.cpp:226). Only the getter runs (`:34`, `:39`, `:55`). A setter that no-ops or writes the wrong field slips through, and the write path is the one the CLAUDE.md identifier/source-filename note cares about.
- **All four `Instruction` members are untested**: `num_successors`, `successor`, `is_terminator`, `parent` (Values.cpp:281-292). Concrete instructions are already in hand at `test_values.py:57`, so these are reachable on the bound base without waiting for #540. `insts[-1].is_terminator`, `insts[-1].num_successors == 0`, and `insts[0].parent == entry` are all trivial and would catch a swapped or stubbed accessor.
- **`Function.function_type` is never read** (Values.cpp:333). `return_type` is checked (`:43`) but not `function_type`; a wrong or dangling pointer wouldn't be caught.

### Minor

- `User.operands` (the list property) untested; only `operand(index)` runs (`:77`). `add.operands == [x, y]` would also pin ordering.
- `BasicBlock.create` static factory untested; blocks are only made via `append_basic_block`. The `parent=None` default and detached-block path never run.
- `Value.type` getter never read on any Value.
- `is_declaration` checked only False (`:42`); `is_var_arg` never checked. The `declare void @g()` in the module is a ready-made True case.
- `Value.__str__` and `Value.__hash__` untested (`str(mod)` is Module's own `__str__`; no Value goes into a set/dict).
- `Value.__eq__` negative branch untested: every `==` compares two Values, so the `try_cast`-fails path (`value == None` returning `False` rather than raising) never runs.
- `Linkage` exercises only EXTERNAL (default) and INTERNAL (`:101`); the other seven are declared but not round-tripped.

### Well-tested

- Function/argument traversal: `functions`, `get_function` hit and miss (`:34-37`), `num_args`, `args`, `arg`, `Argument.arg_no`, `Argument.parent` (`:38-41`).
- Block/instruction traversal: `basic_blocks`, non-null `entry_block`, `BasicBlock.name`/`parent`, `instructions` count, non-null `terminator` (`:52-60`, `:88-90`).
- Use-def spine with pointer-identity `__eq__` (`:74-77`) is the strongest part.
- `Function.create` linkage: `:101-103` checks the printed IR for `define internal`, so it confirms the argument threads through to LLVM rather than just not crashing.
- The `type(insts[0]).__name__ == "Instruction"` assertion (`:63`) is honest for this PR, since leaf Instruction classes land in #540; the comment says as much. Not a defect.
- Leak discipline: every test ends in `assert_no_leaks()` after explicit `del`.

PR #540, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: Instruction subclasses and PHINode incoming values

makslevental:
Two fixes surfaced by concrete instruction downcasting:
- type_hook is keyed on the static pointer type, so specialize it across the
  whole Value/Type hierarchy (SFINAE on is_base_of) rather than only the
  base classes; otherwise accessors returning Instruction*/User* never
  downcast.
- LLVMContext is now held by shared_ptr and kept alive by each Module, so a
  Value/Module lingering past the Context manager's __exit__ no longer
  dereferences a freed context.

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -101,6 +101,7 @@ nanobind_add_module(eudslllvm_ext
   src/IR/Types.cpp
   src/IR/Kinds.cpp
   src/IR/Values.cpp
+  src/IR/Instructions.cpp
 )
 
 set(eudslllvm_ext_libs
--- a/projects/eudsl-llvmpy/src/IR/Instructions.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Instructions.cpp
@@ -1,0 +1,162 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "IR/Common.h"
+
+#include <llvm/IR/InstrTypes.h>
+#include <llvm/IR/Instructions.h>
+
+void populate_instructions(nb::module_ &m) {
+  // ICmp and FCmp share the C++ enum llvm::CmpInst::Predicate, and nanobind
+  // keys enum registration by C++ type, so register it once and expose the two
+  // Python names as aliases.
+  nb::enum_<llvm::CmpInst::Predicate>(m, "CmpPredicate")
+      .value("EQ", llvm::CmpInst::ICMP_EQ)
+      .value("NE", llvm::CmpInst::ICMP_NE)
+      .value("UGT", llvm::CmpInst::ICMP_UGT)
+      .value("UGE", llvm::CmpInst::ICMP_UGE)
+      .value("ULT", llvm::CmpInst::ICMP_ULT)
+      .value("ULE", llvm::CmpInst::ICMP_ULE)
+      .value("SGT", llvm::CmpInst::ICMP_SGT)
+      .value("SGE", llvm::CmpInst::ICMP_SGE)
+      .value("SLT", llvm::CmpInst::ICMP_SLT)
+      .value("SLE", llvm::CmpInst::ICMP_SLE)
+      .value("OEQ", llvm::CmpInst::FCMP_OEQ)
+      .value("OGT", llvm::CmpInst::FCMP_OGT)
+      .value("OGE", llvm::CmpInst::FCMP_OGE)
+      .value("OLT", llvm::CmpInst::FCMP_OLT)
+      .value("OLE", llvm::CmpInst::FCMP_OLE)
+      .value("ONE", llvm::CmpInst::FCMP_ONE)
+      .value("UEQ", llvm::CmpInst::FCMP_UEQ)
+      .value("UNE", llvm::CmpInst::FCMP_UNE);
+  m.attr("ICmpPredicate") = m.attr("CmpPredicate");
+  m.attr("FCmpPredicate") = m.attr("CmpPredicate");
+
+  // Intermediate bases (the Value type_hook never names these, but leaf classes
+  // need them registered as their nanobind base).
+  nb::class_<llvm::UnaryInstruction, llvm::Instruction>(m, "UnaryInstruction");
+  nb::class_<llvm::UnaryOperator, llvm::UnaryInstruction>(m, "UnaryOperator");
+  nb::class_<llvm::BinaryOperator, llvm::Instruction>(m, "BinaryOperator");
+  nb::class_<llvm::CastInst, llvm::UnaryInstruction>(m, "CastInst");
+  nb::class_<llvm::FuncletPadInst, llvm::Instruction>(m, "FuncletPadInst");
+
+  nb::class_<llvm::CmpInst, llvm::Instruction>(m, "CmpInst")
+      .def_prop_ro("predicate", &llvm::CmpInst::getPredicate);
+
+  nb::class_<llvm::CallBase, llvm::Instruction>(m, "CallBase")
+      .def_prop_ro("num_args", &llvm::CallBase::arg_size)
+      .def(
+          "arg_operand",
+          [](llvm::CallBase &self, unsigned i) {
+            return self.getArgOperand(i);
+          },
+          "index"_a, nb::rv_policy::reference_internal)
+      .def_prop_ro(
+          "called_operand",
+          [](llvm::CallBase &self) { return self.getCalledOperand(); },
+          nb::rv_policy::reference_internal);
+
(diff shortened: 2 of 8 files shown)


makslevental:

Reviewed the tests this PR ships (`tests/test_instructions.py`, the `tests/test_values.py` delta) against `src/IR/Instructions.cpp`, `src/IR/Kinds.{h,cpp}` and `src/IR/Ownership.{h,cpp}`. Everything below was reproduced against a build of this stack, not inferred.

---

### 1. `.predicate` raises for half the FCmp predicates, and the test can't see it

`test_fcmp_predicate` checks `FCMP_OGT` only. Eight of the sixteen FCmp predicates have no entry in the `CmpPredicate` enum, and reading `.predicate` on them raises:

```python
# fcmp uno float %a, %b  -> ValueError: 8 is not a valid CmpPredicate.
# fcmp ugt float %a, %b  -> ValueError: 10 is not a valid CmpPredicate.
# fcmp ord float %a, %b  -> ValueError: 7 is not a valid CmpPredicate.
```

Missing: `FCMP_FALSE`, `FCMP_ORD`, `FCMP_UNO`, `FCMP_UGT`, `FCMP_UGE`, `FCMP_ULT`, `FCMP_ULE`, `FCMP_TRUE`.

Four of those are structural, not oversights. `Instructions.cpp:14-34` registers one shared `CmpPredicate` and aliases `ICmpPredicate`/`FCmpPredicate` onto it, so `UGT`/`UGE`/`ULT`/`ULE` can only carry the ICmp values; the FCmp unsigned compares have nowhere to live under that scheme. `ORD`/`UNO`/`TRUE`/`FALSE` look like plain omissions.

The shared-enum trick costs something the tests don't reveal. A parameterized test over all sixteen FCmp and all ten ICmp predicates would have caught it on the first run. It would also catch the flip side: because both aliases are the same object, `icmp.predicate == llvm.FCmpPredicate.EQ` passes too, so `test_icmp_predicate` and `test_fcmp_predicate` cannot actually distinguish the two families.

### 2. `assert_no_leaks()` is blind to the ownership change this PR makes

`Context::release()` (`Ownership.cpp:19-24`) decrements `gLiveContexts` the moment `__exit__` runs, without regard to the shared_ptr's use count. So the counter tracks "the Python wrapper let go", not "the LLVMContext died".

Concretely, in `test_module_outlives_released_context`:

```python
with llvm.Context() as ctx:
    mod = llvm.parse_assembly(_SRC, ctx, "m")
assert llvm.Context._get_live_count() == 0   # already 0 at __exit__
...
del mod
gc.collect()
assert_no_leaks()                             # cannot fail
```

The final `assert_no_leaks()` is vacuous. The count reached 0 before `del mod`, so `del mod` and `gc.collect()` are decorative. If a Module (or a reference cycle through one) retained the `shared_ptr<LLVMContext>` forever, every test in the suite would stay green.

The test proves the direction it set out to prove: the context survives `__exit__`. Nothing in the suite proves the context is *eventually* destroyed, which is the new risk shared ownership introduces. Worth exposing the other half of the pair, e.g. a `_get_live_llvm_contexts()` backed by its own counter in an LLVMContext-owning wrapper, or a `use_count()` probe, so over-retention is observable.

### 3. `Module::take()` escapes the ownership guarantee, untested

`Ownership.h:44-46` states the module "plus a shared reference to the LLVMContext it was built in, so destruction order is safe regardless of when the Context manager exits." `take()` (`Ownership.cpp:49-52`) moves the `unique_ptr<llvm::Module>` out and leaves `ctxKeepAlive` behind on the wrapper. Once the wrapper is collected, the taken `llvm::Module` holds a reference to an `LLVMContext` nothing keeps alive.

`_take()` currently resets immediately, so it isn't observable from Python today, but this is the one path where the invariant the PR just established does not hold, and it's the path the JIT/linker PRs later in the stack will use. No test covers released-context then `_take()`. Either move `ctxKeepAlive` into the returned value's lifetime or note the carve-out in the comment.

### 4. Downcast coverage is 12 of 57 registered classes, and a gap degrades silently

`Instructions.cpp` registers 57 `nb::class_`. The tests exercise twelve: `PHINode`, `ICmpInst`, `FCmpInst`, `CondBrInst`, `UncondBrInst`, `ReturnInst`, `AllocaInst`, `LoadInst`, `StoreInst`, `GetElementPtrInst`, `CallInst`, plus `BinaryOperator` in `test_values.py`.

That would be fine if a missing registration failed loudly. It doesn't. `valueTypeInfo` sets `base = &typeid(llvm::Value)` and every instruction arm returns `pick(&typeid(llvm::Class), base)`, so an unregistered leaf comes back as base `Value`, losing `.opcode`, `.parent` and the whole Instruction surface. Only the unknown-opcode `default:` arm falls back to `Instruction`. A rename in `Instruction.def` or a dropped `nb::class_` line is invisible to the current tests.

The registrations are all correct right now (I checked `fneg` → `FPUnaryOperator`, `fadd` → `FPBinaryOperator`, `trunc` → `TruncInst`, `select` → `SelectInst`, `freeze` → `FreezeInst`, `zext` → `ZExtInst`). Nothing pins that. A table-driven test asserting `hasattr(llvm, Class)` for every `Class` in `Instruction.def` is cheap and would hold the invariant that `Kinds.h`'s own header comment declares.

Untested-but-registered, ranked by how likely they are to be hit: `FPBinaryOperator`/`FPUnaryOperator` (every float arithmetic op), the thirteen-class cast family, `SelectInst`, `SwitchInst`, `InvokeInst`, `FreezeInst`, then the funclet/EH set.

### 5. `test_phi_add_incoming` leaves the module failing verification

```
VerifyError: PHINode should have one entry for each predecessor of its parent basic block!
  %p = phi i32 [ 1, %a ], [ 2, %b ], [ 1, %entry ]
```

`entry` branches to `%a` and `%b`, so it is not a predecessor of `%join`. The comment explains at length why the *type* is safe:

```python
# add_incoming(value, block) is the write path. Feed back an existing
# i32 incoming value (the phi's own operand 0) so the added type matches
# the phi type; a debug LLVM would assert on a mismatch.
```

and is silent about the CFG it invalidates, which reads as if the mutation is well-formed. The test also only reads back through the same three accessors it wrote through, so nothing confirms the IR itself. Use a fixture where `%join` has a third real predecessor, then assert on `str(mod)` (and `mod.verify()`, once that lands) rather than only on `incoming_value(2)`/`incoming_block(2)`.

### 6. Comment rot in `test_instructions_downcast`

```python
# add-style ops become BinaryOperator; the icmp becomes ICmpInst; the
# phi becomes PHINode; the conditional branch becomes CondBrInst.
```

`_PHI_SRC` contains no binary operator, and the test asserts nothing about `BinaryOperator`. Drop the first clause or add an `add` to the fixture and the matching assertion.

---

### Smaller points

- **Bounds tests use index 99.** The guard is `i >= self.arg_size()` / `i >= self.getNumIncomingValues()`. Index 99 exercises nothing about the comparison; the boundary case is `i == num`. A `>` typo would pass every current test. `phi.incoming_value(phi.num_incoming)` and `call.arg_operand(call.num_args)` are the indices that carry signal. (I confirmed index 2 raises correctly today.)

- **`ret void` → `return_value is None` is untested.** It returns `None` correctly. It is the only new null-returning accessor in the PR, and the PR reworks exactly the null branch of the `type_hook`, so it is the natural regression test for that change.

- **`Kinds.h`'s null branch is unguarded.** The header states the invariant that every returned `type_info` must name a nanobind-registered class. The non-null branch enforces it with `pick()`; the null branch returns `&typeid(T)` raw. Every `T` reachable on this branch is registered, so it's latent. It stops being latent the first time an accessor with a static return type of `llvm::Constant*` or `llvm::GlobalValue*` lands before that class is registered, and it will fail only on the null path. Routing the null branch through `pick(&typeid(T), &typeid(llvm::Value))` closes it.

- **`is_conditional` assertions are tautological.** `UncondBrInst.is_conditional` is a lambda returning `false` and `CondBrInst.is_conditional` returns `true`. `_insts_by_class` already matched on class name, so `assert ubrs[0].is_conditional is False` cannot fail independently. The `num_successors` assertions next to them do carry signal, since those call real LLVM code.

- **"until the last Module is gone."** `Ownership.h:16-20` describes multi-module sharing; only the one-module case is tested. Two modules in one context, release, drop one, dereference the other is the case the wording promises.

### What's good

- `test_memory_accessors` asserting `type(load.pointer_operand).__name__ == "AllocaInst"` is exactly right. It tests that the `type_hook` fires through an accessor's return value rather than only through a traversal, which is the actual claim of the `Kinds.h` change.
- `_insts_by_class` / `_only` keeps the tests readable, and `_only`'s tuple-unpack asserts uniqueness for free.
- Every test releases its handles and ends in `assert_no_leaks()`, consistently.
- The `_MEM_SRC` comment about `%gep`/`%fc` being intentionally unused is accurate and saves the next reader a detour.
- `test_values.py`'s `BinaryOperator` assertion correctly replaces the now-stale `"Instruction"` expectation instead of leaving it loosened.

PR #541, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: Constant subclasses and constant construction

makslevental:
Binds the `llvm::Constant` subclasses and constant construction: ConstantInt and ConstantFP, the aggregate and data constants, ConstantExpr, and the global-value constants. `const_int` guards against a value that doesn't fit the target integer type's bit width (interpreted per `signed`), which would otherwise trip an APInt assertion and abort the process; it now raises `ValueError` instead.

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -102,6 +102,7 @@ nanobind_add_module(eudslllvm_ext
   src/IR/Kinds.cpp
   src/IR/Values.cpp
   src/IR/Instructions.cpp
+  src/IR/Constants.cpp
 )
 
 set(eudslllvm_ext_libs
--- a/projects/eudsl-llvmpy/src/IR/Constants.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Constants.cpp
@@ -63,8 +63,20 @@ void populate_constants(nb::module_ &m) {
       "const_int",
       [](llvm::Type *ty, int64_t value, bool isSigned) -> llvm::Constant * {
         auto *ity = llvm::cast<llvm::IntegerType>(ty);
-        return llvm::ConstantInt::get(ity, static_cast<uint64_t>(value),
-                                      isSigned);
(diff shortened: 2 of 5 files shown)


makslevental:

Reviewed `tests/test_constants.py` against the `Constants.cpp` bindings. Updated after weekend refactors — several originally-flagged gaps are now covered.

---

### ~~1. `const_int` guard~~ — FIXED

The range guard now exists (`isIntN`/`isUIntN` check at `Constants.cpp:89-96`, throws `nb::value_error`). `test_const_int_out_of_range_raises` covers both signed and unsigned overflow plus the negative-with-unsigned case.

### ~~2. `zext_value` untested~~ — FIXED

`test_const_int` now asserts `neg.zext_value == 4294967295`.

### ~~3. `GlobalVariable.is_constant` and `.initializer` untested~~ — FIXED

`test_global_variable_is_constant_and_initializer` covers both polarities of `is_constant` and both arms of the initializer null-guard.

### 4. Registered constant class coverage is still incomplete (Suggestion)

Originally 15 of ~20 leaves untested. Now covers `ConstantDataArray`, `ConstantAggregateZero`, `ConstantDataVector`, `ConstantStruct`, `ConstantExpr` in addition to the original 5. Still untested: `ConstantArray`, `ConstantVector`, `ConstantTokenNone`, `BlockAddress`, `GlobalAlias`, `GlobalIFunc`. (`BlockAddress` and `GlobalIFunc` are harder to materialize from textual IR, so the practical gap is smaller than originally stated.)

### ~~5. `const_null` alias untested~~ — FIXED

`assert llvm.null is llvm.const_null` now present.

---

### What's good

- The `type(c).__name__ == ...` downcast assertions are genuinely meaningful — an unresolvable class degrades to base `Value`, so these fail loudly on a registration regression.
- `test_undef_poison_null` correctly distinguishes `PoisonValue` from its base `UndefValue`, verifying the leaf-most downcast wins.
- `test_const_int` exercises the signed-negative round-trip and the range guard.
- Tests are readable, each scoped to its own context block.

PR #545, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: PassBuilder pipeline execution via run_passes

makslevental:
Binds `run_passes`, which runs a textual PassBuilder pipeline over a module. An unknown pass name raises a `RuntimeError` rather than silently doing nothing.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -1,0 +1,33 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "IR/Common.h"
+#include "IR/Ownership.h"
+
+#include <llvm/IR/PassManager.h>
+#include <llvm/Passes/PassBuilder.h>
+
+void populate_passes(nb::module_ &m) {
+  m.def(
+      "run_passes",
+      [](eudsl::Module &mod, const std::string &pipeline) {
+        llvm::PassBuilder pb;
+        llvm::LoopAnalysisManager lam;
+        llvm::FunctionAnalysisManager fam;
+        llvm::CGSCCAnalysisManager cgam;
+        llvm::ModuleAnalysisManager mam;
+        pb.registerModuleAnalyses(mam);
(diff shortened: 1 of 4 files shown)


projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+void populate_passes(nb::module_ &m) {
+  m.def(
+      "run_passes",
+      [](eudsl::Module &mod, const std::string &pipeline) {
+        llvm::PassBuilder pb;
+        llvm::LoopAnalysisManager lam;
+        llvm::FunctionAnalysisManager fam;
+        llvm::CGSCCAnalysisManager cgam;
+        llvm::ModuleAnalysisManager mam;
```
makslevental, line 19: do any of these have useful options which can be set? if so bind them as optional args

makslevental: Still relevant. The current binding takes only a flat pipeline string (`"instcombine,simplifycfg"`). The new pass manager supports pass-level options via the textual syntax (e.g., `"loop-unroll<full-unroll-max=8>"`), which already works through the string interface. However, `PassBuilder` also supports registering custom pipelines and extension points (`registerPipelineParsingCallback`, `registerOptimizerLastEPCallback`) — those are not exposed. For typical use the string form is sufficient; the extension points would matter if users want to inject custom passes.

makslevental: try to mirror what's bound on the mlir python bindings (debug, print_ir_before/after, etc)

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -30,4 +49,31 @@ void populate_passes(nb::module_ &m) {
...
       "module"_a, "pipeline"_a,
       "Parse and run a textual pass pipeline over the module in place.");
+
+  m.def(
+      "run_default_pipeline",
+      [](eudsl::Module &mod, llvm::OptimizationLevel level,
+         nb::object pto_obj) {
+        llvm::PipelineTuningOptions pto;
+        if (!pto_obj.is_none())
+          pto = nb::cast<llvm::PipelineTuningOptions>(pto_obj);
```
makslevental: just use `std::optional<llvm::PipelineTuningOptions> ptoObj`

PR #546, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: Target, TargetMachine, DataLayout, asm and object emission

makslevental:
Binds the codegen target surface: `Target`, `TargetMachine`, and `DataLayout`, host-triple lookup, and assembly and object emission from a module.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -17,6 +17,13 @@
 #include <llvm/MC/TargetRegistry.h>
 #include <llvm/Support/MemoryBuffer.h>
 #include <llvm/Support/SourceMgr.h>
+#include <llvm/Target/TargetMachine.h>
+
+namespace eudsl {
+struct TargetMachine {
+  std::unique_ptr<llvm::TargetMachine> tm;
+};
+} // namespace eudsl
 
 #include <nanobind/stl/string.h>
 #include <nanobind/stl/vector.h>
@@ -129,11 +136,8 @@ void populate_context(nb::module_ &m) {
       })
       .def(
           "set_data_layout_from",
-          [](eudsl::Module &self, nb::handle tm) {
-            // TargetMachine lives in Target.cpp; fetch its data-layout string
-            // through the bound property to avoid a cross-file C++ dependency.
-            std::string dl = nb::cast<std::string>(tm.attr("data_layout_str"));
-            self.get().setDataLayout(dl);
+          [](eudsl::Module &self, eudsl::TargetMachine &tm) {
+            self.get().setDataLayout(tm.tm->createDataLayout());
           },
           "target_machine"_a);
 
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -126,7 +126,16 @@ void populate_context(nb::module_ &m) {
         llvm::WriteBitcodeToFile(self.get(), os);
         os.flush();
         return nb::bytes(buf.data(), buf.size());
-      });
+      })
+      .def(
+          "set_data_layout_from",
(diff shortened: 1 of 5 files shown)


projects/eudsl-llvmpy/src/IR/Context.cpp
```diff
@@ -112,7 +112,16 @@ void populate_context(nb::module_ &m) {
...
         os.flush();
         return nb::bytes(buf.data(), buf.size());
-      });
+      })
+      .def(
+          "set_data_layout_from",
+          [](eudsl::Module &self, nb::handle tm) {
+            // TargetMachine lives in Target.cpp; fetch its data-layout string
+            // through the bound property to avoid a cross-file C++ dependency.
+            std::string dl = nb::cast<std::string>(tm.attr("data_layout_str"));
```
makslevental, line 135: very weird - a cross-file c++ dependency is fine. ie just do `const llvm::TargetMachine& tm` (see my comment below about binding `llvm::TargetMachine` directly)

projects/eudsl-llvmpy/src/IR/Target.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <llvm/TargetParser/Host.h>
+#include <llvm/TargetParser/Triple.h>
+
+#include <memory>
+#include <optional>
+
+namespace {
+struct TM {
+  std::unique_ptr<llvm::TargetMachine> tm;
+};
```
makslevental, line 22: why is this a struct that contains a TargetMachine field instead of just bindings `TargetMachine` itself???

projects/eudsl-llvmpy/src/IR/Target.cpp
```diff
@@ -0,0 +1,81 @@
...
+              throw std::runtime_error(err);
+            llvm::TargetOptions opts;
+            llvm::TargetMachine *tm = target->createTargetMachine(
+                tt, cpu, features, opts, std::nullopt);
+            if (!tm)
+              throw std::runtime_error("could not create TargetMachine for " +
+                                       tripleStr);
+            new (self) TM{std::unique_ptr<llvm::TargetMachine>(tm)};
+          },
+          "triple"_a = "", "cpu"_a = "", "features"_a = "")
```
makslevental, line 60: use none defaults instead of empty strings here. also features should be a list of feature instead of a comma separated string

projects/eudsl-llvmpy/src/IR/Context.cpp
makslevental: Still relevant. `Context.cpp:135` takes a `TM&` (the wrapper struct) rather than `const llvm::TargetMachine&`. Since nanobind handles cross-translation-unit type registration fine, this indirection is unnecessary — `set_data_layout_from` should accept `llvm::TargetMachine&` directly.

projects/eudsl-llvmpy/src/IR/Target.cpp
makslevental: Still relevant. `struct TM { std::unique_ptr<llvm::TargetMachine> tm; }` is still the pattern. Nanobind can bind `llvm::TargetMachine` directly with a custom destructor or via `std::unique_ptr` holder — no need for a wrapper struct. The wrapper adds an indirection (`self.tm->` everywhere) and hides the real type from nanobind's type system, preventing other bindings from accepting `TargetMachine&` directly.

projects/eudsl-llvmpy/src/IR/Target.cpp
makslevental: Still relevant. The `create_target_machine` binding still uses empty string defaults (`cpu=""`, `features=""`) and takes `features` as a single comma-separated string. Should use `nb::none()` defaults (so the caller can distinguish "not specified" from "explicitly empty") and accept `features` as a `list[str]` that gets joined internally with `","`. This matches the ergonomics users expect from Python APIs.

PR #548, opened 2026-08-07
[eudsl-llvmpy] C++ object layer: ORC LLJIT with module add, lookup, and execution

makslevental:
add_module moves the module into a fresh JIT-owned context via a bitcode
round-trip, since the Context/Module lifetime model holds the LLVMContext by
shared_ptr and cannot hand out a unique_ptr<LLVMContext>. The source module is
marked consumed.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/JIT.cpp
+++ b/projects/eudsl-llvmpy/src/IR/JIT.cpp
@@ -16,26 +16,14 @@
 #include <memory>
 #include <string>
 
-namespace {
-struct JIT {
-  std::unique_ptr<llvm::orc::LLJIT> jit;
-};
-} // namespace
-
 void populate_jit(nb::module_ &m) {
-  nb::class_<JIT>(m, "LLJIT")
-      .def("__init__",
-           [](JIT *self) {
-             auto jit = eudsl::unwrap(llvm::orc::LLJITBuilder().create());
-             new (self) JIT{std::move(jit)};
-           })
+  nb::class_<llvm::orc::LLJIT>(m, "LLJIT")
+      .def(nb::new_([]() -> llvm::orc::LLJIT * {
+             return eudsl::unwrap(llvm::orc::LLJITBuilder().create()).release();
+           }))
       .def(
           "add_module",
-          [](JIT &self, eudsl::Module &mod) {
-            // Move the module into the JIT by round-tripping through bitcode
-            // into a fresh, JIT-owned LLVMContext. This avoids extracting a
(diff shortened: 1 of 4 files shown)


projects/eudsl-llvmpy/src/IR/JIT.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <llvm/IR/Module.h>
+#include <llvm/Support/MemoryBuffer.h>
+
+#include <memory>
+#include <string>
+
+namespace {
+struct JIT {
+  std::unique_ptr<llvm::orc::LLJIT> jit;
+};
```
makslevental, line 22: again why do this: why not just bind `llvm::orc::LLJIT` directly?

makslevental: Still relevant. Same pattern as PR #546: `struct JIT { std::unique_ptr<llvm::orc::LLJIT> jit; }` wraps the real type. Nanobind can hold a `std::unique_ptr<LLJIT>` directly without the intermediary struct. The wrapper hides the real type from the type system and forces `self.jit->` indirection everywhere.

projects/eudsl-llvmpy/src/IR/JIT.cpp
```diff
@@ -0,0 +1,49 @@
...
+              llvm::WriteBitcodeToFile(mod.get(), os);
+            }
+            auto ctx = std::make_unique<llvm::LLVMContext>();
+            std::unique_ptr<llvm::MemoryBuffer> memBuf =
+                llvm::MemoryBuffer::getMemBufferCopy(buf, "<jit>");
+            std::unique_ptr<llvm::Module> cloned = eudsl::unwrap(
+                llvm::parseBitcodeFile(memBuf->getMemBufferRef(), *ctx));
+            llvm::orc::ThreadSafeModule tsm(std::move(cloned), std::move(ctx));
+            eudsl::unwrap(self.addIRModule(std::move(tsm)));
+            (void)mod.take();
```
makslevental, line 39: does this also decref the python object? alternatively does `eudsl::Module` warn/error if someone tries to use the python wrapper after `add_module`?

makslevental: Addressed. `mod.take()` marks the module consumed, and `Module::get()` throws `RuntimeError("module has been consumed...")` on any subsequent access. The new `test_consumed_module_raises_on_use` test proves both `str(mod)` and `jit.add_module(mod)` raise after consumption.

PR #565, opened 2026-08-15
[eudsl-llvmpy] Fix lazy-view use-after-free by pinning the owning module

makslevental:
Fixes the use-after-free where a Python wrapper returned by a traversal accessor could outlive the `Module` that owns its storage (dropping the module while holding the wrapper freed the storage → segfault / silent UAF). Follows MLIR's owner-reference model (`PyObjectRef` chain rooted at the storage owner, `mlir/lib/Bindings/Python/IRCore.cpp`).

**Commit 1 — Sequence views.** Each `Sequence<T>` view (`operands`/`users`/`uses`/`instructions`/`args`/`basic_blocks`/`functions`) stores a strong `nb::object` reference to the owning `Module`, and every element yielded by `__getitem__` is `keep_alive`'d to that owner. Resolving the owner from a bare `llvm::Value` uses a new `llvm::Module* → eudsl::Module*` registry (populated in the `Module` ctors, erased in the dtor) + `nb::find`.

**Commit 2 — all other accessors + iterators (general fix).**
- `get_function`/`get_global_variable` were `rv_policy::reference` (pinned nothing), breaking the `reference_internal` chain every downstream scalar accessor (`arg`, `entry_block`, `operand`, `.parent`, `.terminator`, `initializer`, `aliasee`, `resolver`, `block_address`, `use.user`, …) relies on. Made them `reference_internal` so the whole scalar graph roots at the owning `Module`.
- `make_iterator` yielded elements with no keep-alive; passing `nb::keep_alive<0,1>` to each `make_iterator` pins element→iterator, which combined with the existing keep-alive on the `__iter__` def (iterator→container) roots iterated elements at the module too.

Values with no owning module (e.g. constants) get a `none` owner and are unchanged; their lifetime rides the `reference_internal` chain to whatever module-owned parent produced them.

**Tests.** `test_lazy_views.py::test_nested_view_outlives_module` (previously a segfault) plus `test_value_lifetimes.py` — for **every** Value-returning accessor and **every** iterator, hold only the result, drop the module, assert the module stays live (`== 1`) *before* touching the result, then that dropping it releases the module (`== 0`). The count assertion precedes any dereference, so a non-pinning accessor fails cleanly rather than segfaulting. Full suite: 340 passed, leak gate green.

Split out of #564 (this is an ownership fix, unrelated to the `nb::is_generic` types port).

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -117,6 +117,8 @@ void populate_context(nb::module_ &m) {
               std::advance(it, i);
               return &*it;
             };
+            // Pin the module itself: it owns the functions' storage.
+            seq.owner = nb::find(self);
             return seq;
           },
           nb::keep_alive<0, 1>())
--- a/projects/eudsl-llvmpy/src/IR/Context.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Context.cpp
@@ -141,7 +141,7 @@ void populate_context(nb::module_ &m) {
           [](eudsl::Module &self) {
             return nb::make_iterator<nb::rv_policy::reference>(
                 nb::type<eudsl::Module>(), "FunctionIterator",
-                self.get().begin(), self.get().end());
+                self.get().begin(), self.get().end(), nb::keep_alive<0, 1>());
           },
           nb::keep_alive<0, 1>())
       .def(
@@ -149,13 +149,13 @@ void populate_context(nb::module_ &m) {
           [](eudsl::Module &self, const std::string &name) {
             return self.get().getFunction(name);
           },
-          "name"_a, nb::rv_policy::reference)
+          "name"_a, nb::rv_policy::reference_internal)
       .def(
           "get_global_variable",
           [](eudsl::Module &self, const std::string &name) {
             return self.get().getNamedGlobal(name);
           },
-          "name"_a, nb::rv_policy::reference)
+          "name"_a, nb::rv_policy::reference_internal)
       .def(
           "add_named_metadata",
           [](eudsl::Module &self, const std::string &name, llvm::MDNode *node) {
--- a/projects/eudsl-llvmpy/src/IR/Ownership.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Ownership.cpp
@@ -5,12 +5,27 @@
 #include "IR/Ownership.h"
 
 #include <stdexcept>
+#include <unordered_map>
 #include <vector>
 
 namespace eudsl {
 
 static int64_t gLiveContexts = 0;
 static int64_t gLiveModules = 0;
+
+/// Maps each live llvm::Module to its owning eudsl::Module wrapper, so a view
+/// built from a bare llvm::Value can find the wrapper (and pin its Python
+/// object) via moduleWrapperFor.
+static std::unordered_map<const llvm::Module *, Module *> &liveModuleMap() {
+  static std::unordered_map<const llvm::Module *, Module *> map;
+  return map;
+}
+
+Module *moduleWrapperFor(const llvm::Module *m) {
+  auto &map = liveModuleMap();
+  auto it = map.find(m);
+  return it == map.end() ? nullptr : it->second;
+}
 
 Context::Context() : ctx(std::make_shared<llvm::LLVMContext>()) {
   ++gLiveContexts;
@@ -49,15 +64,23 @@ llvm::LLVMContext &Context::get() const {
 Module::Module(const std::string &name, Context &ctx)
     : ctxKeepAlive(ctx.shared()),
       mod(std::make_unique<llvm::Module>(name, ctx.get())), owner(&ctx) {
+  keyMod = mod.get();
+  liveModuleMap()[keyMod] = this;
   ++gLiveModules;
 }
 
 Module::Module(std::unique_ptr<llvm::Module> m, Context &ctx)
     : ctxKeepAlive(ctx.shared()), mod(std::move(m)), owner(&ctx) {
+  keyMod = mod.get();
+  liveModuleMap()[keyMod] = this;
   ++gLiveModules;
 }
 
-Module::~Module() { --gLiveModules; }
+Module::~Module() {
+  if (keyMod)
+    liveModuleMap().erase(keyMod);
+  --gLiveModules;
+}
 
 int64_t Module::liveCount() { return gLiveModules; }
 
--- a/projects/eudsl-llvmpy/src/IR/Ownership.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Ownership.cpp
@@ -13,9 +13,19 @@ namespace eudsl {
 static int64_t gLiveContexts = 0;
 static int64_t gLiveModules = 0;
 
-/// Maps each live llvm::Module to its owning eudsl::Module wrapper, so a view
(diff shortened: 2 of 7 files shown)


makslevental:
## Review — code + test quality (correctness, coverage, honesty)

Multi-agent review (code, tests, comments) of this PR's diff (`eudsl-llvmpy-types-generic...eudsl-llvmpy-view-lifetime`). The core mechanism is sound and the tests are, refreshingly, **not** cheating passes — the `_get_live_module_count() == 1` signal was verified real (see Strengths). But there are two Critical issues, one of which reopens the exact UAF class this PR closes, plus honesty gaps in the test "touch" checks.

### Critical

**1. The `none`-owner path is an untested, still-shipping use-after-free.** `Sequence.__getitem__` element-pinning and `ownerObjectFor` (`Values.cpp:30-49`) derive the pin from the **container's** value and return `none` for a value with no module (constants). The comment assumes a none-owner container only yields context-owned elements — false. `ConstantExpr.operands` is a reachable `Sequence` with `owner == none` whose elements can be **module-owned** `GlobalVariable`s. Probed:
```
add.operands[1] -> ConstantExpr (ptrtoint (ptr @g to i32));  Sequence.owner == none
ce.operands[0]  -> GlobalVariable @g            # module-owned
del add, ce, mod; gc.collect()
_get_live_module_count() == 0                   # element NOT pinned
gv.name == ''                                   # freed storage read back -> UAF
```
So the suite's headline claim ("**any** traversal accessor keeps the owning module alive") is contradicted by a path it never exercises, and that path is a real UAF, not just a coverage hole. Fix should pin from the **element's** owning module (not the container's), and add a parametrized case holding a module-owned element yielded by a none-owner container (today it fails at count `0`).

**2. Unguarded dtor `erase` → stale-key clobber (ABA) after `take()`, reopening the UAF.** `Module::~Module` (`Ownership.cpp:79-83`) does `liveModuleMap().erase(keyMod)` without checking the entry still maps to `this`. `take()` decouples wrapper lifetime from the `llvm::Module`:
1. Module `A` wraps `llvm::Module` at `X` → `map[X] = A`.
2. `A.take()` hands the module to the JIT; `A` stays alive, `keyMod == X`.
3. JIT frees that module; `X` is freed.
4. A new `llvm::Module` is allocated at `X`, wrapped by `B` → ctor `map[X] = B`.
5. `A` destroyed → `erase(X)` **removes B's entry**.
6. `moduleWrapperFor(X)` returns null → `ownerObjectFor` → `none` → `B`'s views/elements unpinned → UAF.

Fix — guard the erase:
```cpp
auto &map = liveModuleMap();
auto it = map.find(keyMod);
if (it != map.end() && it->second == this)
  map.erase(it);
```
(`moduleWrapperFor`/`erase` only hash/compare the pointer, never deref it, so the stale key itself isn't a UAF — the clobber is.)

### Important

**3. `take()` semantic gap (pinning the wrapper ≠ keeping storage alive).** After `take()`/`_take()` the JIT owns the `llvm::Module`; a held wrapper obtained earlier still points into it, and `ownerObjectFor` pins the (now-consumed) wrapper `A`, which no longer owns storage. Probe: `f = mod.get_function("f"); mod._take(); f.name` → `''` (silent UAF). Untested (nothing in `tests/jit/` either), and the `test_value_lifetimes.py` docstring's absolute claim ("holding it after the module handle is dropped is never a use-after-free") oversells what's verified. Either add a consumed-module lifetime test documenting the intended behavior, or narrow the claim.

**4. The `touch` lambdas can't detect a UAF — several are near-tautological.** Reading freed LLVM storage doesn't segfault; it returns plausible garbage (`f.name` → `''` after free). So `bool(str(v))` / `bool(v.name)` are not "still-valid-after-drop" checks (the real protection is the `== 1` count assertion, which correctly precedes the touch). `assert str(held) is not None` (`test_value_lifetimes.py`, iterator test) is *always* true — `str()` never returns `None`. The touch's genuine value is confirming downcast/type while storage is valid; strengthen the weak ones (`bb_instructions_view`, `inst_operand`, `user_getitem`) to assert concrete content (a name / type / operand index) and drop the `is not None`.

**5. Free-threading: unsynchronized static map.** `liveModuleMap()` (`Ownership.cpp:19-28`) is a plain `unordered_map` mutated in both `Module` ctors + dtor and read in `moduleWrapperFor`; the module is built `Py_MOD_GIL_NOT_USED`. nanobind's own `keep_alive` shards+locks under free-threading; this map (and the pre-existing `gLiveModules`/`gLiveContexts` counters) don't. Consistent with an already-documented "thread-safety not verified" caveat, but the PR widens the unprotected mutable-global surface — add a mutex or document it alongside the counters.

**6. Registry comments overclaim liveness + ignore `take()`.** `Ownership.cpp:16-18` ("Maps each **live llvm::Module**…") and `Ownership.h:81-84` ("The **live eudsl::Module wrapping** `m`") — the map is keyed by `keyMod` fixed at construction and only erased in the dtor, never updated by `take()`. After a module is consumed the entry is a stale wrapper/key. Reword to describe wrapper-lifetime keying and the `take()` staleness (ties into #2).

### Suggestions

- **`nb::detail::keep_alive`** (`Sequence.h:43`) is an internal API — acceptable (it's `NB_CORE`-exported and what `reference_internal` uses), but it's the only `detail::` reach-in; a one-line note that no public equivalent exists for the dynamic case would justify it.
- **`get_function`/`get_global_variable` null path untested** — `reference_internal` on a `nullptr` return is verified safe (nanobind yields `None`, `keep_alive` no-ops on `Py_None`), but add `assert mod.get_function("missing") is None` so a future regression is caught.
- **`Sequence.h` doc** (`14-20`) now presents `owner` as the sole anchor but the accessors still `keep_alive<0,1>` the view to its immediate parent — note both. **`ownerObjectFor` doc** (`Values.cpp:30-32`) reads as "constant ⇔ none"; broaden to "…or when no owning module wrapper can be resolved."
- **Cross-context and `replace_all_uses_*` (RAUW)** paths are untested and touch the same owner-resolution logic.

### Strengths (verified, not assumed)

- **The anti-cheating design holds.** Probed: dropping `mod` while holding *nothing* (or only a copied `.name` string) → count `0`, so `ctx`/the `with` block don't independently pin. Reverting `get_function` to `reference`, or removing the `make_iterator` keep_alive, makes the assertions fail **cleanly** (count `0`) *before* any dereference — real signal, no segfault-masking. The count itself is honest (`gLiveModules` tracks actual wrapper dtor).
- **assert-count-before-touch** is the correct load-bearing pattern; consistent explicit `gc.collect()` before every live-count read (per CLAUDE.md); no fragile gc-timing reliance.
- Broad matrix (~27 scalar + 4 iterator cases, stable ids); `test_nested_view_outlives_module` meaningfully checks a view **and** an extracted element pin independently (`type(op0).__name__ == "Argument"`); `test_view_reflects_live_mutation` truly exercises laziness (len 0→1 on a view captured before mutation, would fail on an eager snapshot).
- Mechanism confirmed correct: `make_iterator` `keep_alive<0,1>` → `__next__` (element→iterator) composes with the outer `__iter__` keep_alive (iterator→container) — full chain, not redundant; `reference_internal` null-return safe; `keep_alive` nurse/patient order right; `ownerObjectFor` `dyn_cast` ladder complete for module-owned Values (constants correctly fall through). Style/CLAUDE.md conformant; no forward-reference comments.

*Automated multi-agent review via Claude Code. (A fourth, silent-failure pass did not return in time; its focus — the silent `none`-owner reversion to UAF — is covered by Critical #1.)*

PR #579, opened 2026-08-17
[eudsl-llvmpy] Add MIR submodule foundation + LowLevelType (LLT) bindings

makslevental:
Introduce the llvm.mir submodule (src/MIR/, populate_mir) and link the
CodeGen libraries needed for Machine IR: LLVMCodeGenTypes, LLVMGlobalISel,
LLVMSelectionDAG, LLVMMIRParser. Bind LowLevelType (LLT) -- the
target-independent generic-MIR type -- as the first MIR class; it also proves
the new libraries link. Extend the C++ coverage gate to cover src/MIR.

First PR in the MIR bindings/DSL stack.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -39,12 +39,30 @@ void populate_mir(nb::module_ &m) {
                    [](const llvm::LLT &self) { return self.isPointer(); })
       .def_prop_ro("is_vector",
                    [](const llvm::LLT &self) { return self.isVector(); })
+      .def_prop_ro("is_integer",
+                   [](const llvm::LLT &self) { return self.isInteger(); })
+      .def_prop_ro("is_float",
+                   [](const llvm::LLT &self) { return self.isFloat(); })
       .def_prop_ro("is_valid",
                    [](const llvm::LLT &self) { return self.isValid(); })
-      .def("__eq__", [](const llvm::LLT &self,
-                        const llvm::LLT &other) { return self == other; })
-      .def("__ne__", [](const llvm::LLT &self,
-                        const llvm::LLT &other) { return self != other; })
+      .def("__eq__",
+           [](const llvm::LLT &self, nb::handle other) {
+             llvm::LLT o;
+             if (!nb::try_cast<llvm::LLT>(other, o))
+               return false;
+             return self == o;
+           })
+      .def("__ne__",
+           [](const llvm::LLT &self, nb::handle other) {
+             llvm::LLT o;
+             if (!nb::try_cast<llvm::LLT>(other, o))
+               return true;
+             return self != o;
+           })
+      .def("__hash__",
+           [](const llvm::LLT &self) {
+             return static_cast<Py_ssize_t>(self.getUniqueRAWLLTData());
+           })
       .def("__str__",
            [](const llvm::LLT &self) { return eudsl::toString(self); });
 }
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -45,20 +45,18 @@ void populate_mir(nb::module_ &m) {
                    [](const llvm::LLT &self) { return self.isFloat(); })
       .def_prop_ro("is_valid",
                    [](const llvm::LLT &self) { return self.isValid(); })
-      .def("__eq__",
-           [](const llvm::LLT &self, nb::handle other) {
-             llvm::LLT o;
-             if (!nb::try_cast<llvm::LLT>(other, o))
-               return false;
-             return self == o;
-           })
-      .def("__ne__",
-           [](const llvm::LLT &self, nb::handle other) {
-             llvm::LLT o;
-             if (!nb::try_cast<llvm::LLT>(other, o))
-               return true;
-             return self != o;
-           })
+      .def(
+          "__eq__",
+          [](const llvm::LLT &self, const llvm::LLT &other) {
+            return self == other;
+          },
+          nb::is_operator())
+      .def(
+          "__ne__",
+          [](const llvm::LLT &self, const llvm::LLT &other) {
+            return self != other;
+          },
+          nb::is_operator())
       .def("__hash__",
            [](const llvm::LLT &self) {
              return static_cast<Py_ssize_t>(self.getUniqueRAWLLTData());
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -1,0 +1,50 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "IR/Common.h"
+
+#include <llvm/CodeGenTypes/LowLevelType.h>
+
+// LowLevelType (LLT) is the generic-MIR type: a target-independent "bag of
+// bits" describing a scalar/pointer/vector operand, distinct from the uniqued
+// llvm::Type hierarchy. It is a small value type (not context-owned and not
+// polymorphic), so it is bound by value with no ownership/keep_alive plumbing.
+// Binding it first also proves LLVMCodeGenTypes links into the extension.
+void populate_mir(nb::module_ &m) {
+  nb::class_<llvm::LLT>(m, "LLT")
+      .def_static("scalar", &llvm::LLT::scalar, "size_in_bits"_a)
+      .def_static("pointer", &llvm::LLT::pointer, "address_space"_a,
+                  "size_in_bits"_a)
+      .def_static(
+          "fixed_vector",
+          [](unsigned numElements, unsigned scalarSizeInBits) {
+            return llvm::LLT::fixed_vector(numElements, scalarSizeInBits);
+          },
+          "num_elements"_a, "scalar_size_in_bits"_a)
+      .def_prop_ro("size_in_bits",
+                   [](const llvm::LLT &self) {
+                     return self.getSizeInBits().getKnownMinValue();
+                   })
+      .def_prop_ro(
+          "scalar_size_in_bits",
+          [](const llvm::LLT &self) { return self.getScalarSizeInBits(); })
+      .def_prop_ro("num_elements",
+                   [](const llvm::LLT &self) { return self.getNumElements(); })
+      .def_prop_ro("address_space",
+                   [](const llvm::LLT &self) { return self.getAddressSpace(); })
+      .def_prop_ro("is_scalar",
+                   [](const llvm::LLT &self) { return self.isScalar(); })
+      .def_prop_ro("is_pointer",
+                   [](const llvm::LLT &self) { return self.isPointer(); })
+      .def_prop_ro("is_vector",
+                   [](const llvm::LLT &self) { return self.isVector(); })
+      .def_prop_ro("is_valid",
+                   [](const llvm::LLT &self) { return self.isValid(); })
+      .def("__eq__", [](const llvm::LLT &self,
+                        const llvm::LLT &other) { return self == other; })
+      .def("__ne__", [](const llvm::LLT &self,
+                        const llvm::LLT &other) { return self != other; })
+      .def("__str__",
+           [](const llvm::LLT &self) { return eudsl::toString(self); });
+}
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -47,7 +47,8 @@ jobs:
           clang-format --version
           clang-format --style=file --dry-run --Werror \
             projects/eudsl-llvmpy/src/IR/*.cpp \
-            projects/eudsl-llvmpy/src/IR/*.h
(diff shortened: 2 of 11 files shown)


makslevental (commented):
## 🤖 Automated review — MIR foundation (LLT bindings)

Reviewed the full diff of this PR (base `main`): CMake/coverage wiring, the `llvm.mir` submodule, and the `LowLevelType` (LLT) binding + `test_foundation.py`. This is a clean, well-scoped foundation PR. Findings below; nothing here is a merge blocker.

### Important

- **`__eq__`/`__ne__` without `__hash__` makes `LLT` unhashable** — `src/MIR/Machine.cpp:44-47`. Defining `__eq__` in Python silently sets `__hash__` to `None`, so `set`/`dict` keys of `LLT` will raise `TypeError`. The existing IR bindings deliberately pair the two (`src/IR/Types.cpp:49-56`, `src/IR/Values.cpp:125-132` define `__eq__` **and** `__hash__`). LLT is a small value type that's a natural dict/set key (e.g. legalization tables keyed by type), so it should follow the same pattern. `LLT` has a `hash_value`/`getUniqueRAWLLTData`-style raw payload you can hash, or hash `getUniqueRAWLLTData()`/`toString`.

### Suggestions

- **Kind accessors on the wrong LLT kind will `assert`-abort the interpreter** — `src/MIR/Machine.cpp:26-33`. `getNumElements()` (header line 381 `assert(isVector())`) and `getScalarSizeInBits()`/`getAddressSpace()` assert on mismatched kinds in an assertions-enabled LLVM. Calling `mir.LLT.scalar(32).num_elements` from Python would hard-crash rather than raise. Not required for a foundation PR, but worth a follow-up guard (raise a Python exception) or at least a docstring note. Consider a test documenting the current behavior once decided.
- **Missing faithful predicates** — the LLT surface omits `is_integer`/`is_float` (header lines 264-265) and `is_valid` is bound but there's no way to *construct* an invalid LLT from Python to exercise the `False` branch. Per the project convention of surfacing the full API, `is_integer`/`is_float` are cheap adds; the `scalar` factory only ever yields integers, so `is_float` would currently always be `False`. Fine to defer, but note the surface is integer-only right now.
- **`size_in_bits` uses `getKnownMinValue()`** — `src/MIR/Machine.cpp:24`. Correct for the fixed vectors/scalars bound here, but the name hides that it's the *min* value (matters once scalable vectors are bound). A short comment would prevent a future footgun.

### Test coverage

- `test_foundation.py` is honest and covers the bound surface well: scalar size, predicates, pointer (size + addrspace), fixed vector (elements/scalar/total size), equality both ways, and `__str__` against real LLVM output (`s32`, `<4 x s32>`). Good.
- Gaps (all minor): no test for `__ne__` returning `True`/`False` symmetry beyond the `!=` case, no negative/error-path test (invalid LLT, wrong-kind accessor), and no `__hash__` test (which would currently fail — see above). The docstring correctly explains why there's no `assert_no_leaks()` here (LLT is a non-context value type).

### Strengths

- Library linking is minimal and correct (`LLVMCodeGenTypes`/`GlobalISel`/`SelectionDAG`/`MIRParser`), and binding LLT first as a link-proof is a nice incremental choice.
- Submodule wiring exactly mirrors the existing `populate_*` pattern; the `mir.py` shim matches `ir.py`.
- Coverage gate (`cpp_coverage.sh`, CMake comment) and the format check (`format.yml`) were both extended to `src/MIR` in the same PR — good discipline, keeps later PRs green.
- The file-level comment explaining LLT's value-type nature (no ownership/keep_alive) is accurate and genuinely useful.

_No changes requested — the `__hash__` omission is the one item worth addressing before the DSL PRs start using LLT as a key._

makslevental (commented):
## 🤖 Comprehensive automated review — MIR foundation (LLT bindings)

Reviewed the full diff (base `main`) with four specialized passes: **code quality**, **type design**, **test coverage**, and **comments**. This is a clean, well-scoped foundation PR. No merge blockers; the two Important items below are worth addressing before the DSL PRs start consuming `LLT` as a value/key.

### Critical
None.

### Important

**1. `LLT` defines `__eq__`/`__ne__` but no `__hash__` → the type is unhashable.** `src/MIR/Machine.cpp:44-47`
Registering `__eq__` without `__hash__` sets `tp_hash` to `None` (CPython rule, which nanobind mirrors), so `LLT` can't be a `dict` key or `set` member — `hash(mir.LLT.scalar(32))` raises `TypeError`. This is inconsistent with the two other value-comparable classes, which both pair `__eq__` with `__hash__` (`src/IR/Types.cpp:49,56`, `src/IR/Values.cpp:125,132`; `tests/types/test_types.py:46` even asserts hashability via set membership). LLT is a natural value key — LLVM ships `DenseMapInfo<LLT>` — and the header exposes the exact primitive: `getUniqueRAWLLTData()` (`LowLevelType.h:703`). Add `__hash__` returning that so equal LLTs hash equal. No current test would catch this gap.

**2. Forward-reference comment names future work — violates CLAUDE.md.** `src/eudslllvm_ext.cpp:63-64`
The comment describes "the ... LLT **and, building on it, the MachineFunction/MachineInstr object model**", but only LLT is bound in this PR. CLAUDE.md's "no forward-reference comments" rule forbids this (it rots under rebase/squash), and every sibling submodule comment describes only present contents. Trim to describe LLT only; expand it in the PR that actually adds MachineFunction/MachineInstr.

### Suggestions

**3. Kind-specific accessors abort (debug) / return garbage (release) on the wrong kind.** `src/MIR/Machine.cpp:27,31`
`num_elements` (`getNumElements` asserts `isVector()`, `LowLevelType.h:381`) and `address_space` (`getAddressSpace` asserts pointer, `:504`) crash the interpreter in an assertions-enabled build and read undefined bits under `NDEBUG` (release wheels) — e.g. `mir.LLT.scalar(32).num_elements`. This mirrors the raw LLVM API (consistent with existing IR bindings), so it's a judgment call, but a value type reachable from Python arguably shouldn't be able to abort the process from a plain attribute access. Same class of issue: `pointer(as, 0)` asserts `size > 0` (`:116`).

**4. `__eq__`/`__ne__` use a typed `LLT` arg instead of the `handle`+`try_cast` pattern.** `src/MIR/Machine.cpp:44-47`
The IR bindings take `nb::handle other` and `try_cast`, cleanly returning `False` for non-LLT operands (`src/IR/Types.cpp:49-55`, and `test_types.py:49-50` pins `i32 != 5`). LLT's typed signature means `llt == 5` may raise/behave via nanobind's richcompare fallback rather than returning `False`. Minor consistency nit worth aligning.

**5. Integer/float axis of LLT is currently unreachable from Python.** `src/MIR/Machine.cpp`
`is_integer`/`is_float` (`LowLevelType.h:264-265`) aren't bound, and the `scalar()` factory produces `ANY_SCALAR` (not `INTEGER`/`FLOAT`), with no `integer`/`floatingPoint` factories exposed. The sibling `Type` binding does expose `is_integer`/`is_floating_point` (`Types.cpp:42-44`). Fine to defer in a foundation PR, but the asymmetric predicate subset is worth calling out.

**6. Newly linked CodeGen libs are unused by this diff.** `CMakeLists.txt:153-156`
Only `LLVMCodeGenTypes` is exercised (LLT's `print`/`operator==`). `LLVMGlobalISel`/`LLVMSelectionDAG`/`LLVMMIRParser` aren't referenced yet. If the intent is to prove the whole link surface up front, fine; otherwise consider deferring each to the PR that first uses it. Relatedly, `test_foundation.py:8-9`'s docstring claim that this "proves the CodeGen/CodeGenTypes libraries link" overclaims — only `LLVMCodeGenTypes` is proven (matching the precise comment at `Machine.cpp:12`).

### Test coverage notes
- `test_foundation.py` is honest (asserts derived values like `size_in_bits == 128` and real `str()` output `s32`/`<4 x s32>`, not tautologies) and hits every bound lambda, so the 100% `src/MIR` C++ gate is achievable.
- Gaps (cheap to close): no `__ne__`/`__eq__` **False**-case (`scalar(32) == scalar(64)` is never asserted `False`); `scalar_size_in_bits` tested only on a vector, never a scalar/pointer; `address_space` tested only with AS `0` (the default — doesn't prove decoding); pointer `__str__` untested; no hashing/set test (which would surface #1).

### Strengths
- Correct ownership model: bound by value, no `keep_alive`/`rv_policy`, matching LLT's non-context-owned POD nature; the rationale comment at `Machine.cpp:9-13` is accurate and genuinely useful.
- `fixed_vector` correctly uses a lambda to disambiguate the overloaded `LLT::fixed_vector`, while `scalar`/`pointer` are bound directly — a deliberate, correct distinction.
- Names mirror LLVM verbatim (`size_in_bits`, `scalar_size_in_bits`, `num_elements`, `address_space`, predicates) and `pointer` arg order matches the header exactly.
- Good encapsulation: factory-only construction; the private bit-packed representation is never exposed, so Python can't assemble an inconsistent LLT.
- Submodule wiring mirrors the existing `populate_*`/`ir.py` pattern exactly; coverage script **and** format workflow were both extended to `src/MIR` in the same PR — keeps later PRs green.
- The `assert_no_leaks()` omission is correctly justified (LLT never touches a Context, so there is nothing the leak checker could observe).

_Recommended: address #1 (`__hash__`) and #2 (forward-reference comment) before merge; the rest are optional polish, several deferrable to later stack PRs._

makslevental:
Addressed the automated review in 36a28d5.

**Fixed**
- **`__hash__` (Important):** `LLT` now defines `__hash__` returning `getUniqueRAWLLTData()`, so equal LLTs hash equal and `LLT` works as a `dict`/`set` key — matching the `__eq__`/`__hash__` pairing in the IR `Type`/`Value` bindings.
- **Forward-reference comment (Important):** trimmed the `llvm.mir` submodule comment in `eudslllvm_ext.cpp` to describe only LLT; the MachineFunction/MachineInstr mention will land in the PR that adds them (CLAUDE.md: no forward-reference comments).
- **`__eq__`/`__ne__` casting:** switched to the `nb::handle` + `try_cast` pattern used by the IR bindings, so `llt == 5` returns `False`/`True` rather than going through nanobind's richcompare fallback.
- **`is_integer`/`is_float`:** bound to surface the full LLT predicate API (both are `False` for the `scalar()`/ANY_SCALAR factory, documented in the test).
- **Docstring overclaim:** `test_foundation.py` now says only `LLVMCodeGenTypes` is proven to link, matching the precise comment in `Machine.cpp`.
- **Test coverage:** added hashing/set-membership, `__eq__`/`__ne__` `False` cases + non-LLT operands, `scalar_size_in_bits` on scalar/pointer, non-zero address space + pointer `__str__`, and the new predicates. All of `src/MIR` is exercised, keeping the 100% C++ gate green.

**Deferred (with reasoning)**
- **Kind accessors asserting on the wrong kind** (`num_elements`/`address_space`/`scalar_size_in_bits`): left as-is. This mirrors the raw LLVM API and matches the existing IR bindings; adding Python-exception guards would diverge from the project's "mirror the LLVM C++ API" convention. Worth revisiting as a cross-cutting decision for all bindings rather than one-off here.
- **Unused CodeGen libs** (`LLVMGlobalISel`/`LLVMSelectionDAG`/`LLVMMIRParser`): kept, intentionally linking the whole CodeGen surface up front for the MIR stack (the next PR consumes `LLVMMIRParser`).

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -0,0 +1,68 @@
...
+               return false;
+             return self == o;
+           })
+      .def("__ne__",
+           [](const llvm::LLT &self, nb::handle other) {
+             llvm::LLT o;
+             if (!nb::try_cast<llvm::LLT>(other, o))
+               return true;
+             return self != o;
+           })
```
makslevental: if we use `llvm::LLT` as the type of the other wouldn't it automatically reject? object which don't match in type? that would save the `try_cast` right?

makslevental: Yes — good call. Typing the operand as `llvm::LLT` (by value) + `nb::is_operator()` makes nanobind reject a non-LLT RHS itself: on a failed argument cast an operator-flagged method returns `NotImplemented` rather than raising `TypeError`, so Python falls back to the default identity comparison (`llt == 5` → `False`, `llt != 5` → `True`) — exactly the behavior the `try_cast` was hand-rolling. Dropped the `nb::handle` + `try_cast` on both `__eq__`/`__ne__`; `test_foundation.py`'s mismatched-operand cases still pass.

The one thing `is_operator()` adds that a plain `llvm::LLT` parameter alone wouldn't: without it, a mismatched RHS raises `TypeError` instead of yielding `NotImplemented`, so the `== 5` / `!= 5` cases would throw rather than return `False`/`True`. Kept `__hash__` as-is.

Fixed in dfd37b1 and cascaded up the stack.

makslevental: Follow-up: applied the same simplification to the sibling IR bindings that used the identical `try_cast` pattern — `Type.__eq__` (`src/IR/Types.cpp`) and `Value.__eq__` (`src/IR/Values.cpp`) now take the operand by value + `nb::is_operator()` (pointer-identity compare; Python derives `!=` from `__eq__`, so no separate `__ne__` needed). `test_cpp_coverage.py`'s `== "not a value"` / `!= "not a value"` / `== "not a type"` cases still pass, and `src/IR` stays at 100% line+function coverage. Folded into the same commit and cascaded up the stack.

PR #580, opened 2026-08-17
[eudsl-llvmpy] Add MIR inspection from codegen + .mir round-trip

makslevental:
run_codegen_to_mir lowers an IR module through instruction selection and hands
back a MachineModuleInfo that owns the resulting MachineFunctions (only the ISel
passes are added, so they are retained -- the state llc -stop-after=finalize-isel
produces). Bind the MIR object model on top: MachineFunction, MachineBasicBlock,
MachineInstr, MachineOperand, Register. Add .mir text round-trip via to_mir()
(MIRPrinter) and parse_mir (MIRParser).

Second PR in the MIR bindings/DSL stack; builds on the LLT foundation.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -16,8 +16,11 @@
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/CodeGen/TargetSubtargetInfo.h>
 #include <llvm/CodeGenTypes/LowLevelType.h>
+#include <llvm/IR/Constants.h>
 #include <llvm/IR/DiagnosticInfo.h>
 #include <llvm/IR/DiagnosticPrinter.h>
+#include <llvm/IR/Function.h>
+#include <llvm/IR/GlobalValue.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/Support/MemoryBuffer.h>
@@ -168,10 +171,33 @@ void populate_mir(nb::module_ &m) {
   // target's real registers).
   nb::class_<llvm::Register>(m, "Register")
       .def_prop_ro("id", [](llvm::Register &self) { return self.id(); })
+      .def_prop_ro("is_valid",
+                   [](llvm::Register &self) { return self.isValid(); })
       .def_prop_ro("is_virtual",
                    [](llvm::Register &self) { return self.isVirtual(); })
       .def_prop_ro("is_physical",
-                   [](llvm::Register &self) { return self.isPhysical(); });
+                   [](llvm::Register &self) { return self.isPhysical(); })
+      .def_prop_ro("virt_reg_index",
+                   [](llvm::Register &self) {
+                     if (!self.isVirtual())
+                       throw nb::value_error("register is not virtual");
+                     return self.virtRegIndex();
+                   })
+      .def(
+          "__eq__",
+          [](llvm::Register &self, llvm::Register other) {
+            return self == other;
+          },
+          nb::is_operator())
+      .def(
+          "__ne__",
+          [](llvm::Register &self, llvm::Register other) {
+            return self != other;
+          },
+          nb::is_operator())
+      .def("__hash__", [](llvm::Register &self) {
+        return static_cast<Py_ssize_t>(self.id());
+      });
 
   // llvm::MachineOperand -- one operand of a MachineInstr. is_def/is_use are
   // register-only in LLVM (they assert otherwise), so they are guarded to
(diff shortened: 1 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #580 (MIR inspection + `.mir` round-trip)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`src/MIR/Machine.cpp`, `tests/mir/test_inspect.py`, `tests/mir/test_roundtrip.py`). Overall this is well-crafted, carefully-documented code with sound ownership design and strong behavior-focused tests. The findings below are mostly hardening/polish; a couple are worth addressing before merge.

> Note: this is review-only — nothing was changed. Line numbers are approximate to the current `Machine.cpp`.

---

### 🔴 Critical / High

**1. `parse_mir` and `run_codegen_to_mir` swallow the underlying LLVM diagnostic.** (`src/MIR/Machine.cpp` ~278–292, ~270)
LLVM's MIR parser and codegen passes report errors through `LLVMContext::diagnose(...)`, *not* through their return values — `parseIRModule()` returns null / `parseMachineFunctions()` returns `true` only after the rich `SMDiagnostic` (line, column, caret, reason) has already been dispatched. No custom diagnostic handler is installed, so the default handler prints to process `stderr` (often invisible under a Python/nanobind harness) and the caller only sees a generic `RuntimeError: failed to parse machine functions` — no line, no reason. `pm->run(*module)` in `run_codegen_to_mir` has the same issue and additionally ignores its result, so an errored codegen can return a possibly-empty MMI as if it succeeded. The sibling `parse_assembly` (`IR/Context.cpp`) already does this correctly: capture into an `SMDiagnostic`/scoped handler and thread the real message into the thrown exception. Recommend mirroring that pattern in both entry points. Note the two negative parse tests only assert `pytest.raises(RuntimeError)`, so they pass while masking this.

**2. `opcode_name` dereferences a 3-link chain with no null guard.** (`src/MIR/Machine.cpp` ~140–146)
`self.getMF()->getSubtarget().getInstrInfo()` — `getMF()` is null for a `MachineInstr` not attached to a function, and `getInstrInfo()` is documented as possibly null for some targets. Either is a hard segfault (not a Python exception). The happy path (instructions reached via `MachineBasicBlock.instructions`) keeps `getMF()` non-null, but this is the one accessor that skips the guarding discipline applied so carefully to `reg`/`imm`/`operand`. Recommend at least null-checking `tii` and throwing.

---

### 🟠 Important

**3. CLAUDE.md C++ brace-style violations.** (`src/MIR/Machine.cpp` ~219–221, ~223–225, ~254–256, ~278–280)
Four `if` statements have a multi-line `throw` body but omit braces. CLAUDE.md: "A conditional takes braces whenever its condition spans multiple lines **or its body spans multiple lines**" — and its example is exactly a multi-line `throw`. In-repo precedent: `IR/Ownership.cpp:107-111` braces the same pattern. `clang-format` won't add these, so it needs a manual fix. (The single-line-body throw sites elsewhere are correctly brace-free.)

---

### 🟡 Suggestions

**4. `MachineModuleInfo` struct invariants are unenforced.** (`src/MIR/Machine.cpp` ~37–56)
The `pm` XOR `ownedMmi` mutual exclusion, the `mmi` handle pointing into the *populated* owner, and the destruction-order requirement are all expressed only by comments + construction discipline. Destruction order in particular is correct *only* by member-declaration order — a stylistic reorder would silently introduce a use-after-free with no compiler complaint. Consider collapsing `pm`+`ownedMmi` into a `std::variant` with an `mmi()` accessor derived from the active alternative (makes the invariant unrepresentable-when-violated and removes the redundant stored handle), or at minimum a private ctor + named factories so the ordering contract isn't a silent property of field layout. Also: the local `MachineModuleInfo` shadows `llvm::MachineModuleInfo` in the anon namespace — survivable only because every LLVM reference is fully qualified; any future unqualified use silently binds to the wrapper. A rename (e.g. `MirModule`) would remove that hazard.

**5. `to_mir()` silently omits functions with no MachineFunction.** (`src/MIR/Machine.cpp` ~40–48)
Correct for declarations (matches `llc`), but the code doesn't distinguish "declaration, expected" from "definition that failed to lower, unexpected." A comment and/or an `isDeclaration()` distinction would prevent silent function loss after a partial codegen failure.

**6. `LCOV_EXCL_LINE` on reachable error branches.** (`src/MIR/Machine.cpp` ~254–256, ~278–280)
`addISelPasses()` failing and `createMIRParser` returning null are rare but genuinely reachable (misconfigured `TargetMachine`), not dead code. Excluding them from coverage permanently hides that their failure behavior is never exercised. The throw messages themselves are fine — the concern is the coverage suppression.

**7. Comment accuracy nits.**
- `~211` — "The result of `run_codegen_to_mir`" is incomplete; `parse_mir` returns this same class. Add "or `parse_mir`". (The struct-level comment at ~29 correctly covers both.)
- `~237–240` — `FreeMachineFunctionPass` is appended by the emit pipeline (`addPassesToEmitFile`), not by `addMachinePasses`; the conclusion is right but the attribution is imprecise.
- `~29–36` — ownership prose lists Module + MMI but never mentions `ctxKeepAlive`, the first pinned dependency.
- `~44–45` — "`llc -stop-after`" is missing its argument; tighten to `-stop-after=finalize-isel` for consistency.

---

### 🧪 Tests — strong, with a few gaps

Coverage is strong and behavior-focused (opcode mnemonics, def/use structure, immediate values, print∘parse∘print idempotence), every test ends with `assert_no_leaks()`, and reference shapes are honestly derived from real `llc -stop-after=finalize-isel` output. `test_mir_print_parse_is_idempotent` also exercises the distinct parse-path ownership (`ownedMmi`).

**CI note:** the AArch64 `skipif` gate does *not* dead-code these in CI — `EUDSL_LLVMPY_TARGETS` defaults to the native arch, and the matrix includes `ubuntu-22.04-arm` + `macos-14` (ARM ⇒ AArch64 native), where they run (the coverage job runs on `macos-14` too). Consequence: the x86 legs (`ubuntu-22.04`, `windows-2022`) skip the whole module. Acceptable since the bindings are target-agnostic.

Gaps worth considering:
- `is_def`/`is_use` returning `false` on a **non-register** operand is never asserted — the single most non-obvious guarded branch (~110–118) is currently unexecuted; a regression removing the `isReg() &&` guard would crash uncaught.
- **Module-consumed** contract after `run_codegen_to_mir` (`mod.take()`) isn't tested — no assertion that touching `mod` afterward raises.
- **Child lifetime after dropping the MMI handle** isn't tested — bind `mf`, drop the `mmi` Python ref, then access `mf`/blocks. This is exactly the `reference_internal` guarantee most likely to regress into a UAF.
- Only single-block / single-function modules are inspected, so `MachineBasicBlock.number` is only ever `0` and `to_mir`'s multi-function loop is uncovered — despite nested-control-flow support landing earlier in this stack.
- `Register.is_virtual`/`is_physical` are each asserted only in their true case.

---

### ✅ Handled well
- Operand accessors correctly guard LLVM's assert-prone APIs: `reg`/`imm` throw `value_error` on kind mismatch, `is_def`/`is_use` guard on `isReg()`, `operand` bounds-checks and throws `index_error`, `machine_function` distinguishes "no such function" from "no MachineFunction" with distinct `key_error`s.
- Ownership/lifetime design is sound: member order gives correct reverse destruction (MMI → module → context), both construction paths are internally consistent (no double-free), and `keep_alive<0,2>` / `<0,3>` correctly pin the `TargetMachine`.
- Comments are unusually high quality — every load-bearing technical claim checks out against the LLVM headers, and no forward-reference comments (CLAUDE.md-compliant).

---

makslevental:
Addressed the automated review. Changes are in the updated `inspect` commit and cascaded up the rest of the stack (#582–#591).

**Critical / High**
- **Diagnostics no longer swallowed (#1):** `parse_mir` and `run_codegen_to_mir` now install a scoped `LLVMContext` diagnostic handler (`ScopedDiagnosticCapture`, restored on scope exit) that captures the real error — MIR/codegen report through `LLVMContext::diagnose`, not their return values — and threads it into the thrown message. `run_codegen_to_mir` throws on a captured error instead of returning a silently-empty module (this also covers the GlobalISel `DisableWithDiag` path added upstack). The two negative parse tests now assert the real detail is present.
- **`opcode_name` segfault guard (#2):** null-check `getMF()` and `getInstrInfo()`; either now raises a Python exception instead of dereferencing null.

**Important**
- **Brace style (#3):** all multi-line `throw` bodies are now braced (CLAUDE.md).

**Suggestions**
- **#5 / #7 comment fixes:** `to_mir` documents why functions without a MachineFunction are skipped; `FreeMachineFunctionPass` is correctly attributed to `addPassesToEmitFile` (not `addMachinePasses`); the ownership comment now mentions `ctxKeepAlive`; `llc -stop-after=finalize-isel` is spelled with its argument.
- **#6 (`LCOV_EXCL` on error branches):** kept the suppressions (converted to `EXCL_START/STOP`) on the `createMIRParser`-null and `addISelPasses`-fail branches. Both require a broken `TargetMachine` to reach and aren't reproducible from Python, and the CI gate is 100% line coverage — so exercising them isn't feasible without dropping the gate. `src/MIR` stays at 100% line + function coverage.

**Test coverage gaps** — all added: non-register `is_def`/`is_use`, the module-consumed contract after `run_codegen_to_mir`, `MachineFunction` lifetime after dropping the MMI handle, multi-block block numbering, multi-function `to_mir`, and the `Register` virtual/physical false cases.

**Not applied — #4 (struct invariants / rename).** I did not rename `MachineModuleInfo` → `MirModule` or add a private-ctor + factories. The upstack PRs (#582–#590) grow this exact struct — a third ownership mode (`mmiwp` build path), plus `tm`/`emitted` fields and an `emit_object` pipeline — and construct it by aggregate init. The rename + private ctor fight that design at every branch (semantic merge hazards), for a benign shadow (every `llvm::` reference is already fully qualified). Given the cost/benefit across the stack, I left the struct as the upstack design expects. Happy to revisit as a dedicated cross-stack change if you'd like the invariant enforcement.

Verified: `emit-jit` (the cumulative top of the C++) builds and all 68 `tests/mir` pass; an intermediate branch builds; `src/MIR` C++ coverage is 100% line + function.

PR #582, opened 2026-08-17
[eudsl-llvmpy] Build generic (GlobalISel) MIR with MachineIRBuilder

makslevental:
Add create_machine_function to make a fresh, empty MachineFunction to build
into (a void() IR stub + owned MachineModuleInfo + entry block), and bind
MachineIRBuilder with typed generic builders (build_constant/add/sub/mul/copy)
that take the result type as an LLT and return the def Register so builds chain.
MachineFunction.create_generic_vreg exposes raw vreg creation.

Third PR in the MIR bindings/DSL stack; builds on the inspection PR.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -5,18 +5,23 @@
 #include "IR/Common.h"
 #include "IR/Ownership.h"
 
+#include <llvm/CodeGen/GlobalISel/MachineIRBuilder.h>
 #include <llvm/CodeGen/MIRParser/MIRParser.h>
 #include <llvm/CodeGen/MIRPrinter.h>
 #include <llvm/CodeGen/MachineBasicBlock.h>
 #include <llvm/CodeGen/MachineFunction.h>
 #include <llvm/CodeGen/MachineInstr.h>
+#include <llvm/CodeGen/MachineInstrBuilder.h>
 #include <llvm/CodeGen/MachineModuleInfo.h>
 #include <llvm/CodeGen/MachineOperand.h>
+#include <llvm/CodeGen/MachineRegisterInfo.h>
 #include <llvm/CodeGen/TargetInstrInfo.h>
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/CodeGen/TargetSubtargetInfo.h>
 #include <llvm/CodeGenTypes/LowLevelType.h>
+#include <llvm/CodeGenTypes/LowLevelType.h>
 #include <llvm/IR/Constants.h>
+#include <llvm/IR/DerivedTypes.h>
 #include <llvm/IR/DiagnosticInfo.h>
 #include <llvm/IR/DiagnosticPrinter.h>
 #include <llvm/IR/Function.h>
@@ -486,6 +491,12 @@ void populate_mir(nb::module_ &m) {
             return &self.getFunction();
           },
           nb::rv_policy::reference_internal)
+      .def(
+          "create_generic_vreg",
+          [](llvm::MachineFunction &self, llvm::LLT ty) -> llvm::Register {
+            return self.getRegInfo().createGenericVirtualRegister(ty);
+          },
+          "type"_a, "Create a new generic virtual register of the given LLT.")
       .def("__str__",
            [](llvm::MachineFunction &self) { return eudsl::toString(self); });
 
@@ -617,4 +628,81 @@ void populate_mir(nb::module_ &m) {
                                      /*pm=*/nullptr, std::move(ownedMmi), mmi};
       },
       "text"_a, "context"_a, "target_machine"_a, nb::keep_alive<0, 3>());
+
+  // Create a fresh, empty MachineFunction to build into: make a void() IR
+  // Function stub named `name` (a MachineFunction must attach to one), get its
+  // MachineFunction from a freshly owned MachineModuleInfo, and give it a
+  // single empty entry block. Consumes the module into the returned wrapper
+  // (like run_codegen_to_mir), which owns everything the MachineFunction
+  // depends on.
+  m.def(
+      "create_machine_function",
+      [](eudsl::Module &mod, llvm::TargetMachine &tm, const std::string &name) {
+        std::shared_ptr<llvm::LLVMContext> ctxKeepAlive =
+            mod.context().shared();
+        std::unique_ptr<llvm::Module> module = mod.take();
+        module->setDataLayout(tm.createDataLayout());
+
+        llvm::FunctionType *fnTy =
+            llvm::FunctionType::get(llvm::Type::getVoidTy(module->getContext()),
+                                    /*isVarArg=*/false);
+        llvm::Function *f = llvm::Function::Create(
+            fnTy, llvm::GlobalValue::ExternalLinkage, name, *module);
+
+        auto ownedMmi = std::make_unique<llvm::MachineModuleInfo>(&tm);
+        llvm::MachineFunction &mf = ownedMmi->getOrCreateMachineFunction(*f);
+        mf.push_back(mf.CreateMachineBasicBlock());
+
+        llvm::MachineModuleInfo *mmi = ownedMmi.get();
+        return new MachineModuleInfo{std::move(ctxKeepAlive), std::move(module),
+                                     /*pm=*/nullptr, std::move(ownedMmi), mmi};
+      },
+      "module"_a, "target_machine"_a, "name"_a, nb::keep_alive<0, 2>());
+
+  // llvm::MachineIRBuilder -- the GlobalISel builder for generic (G_*) MIR. The
+  // typed helpers take the result type as an LLT (a fresh generic vreg is
+  // created for it) and the operands as Registers, returning the def Register
+  // so builds chain. Construction positions the builder at the end of the
+  // function's entry block.
+  nb::class_<llvm::MachineIRBuilder>(m, "MachineIRBuilder")
+      .def(
+          "__init__",
+          [](llvm::MachineIRBuilder *self, llvm::MachineFunction &mf) {
+            new (self) llvm::MachineIRBuilder(mf);
+            self->setMBB(mf.front());
+          },
+          "machine_function"_a, nb::keep_alive<1, 2>())
+      .def(
+          "build_constant",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty,
+             int64_t value) -> llvm::Register {
+            return self.buildConstant(ty, value).getReg(0);
+          },
+          "type"_a, "value"_a)
+      .def(
+          "build_add",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty, llvm::Register lhs,
+             llvm::Register rhs) -> llvm::Register {
+            return self.buildAdd(ty, lhs, rhs).getReg(0);
+          },
+          "type"_a, "lhs"_a, "rhs"_a)
+      .def(
+          "build_sub",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty, llvm::Register lhs,
+             llvm::Register rhs) -> llvm::Register {
+            return self.buildSub(ty, lhs, rhs).getReg(0);
+          },
+          "type"_a, "lhs"_a, "rhs"_a)
+      .def(
+          "build_mul",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty, llvm::Register lhs,
+             llvm::Register rhs) -> llvm::Register {
+            return self.buildMul(ty, lhs, rhs).getReg(0);
+          },
+          "type"_a, "lhs"_a, "rhs"_a)
+      .def(
+          "build_copy",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty, llvm::Register src)
+              -> llvm::Register { return self.buildCopy(ty, src).getReg(0); },
+          "type"_a, "src"_a);
 }
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -19,7 +19,6 @@
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/CodeGen/TargetSubtargetInfo.h>
 #include <llvm/CodeGenTypes/LowLevelType.h>
-#include <llvm/CodeGenTypes/LowLevelType.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DerivedTypes.h>
 #include <llvm/IR/DiagnosticInfo.h>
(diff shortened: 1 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #582 (Build generic GlobalISel MIR with `MachineIRBuilder`)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`src/MIR/Machine.cpp` + `tests/mir/test_build_generic.py`). The design is clean and idiomatic: the typed builders mirror LLVM's own `DstOp(LLT)`/`SrcOp(Register)` overloads, the `keep_alive` chain is correct, and `create_machine_function` faithfully reuses the ownership struct from the prior PRs. The dominant theme worth attention is **release-build safety**: nearly every user-supplied precondition on the new build surface is enforced only by LLVM `assert`s, which are compiled out under `NDEBUG` in the shipped wheel.

> Review-only — nothing was changed. Line numbers are approximate to the current `Machine.cpp` / test file.

---

### 🔴 Critical / High — release-build safety (asserts vanish under `NDEBUG`)

The neighboring `parse_mir` / `run_codegen_to_mir` establish the right pattern in this same file: check the LLVM result and `throw`. The new build helpers instead lean on LLVM asserts for user input, so in a release wheel these become segfaults or silent malformed MIR with no Python exception.

**1. `MachineIRBuilder.__init__` calls `mf.front()` with no empty-block guard.** (`Machine.cpp` ~343–345)
`setMBB(mf.front())` on a `MachineFunction` with zero blocks dereferences the ilist sentinel — UB in both debug and release (the ilist doesn't assert here). `create_machine_function` seeds one block, but the ctor is public and accepts *any* MachineFunction — e.g. one from `parse_mir` with an empty body, or a declaration. `MachineFunction::empty()` exists and isn't consulted. Recommend guarding: `if (mf.empty()) throw nb::value_error(...)`.

**2. `build_add`/`build_sub`/`build_mul` — type mismatch and cross-function operands are assert-only.** (`Machine.cpp` ~356–374)
The only validation is `validateBinaryOp`'s `assert(Res == Op0 && Res == Op1)`. Under `NDEBUG`: (a) a result `ty` that disagrees with the operand LLTs silently builds malformed MIR that faults in a much later pass/printer; (b) a `Register` minted by a *different* MachineFunction is not checked against `self`'s `MachineRegisterInfo` — `getType` returns a default `LLT{}` for an out-of-bounds vreg rather than erroring, yielding corrupt MIR referencing a nonexistent register. Recommend validating in C++ (regardless of `NDEBUG`): operand types `== ty`, each register `isVirtual()` and in-bounds for this MF's MRI.

**3. `build_constant` — invalid/pointer/scalable LLT is assert-only.** (`Machine.cpp` ~349–353)
Width is derived from `ty.getScalarSizeInBits()`; a pointer LLT, scalable vector, or invalid `LLT{}` reaches asserting accessors that vanish in release, silently emitting a wrong-width `G_CONSTANT`. Recommend rejecting non-scalar/non-fixed-vector `ty` with a Python exception.

**4. `create_machine_function` — duplicate/empty name silently diverges.** (`Machine.cpp` ~312–333)
`Function::Create` does not fail on a name collision — it appends a numeric suffix (`f` → `f.1`). Since the module is user-supplied via `mod.take()` and may already contain functions, `create_machine_function(mod, tm, "f")` can return an MMI where `mmi.machine_function("f")` then throws `key_error("no function named 'f'")` — a self-contradictory-looking failure with no hint the name was rewritten. Empty name behaves similarly. Recommend rejecting empty names and checking `module->getFunction(name)` first.

---

### 🟠 Important

**5. `create_generic_vreg` renames the LLVM method — CLAUDE.md "mirror the LLVM API" violation.** (`Machine.cpp` ~213–216)
The lambda calls `createGenericVirtualRegister`; CLAUDE.md requires binding names as LLVM spells them ("do not rename or merge a method because it reads as semantically equivalent"). `vreg` is LLVM slang, not the method name. Recommend `create_generic_virtual_register`.

**6. `create_machine_function` surfaces no failure if the target can't build the MF.** (`Machine.cpp` ~325–327)
Unlike its immediate siblings (`run_codegen_to_mir`, `parse_mir`) which check returns and throw, this factory validates nothing around `getOrCreateMachineFunction` / target setup, so a malformed/incompatible `TargetMachine` aborts inside LLVM rather than raising the clean exceptions the rest of the module provides. Recommend mirroring the siblings' defensive style.

---

### 🟡 Suggestions

**7. `Register` crosses the boundary as a bare MF-agnostic value.** (`Machine.cpp` ~105–110, ~213–216, ~356–388)
Memory-safe (trivially-copyable int) but *semantically* unsound: a vreg is only meaningful relative to its MachineFunction's `MachineRegisterInfo`, yet nothing pairs a `Register` with its owner or lets client code cross-check. This is the root enabler of finding #2's cross-function misuse. If cross-MF safety matters later, consider a small wrapper pairing `Register` with its MF, or making registers obtainable only through a builder.

**8. `build_*` expose only the fresh-vreg (LLT) result form.** (`Machine.cpp` ~340–388)
LLVM's `buildAdd`/etc. accept a `DstOp` that can be an existing `Register` or an `LLT`. Binding only the `LLT` form bakes in one choice — acceptable as SSA-style DSL scaffolding (and the returned def Register chains nicely), but a real completeness gap if arbitrary destination registers are ever needed. Also: the builder bakes in "single entry block, straight-line" (no insertion-point control, block creation, or terminators) — fine as a first cut, but it's an ambient assumption not stated in the type.

**9. Comment/label nits.**
- `test_build_generic.py:8` — the docstring's **"Phase 1 object model / printer"** is a forward/version reference that CLAUDE.md's "No forward-reference comments" rule warns goes stale on rebase/squash. Drop to "the MIR object model / printer". *(All `Machine.cpp` comments were verified clause-by-clause and are accurate, including "end of the entry block" — `setMBB` does set the insert point to `MBB.end()`.)*
- `test_build_generic.py:11–13` — "runs on any runner, no specific backend required" is accurate in spirit but the host/native backend *is* still required (`create_machine_function` calls `tm.createDataLayout()`); optionally tighten to "no non-host backend required".

---

### 🧪 Tests — solid happy path, but ownership/error paths and the new consume contract are untested

The three tests are DAMP, leak-checked, and correctly target-independent — `triple=None` is genuinely sound for representation-only G_* MIR (no legalization/selection runs), giving these broader CI reach than the AArch64-gated sibling files. The full opcode sequence assertion (including that `build_copy` lowers to plain `COPY`, not `G_COPY`) is a nice, regression-prone detail to pin.

Gaps worth considering:
- **`create_machine_function` module-consume contract is untested** — the sibling `tests/jit/test_jit.py` pins exactly this (`_is_consumed`, `RuntimeError match="has been consumed"` on reuse); this factory has the same `mod.take()` semantics but no equivalent assertion.
- **`build_copy`'s returned Register is never asserted** — it's called only for its side-effect; every other `build_*` return is the advertised chainable def, so this one is unverified.
- **Builder-outliving-its-MachineFunction is untested** — `keep_alive<1,2>` is the load-bearing safety mechanism here, and a wrong index would be a silent segfault, not a caught error.
- **`create_generic_vreg` test name overstates coverage** — `test_..._has_requested_type` only asserts `is_virtual`, never that the vreg carries the requested LLT (there's no LLT read-back binding — `MachineRegisterInfo::getType`). Either add the read-back or rename the test.
- No negative-path coverage for any of findings #1–#4 (empty-block MF, mismatched LLT, cross-function register, duplicate/empty name).
- `build_add/sub/mul` results aren't individually asserted `is_virtual`; only the two constants are.

---

### ✅ Handled well
- **Ownership/lifetime is correct.** `create_machine_function`'s `keep_alive<0,2>` is right (module is consumed via `take()`, context held by the struct's shared_ptr, only the `TargetMachine` needs pinning) and consistent with the codegen/parse factories. `MachineIRBuilder`'s `keep_alive<1,2>` correctly pins the MachineFunction, and combined with the prior PR's `reference_internal` handoff forms an unbroken builder → MF → MMI → module/context/tm chain — the builder cannot outlive its storage.
- `Register` returned by value is safe (trivially copyable).
- The typed builders mirror LLVM's `DstOp`/`SrcOp` design rather than inventing semantics; returning the def Register for chaining is idiomatic and ergonomic.
- Python imports at module top, `assert_no_leaks()` used correctly, license header present — all CLAUDE.md-compliant. All `Machine.cpp` comments are accurate.

---

makslevental:
Addressed the automated review. Changes are in the updated `build-generic` commit and cascaded up the rest of the stack (#583–#591), including renaming the callers the method rename touched.

**Critical / High — release-build safety** (the asserts these relied on are compiled out under `NDEBUG` in the shipped wheel, so bad input would segfault or emit malformed MIR with no exception):
- **#1 `MachineIRBuilder.__init__`:** rejects a block-less `MachineFunction` (`setMBB(mf.front())` would deref the ilist sentinel). The ctor is public and accepts any MF — a bodyless one from `parse_mir` now raises instead of faulting.
- **#2 `build_add`/`build_sub`/`build_mul` (and `build_copy`):** each operand must be a virtual register of *this* builder's `MachineFunction` carrying the result type. A single `MachineRegisterInfo::getType` compare rejects both a type mismatch and a register minted by a different `MachineFunction` (`getType` returns an invalid `LLT{}` for a non-vreg / out-of-bounds / wrong-function register).
- **#3 `build_constant`:** rejects a non-scalar / non-fixed-vector result type (pointer, scalable, invalid) that would emit a wrong-width `G_CONSTANT`.
- **#4 `create_machine_function`:** rejects an empty name and a name already in the module (`Function::Create` silently suffixes a collision → a confusing later "no function named"). Validated before `take()`, so a rejected call leaves the module usable.

**Important**
- **#5:** renamed `create_generic_vreg` → `create_generic_virtual_register` to mirror `createGenericVirtualRegister` (CLAUDE.md). Propagated the caller rename through the DSL (`dsl/machine.py`) and the `cf_primitives`/`build_target` tests upstack.
- **#6:** `create_machine_function` — the reachable bad-input failures are now guarded (#1/#4). Unlike `parse_mir`/`run_codegen_to_mir`, `getOrCreateMachineFunction` has no failure return to check, so there's no additional LLVM result to defensively test here.

**Comment nits (#9):** dropped the "Phase 1" forward reference and tightened the backend note to "no non-host backend required".

**Tests added:** module-consume contract, block-less-MF rejection, non-scalar constant rejection, operand type mismatch, cross-function register rejection, builder outliving its `MachineFunction` handle (`keep_alive<1,2>`), `build_copy` return + all `build_*` results asserted virtual; the generic-vreg test is renamed and now verifies the LLT indirectly (accepted as an operand of that type). `src/MIR` stays at 100% line + function coverage.

**Not applied (design suggestions the reviewer flagged optional):** #7 (wrap `Register` with its `MachineFunction`) and #8 (arbitrary `DstOp` / multi-block builder control). The per-build validation in #2 closes the cross-function safety hole #7 was concerned about; the wrapper/expanded builder are larger design changes better done as their own step.

Verified: PR #582's branch builds and its 41 `tests/mir` pass; the cumulative top of the stack builds and all 76 `tests/mir` pass.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -617,4 +644,112 @@ void populate_mir(nb::module_ &m) {
...
+      [](eudsl::Module &mod, llvm::TargetMachine &tm, const std::string &name) {
+        // Validate before take() so a rejected call leaves the module usable.
+        // Function::Create does not fail on a name collision -- it appends a
+        // numeric suffix (`f` -> `f.1`), which would then make
+        // machine_function(name) throw a confusing "no function named" error.
+        if (name.empty())
+          throw nb::value_error("function name must not be empty");
+        if (mod.get().getFunction(name))
+          throw nb::value_error(
+              ("module already has a function named '" + name + "'").c_str());
```
makslevental: needs braces

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -617,4 +644,112 @@ void populate_mir(nb::module_ &m) {
...
+
+        std::shared_ptr<llvm::LLVMContext> ctxKeepAlive =
+            mod.context().shared();
+        std::unique_ptr<llvm::Module> module = mod.take();
+        module->setDataLayout(tm.createDataLayout());
+
+        llvm::FunctionType *fnTy =
+            llvm::FunctionType::get(llvm::Type::getVoidTy(module->getContext()),
+                                    /*isVarArg=*/false);
```
makslevental: why assume void type?

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -617,4 +644,112 @@ void populate_mir(nb::module_ &m) {
...
+        llvm::Function *f = llvm::Function::Create(
+            fnTy, llvm::GlobalValue::ExternalLinkage, name, *module);
```
makslevental: why llvm::GlobalValue::ExternalLinkage?

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Agreed — the condition and body span multiple lines, so per the LLVM brace rule in CLAUDE.md this takes braces. Will fix.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: No principled reason — it's a hardcoded assumption. `create_machine_function` only needs *an* IR function to anchor the MachineFunction (MIR is built into its entry MBB; the IR signature is never read for that). But that's exactly the kind of baked-in choice CLAUDE.md's "do not hardcode assumptions" rule warns against, so I'll surface the function type as an argument (defaulting to `void()`).

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Same story — a hardcoded default rather than a surfaced one. `Function::Create`'s own default is `ExternalLinkage`, so I'll add a `linkage` argument defaulting to that, mirroring how `Function.create` exposes linkage per CLAUDE.md.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -617,4 +644,112 @@ void populate_mir(nb::module_ &m) {
...
+      .def(
+          "build_constant",
+          [](llvm::MachineIRBuilder &self, llvm::LLT ty,
+             int64_t value) -> llvm::Register {
+            // buildConstant derives the width from ty's scalar size; a pointer,
+            // scalable-vector, or invalid LLT hits asserting accessors that
+            // vanish under NDEBUG and would emit a wrong-width G_CONSTANT.
+            if (!(ty.isScalar() || ty.isFixedVector()))
+              throw nb::value_error("build_constant requires a scalar or "
+                                    "fixed-vector type");
```
makslevental: needs braces

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -617,4 +644,112 @@ void populate_mir(nb::module_ &m) {
...
+  nb::class_<llvm::MachineIRBuilder>(m, "MachineIRBuilder")
+      .def(
+          "__init__",
+          [](llvm::MachineIRBuilder *self, llvm::MachineFunction &mf) {
+            // setMBB(mf.front()) on a block-less MachineFunction dereferences
+            // the ilist sentinel (UB, no assert). create_machine_function seeds
+            // a block, but this ctor is public and accepts any MachineFunction.
+            if (mf.empty())
+              throw nb::value_error(
+                  "MachineFunction has no basic block to build into");
```
makslevental: needs braces

makslevental: Agreed — multi-line body, takes braces. Will fix.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Agreed — multi-line body, takes braces. Will fix.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Fixed in 69835493 — braces added.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Fixed in 69835493 — braces added.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Fixed in 69835493 — braces added.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Fixed in 69835493 — `function_type` is now a surfaced argument (defaults to `void()`).

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Fixed in 69835493 — `linkage` is now a surfaced argument (defaults to `ExternalLinkage`).

PR #583, opened 2026-08-17
[eudsl-llvmpy] Add @machine_function DSL + MachineValue for generic MIR

makslevental:
A Pythonic front end for building generic (GlobalISel) MIR, mirroring the IR
DSL's @function/ArithValue. Parameters annotated with an LLT arrive as
MachineValues over fresh generic vregs; `+ - *` emit G_ADD/G_SUB/G_MUL through
a contextual MachineIRBuilder, and Python ints coerce to G_CONSTANTs. Returns a
DSLMachineFunction exposing the built MachineFunction and its .to_mir().

Fourth PR in the MIR bindings/DSL stack; builds on the MachineIRBuilder PR.

### DIFF
--- a/projects/eudsl-llvmpy/src/llvm/dsl/machine.py
+++ b/projects/eudsl-llvmpy/src/llvm/dsl/machine.py
@@ -4,27 +4,21 @@
 """The @machine_function decorator and MachineValue: a Pythonic front end for
 building generic (GlobalISel) MIR, mirroring the IR DSL's @function/ArithValue.
 
-Unlike the IR builder, MachineIRBuilder is not wired into the C++ InsertPoint
-stack, so the "current builder" is tracked here with a plain stack pushed for
-the duration of tracing a @machine_function body. A MachineValue wraps a
-Register plus its LLT and overloads `+ - *` onto build_add/build_sub/build_mul;
-Python ints coerce to a G_CONSTANT of the other operand's type.
+The "current builder" is tracked in C++ on a thread-local stack (the same
+mechanism the IR builder uses): `with MachineIRBuilder(mf):` makes it current
+and current_machine_builder() reads it back. A MachineValue wraps a Register
+plus its LLT and overloads `+ - *` onto build_add/build_sub/build_mul; Python
+ints coerce to a G_CONSTANT of the other operand's type.
 """
 
 import inspect
 
-from ..eudslllvm_ext.mir import LLT, MachineIRBuilder, create_machine_function
-
-_builder_stack = []
-
-
-def current_machine_builder():
-    """The MachineIRBuilder for the @machine_function body being traced."""
-    if not _builder_stack:
-        raise RuntimeError(
-            "no current MachineIRBuilder; use this only inside a @machine_function body"
-        )
-    return _builder_stack[-1]
+from ..eudslllvm_ext.mir import (
+    LLT,
+    MachineIRBuilder,
+    create_machine_function,
+    current_machine_builder,
+)
 
 
 class MachineValue:
@@ -33,16 +27,41 @@ class MachineValue:
     def __init__(self, reg, llt):
         self.reg = reg
         self.llt = llt
+        # Anchor to the builder tracing this value's MachineFunction. A
+        # MachineValue carries a bare reg + llt, so a value that escapes its
+        # @machine_function body could otherwise emit into a different (e.g.
+        # nested) function's builder, referencing vregs it doesn't own. A
+        # MachineValue is only ever constructed while a builder is current.
+        self._builder = current_machine_builder()
+
+    def _require_current(self):
+        if self._builder is not current_machine_builder():
+            raise RuntimeError(
+                "MachineValue used outside the @machine_function body that "
+                "created it (it belongs to a different MachineFunction)"
+            )
 
     def _coerce(self, other):
-        """A MachineValue passes through; a Python int becomes a G_CONSTANT."""
+        """A MachineValue passes through; a Python int becomes a G_CONSTANT of
+        this value's LLT. Only an exact int (not bool, not float) is accepted --
+        int(other) would silently truncate a float or parse a str into a
+        wrong-typed constant that survives into the release wheel."""
+        self._require_current()
         if isinstance(other, MachineValue):
+            other._require_current()
             return other
-        reg = current_machine_builder().build_constant(self.llt, int(other))
+        if type(other) is not int:
+            raise TypeError(
+                f"cannot coerce {other!r} to a {self.llt} constant; "
+                "expected a MachineValue or an int"
+            )
+        reg = current_machine_builder().build_constant(self.llt, other)
         return MachineValue(reg, self.llt)
 
     def _binary(self, other, method, *, reflected=False):
         other = self._coerce(other)
+        if self.llt != other.llt:
+            raise TypeError(f"mismatched types: {self.llt} and {other.llt}")
         b = current_machine_builder()
         lhs, rhs = (other, self) if reflected else (self, other)
         reg = getattr(b, method)(self.llt, lhs.reg, rhs.reg)
@@ -71,13 +90,14 @@ class DSLMachineFunction:
     """The result of @machine_function: the built MachineFunction plus the
     MachineModuleInfo that owns it."""
 
-    def __init__(self, mmi, name):
+    def __init__(self, mmi, name, machine_function):
         self.mmi = mmi
         self.name = name
+        self._machine_function = machine_function
 
     @property
     def machine_function(self):
-        return self.mmi.machine_function(self.name)
+        return self._machine_function
 
     def to_mir(self):
         return self.mmi.to_mir()
@@ -94,26 +114,38 @@ def _resolve_llt(annotation):
 def machine_function(*, module, target, name=None):
     """Build a generic-MIR MachineFunction by tracing `f`. Each parameter is
     annotated with an LLT and arrives as a MachineValue over a fresh generic
-    vreg; the body emits instructions through the contextual builder."""
+    vreg; the body emits instructions through the contextual builder.
+
+    The body's return value is not turned into a terminator -- this traces a
+    straight-line generic-MIR fragment, so `return a + b` just ends tracing.
+    If the body raises, the partially-built MachineFunction is left in `mmi`
+    (there is no rollback), but the exception propagates unchanged.
+    """
 
     def decorator(f):
-        sig = inspect.signature(f)
-        param_llts = [_resolve_llt(p.annotation) for p in sig.parameters.values()]
+        params = list(inspect.signature(f).parameters.values())
+        for p in params:
+            if p.kind not in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD):
+                raise TypeError(
+                    f"@machine_function parameter {p.name!r} must be positional; "
+                    "*args, **kwargs, and keyword-only parameters are not supported"
+                )
+            if p.annotation is inspect.Parameter.empty:
+                raise TypeError(
+                    f"@machine_function parameter {p.name!r} is missing an LLT "
+                    "annotation"
+                )
+        param_llts = [_resolve_llt(p.annotation) for p in params]
         fn_name = name or f.__name__
 
         mmi = create_machine_function(module, target, fn_name)
         mf = mmi.machine_function(fn_name)
-        builder = MachineIRBuilder(mf)
-
-        _builder_stack.append(builder)
-        try:
+        with MachineIRBuilder(mf):
             args = [
                 MachineValue(mf.create_generic_virtual_register(llt), llt)
                 for llt in param_llts
             ]
             f(*args)
-        finally:
-            _builder_stack.pop()
-        return DSLMachineFunction(mmi, fn_name)
+        return DSLMachineFunction(mmi, fn_name, mf)
 
     return decorator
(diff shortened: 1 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #583 (`@machine_function` DSL + `MachineValue`)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`src/llvm/dsl/machine.py`, `dsl/__init__.py`, `tests/mir/test_dsl_machine.py`). This is a clean, minimal Pythonic front end that faithfully mirrors the `@function`/`ArithValue` idiom — low cognitive cost for anyone who knows the IR DSL, correct reflected-operator handling, sound `try/finally` builder-stack cleanup, and accurate docstrings. The dominant finding is a single **type-safety guard that exists in the sibling `ArithValue` but was dropped here**, which then shows up across the code, error-handling, type, and test reviews.

> Review-only — nothing was changed. Line numbers refer to `machine.py` unless noted.

---

### 🔴 High — dropped type-agreement guard → silently invalid MIR

**1. `_binary` never checks that two `MachineValue` operands share an LLT.** (`machine.py:44-49`)
This is the headline issue, flagged independently by the code, type-design, and silent-failure reviews. The sibling `ArithValue._binary` (`dsl/values.py:49-50`) raises `TypeError` when `self.type != other.type`; `MachineValue` omits any equivalent. So `a_s32 - b_s64` emits `build_sub(self.llt=s32, a.reg, b.reg)` — a type-mismatched generic instruction — and stamps the result with `self.llt` regardless of the operands' real types. In a debug build the MachineVerifier would assert; **in the shipped release wheel those asserts are compiled out (`NDEBUG`)**, so this constructs structurally invalid MIR that miscompiles or crashes far downstream, with no DSL-level error. Recommend restoring the `self.llt != other.llt → TypeError` guard the design says it mirrors.

**2. `_coerce` silently truncates non-int operands.** (`machine.py:38-41`)
`int(other)` accepts anything `int()` can consume: `a + 3.7` → a `G_CONSTANT` of `3`, `a + True` → `1`, `a + "5"` → `5`, all with no warning. The docstring says "a Python int becomes a G_CONSTANT," but nothing enforces that. `ArithValue._coerce` distinguishes float vs int by operand type. A float operand silently truncating into an integer constant is a silent-correctness footgun that survives into the release wheel. Recommend restricting to exact `int` (excluding `bool`) and raising a domain-specific `TypeError` naming the operand and target LLT otherwise.

---

### 🟠 Important

**3. `MachineValue` has no owner; leaked values emit into the wrong builder.** (`machine.py:19-25`, `44-49`)
A `MachineValue` carries only `reg` + `llt`, no reference to its builder/MachineFunction — every operator resolves the builder from module-global `_builder_stack`. If a value escapes its `@machine_function` body and is later operated on: with an empty stack it raises `RuntimeError` (good, and tested); but with a *different* (e.g. nested) `@machine_function` on the stack, `current_machine_builder()` returns the **wrong builder** and silently emits instructions/constants into a different MachineFunction, referencing vregs that belong to the original — cross-function corruption with no error. This is the same ambient-context shape `ArithValue` uses (so not worse than precedent), but `ArithValue` piggybacks on the shared C++ InsertPoint stack whereas this introduces a parallel Python-only stack. Consider anchoring each `MachineValue` to its owning builder and validating it in `_binary`/`_coerce`.

**4. `(reg, llt)` consistency is caller-asserted, never derived from MRI.** (`machine.py:32-34`)
`__init__` is a bare field-assign with no validation; `llt` is a stored copy, not read back from `MachineRegisterInfo::getType(reg)`. Every path the module itself takes keeps them consistent (regs are minted with the llt they're paired with), but a single stray `MachineValue(reg, wrong_llt)` silently poisons all downstream ops. Contrast `ArithValue`, which is a `Value` subclass reading its type directly off the binding, so it *cannot* desync. Storing `llt` separately signals it can drift.

---

### 🟡 Suggestions

**5. Partial `MachineFunction` leaks into `mmi` if the body raises.** (`machine.py:104-116`)
The `try/finally` correctly pops the stack and re-raises (no swallowing) — but on exception `mmi` retains a partially-built function named `fn_name` with whatever instructions were emitted before the failure, and there's no rollback. Since `create_machine_function` consumes the module, retry paths are limited, but at minimum document that a failed trace leaves partial state.

**6. `f(*args)` breaks for keyword-only / var-args params.** (`machine.py:104-113`)
The decorator iterates `sig.parameters.values()` uniformly and calls positionally: a keyword-only param makes `f(*args)` raise, and `*args: s32` is either treated as one fixed vreg or rejected with a misleading message. `func.py` shares this limitation (not a regression), but a `p.kind` guard with a clear error — or a docstring note — would help. Also, an unannotated param surfaces as `got <class 'inspect._empty'>`, which is opaque; a "parameter `p` is missing an LLT annotation" message would be clearer.

**7. `DSLMachineFunction.machine_function` re-resolves by name each call.** (`machine.py:77-79`)
It re-looks-up by name on every access (returning a fresh wrapper each time, so identity comparisons across accesses won't hold) and depends on unverified name-uniqueness, rather than caching the `mf` already obtained in the decorator. `DSLFunction` holds the resolved object directly. Also relevant: if `create_machine_function`'s `Function::Create` auto-renames on a collision with a function already in the module (`f`→`f.1`), this by-original-name lookup would resolve the wrong function; consider asserting the created function's name matches `fn_name`.

**8. Minor:** `_builder_stack` is a plain module global (not thread-safe — low impact, single-threaded in practice; a one-line caveat would help); the DSL `return` value is discarded (no terminator emitted, unlike `@function`'s `builder.ret(...)`) — fine for generic-MIR fragments but worth a docstring note since `return a + b` in the tests reads as meaningful; and two small docstring imprecisions — `DSLMachineFunction` "the built MachineFunction plus the MMI" (it stores `mmi`+`name` and resolves the MF lazily) and `machine_function`'s docstring referring to `f` (a param of the inner `decorator`, not the decorator factory).

---

### 🧪 Tests — broad happy path, but operand order and stateful edges untested

Every operator (forward + all three reflected), both int-coercion positions, both documented error paths, `MachineValue` attributes, `to_mir`, and default naming are exercised, all with `assert_no_leaks()` and `triple=None` (runs on any runner). But the suite has one structural weakness that undercuts its purpose:

- **No test inspects instruction operands — only `opcode_name`.** The entire reason the reflected operators exist is operand *order*, and subtraction is the only operator where order is observable. Nothing asserts that `10 - a` emits `G_SUB(const, a)` vs `a - 10` emitting `G_SUB(a, const)`. If the `reflected` flag were dropped from `__rsub__` or the `lhs,rhs` ternary inverted, **no test would fail**. Highest-value gap: assert the def/use registers of the emitted `G_SUB` in both forward and reflected forms.
- **`_builder_stack` cleanup after an exception in the body is untested** — `test_unannotated_parameter_raises` raises *before* the push, so the `finally` path isn't covered.
- Untested: nested/re-entrant `@machine_function`; mismatched-LLT operands (finding #1); non-int/float coercion (finding #2); explicit `name=` override (only the default branch is covered); and constant *values* (three `G_CONSTANT`s are counted but never checked to carry 1/2/10). `test_reflected_operators` also uses membership (`"G_ADD" in ops`) rather than the exact-sequence assertion used elsewhere — inconsistent rigor.

---

### ✅ Handled well
- **Reflected operators are correct**: `_coerce` stamps the constant with `self.llt` first, then `(other, self)` ordering gives `const - self` for `10 - a` — verified against the tests.
- **Exception cleanup is sound**: `try/finally` pops the builder even when the body raises, and the exception propagates unchanged — no swallowing, no empty `except`.
- **`_resolve_llt`** fails loud: accepts only `LLT`, raises `TypeError` with the offending annotation's `repr`.
- **CLAUDE.md-compliant**: imports at module top, `assert_no_leaks()` without preceding `gc.collect()`, `build_*` names mirror LLVM, and **no forward-reference/"Phase N" comments** anywhere in the code (despite the PR being fourth in the stack).
- **Docstrings are unusually accurate** — including the subtle "ints coerce to a G_CONSTANT of the other operand's type," which holds under tracing of both `a + 1` and `1 + a`.

---

makslevental:
Addressed the automated review. Changes are in the updated `dsl-func` commit and cascaded up the rest of the stack (#584–#591).

**High — restored the type-safety guards the sibling `ArithValue` has** (under `NDEBUG` the MachineVerifier asserts are gone, so these built structurally invalid MIR with no Python-level error):
- **#1** `MachineValue._binary` now raises `TypeError` on mismatched-LLT operands (`self.llt != other.llt`), matching `ArithValue._binary`.
- **#2** `_coerce` accepts only an exact `int` (not `bool`/`float`/`str`) and raises a domain-specific `TypeError` naming the operand and target LLT — `int(other)` was silently truncating `3.7` → `G_CONSTANT 3`, `True` → `1`, `"5"` → `5`.

**Important**
- **#3** Each `MachineValue` now anchors to the builder that created it and validates it in `_coerce` (which `_binary` and the upstack `_cmp` both call first). A value that escapes its `@machine_function` body — e.g. reused inside a nested function — now raises instead of silently emitting into the wrong builder with cross-function vregs. (The C++ `requireVRegOfType` guard from #582 doesn't catch this when the foreign vreg's index happens to collide with a same-typed vreg in the other function, so it needs the DSL-level anchor.)
- **#4** documented in the same anchoring comment; the drift risk is now backstopped by #3 plus #582's `requireVRegOfType` (which reads the real `MachineRegisterInfo::getType`).

**Suggestions**
- **#6** `machine_function` rejects non-positional params (`*args`/`**kwargs`/keyword-only) and gives a clear "missing an LLT annotation" message for unannotated params.
- **#7** `DSLMachineFunction` caches the `MachineFunction` resolved in the decorator instead of re-looking-it-up by name each access. (#582 already made `create_machine_function` reject a duplicate name, so the auto-rename-collision concern is closed at the source.)
- **#5 / #8** docstrings now note: the module-global builder stack is not thread-safe, the body's `return` is not turned into a terminator (straight-line fragment), and a failed trace leaves partial state in `mmi` (no rollback).

**Tests added:** mismatched-LLT operands, non-int coercion (float/bool/str), cross-function value reuse, builder-stack cleanup after the body raises, keyword-only param rejection, non-LLT annotation, explicit `name=` override, and — the highest-value gap — `G_SUB` operand order asserted by register identity in **both** forward (`a - 10`) and reflected (`10 - a`) forms, so dropping the reflected flag would now fail a test. `dsl/machine.py` Python coverage is 100%.

Note on `#2`: I couldn't assert the `G_CONSTANT` *values* (1/2/10) as the review suggested — a `G_CONSTANT` stores a `CImm` operand, which the current `MachineOperand.imm` binding doesn't read (it's for `MO_Immediate`). The operand-order tests pin the structure via register identity instead; a constant-value read-back would need a new binding, out of scope here.

Verified: the cumulative top of the stack builds and all 85 `tests/mir` pass; full Python coverage clears the 99% gate.

projects/eudsl-llvmpy/src/llvm/dsl/machine.py
```diff
@@ -0,0 +1,165 @@
...
+Register plus its LLT and overloads `+ - *` onto build_add/build_sub/build_mul;
+Python ints coerce to a G_CONSTANT of the other operand's type.
+"""
+
+import inspect
+
+from ..eudslllvm_ext.mir import LLT, MachineIRBuilder, create_machine_function
+
+# The active MachineIRBuilder is tracked on a plain module-global stack (not
+# thread-local); @machine_function tracing is assumed single-threaded.
```
makslevental: assumed by whom? you or LLVM?

makslevental: By us — this DSL layer, not LLVM. It's a property of *this* module's design: `_builder_stack` is a plain module-global, so two threads tracing `@machine_function` bodies at once would clobber the same stack. I chose that (mirroring how the IR DSL tracks its "current builder") over thread-local state for simplicity. It is **not** an LLVM-documented guarantee — though separately, LLVM's MachineIRBuilder/MachineFunction aren't safe for concurrent building either, so thread-local state wouldn't buy real parallelism here anyway.

I'll reword the comment to make the source of the constraint explicit, e.g. "tracked on a module-global stack, so tracing must be single-threaded" rather than the passive "is assumed single-threaded."

makslevental: Correction to my previous reply — I had this backwards. The IR DSL's `current_builder()`/`InsertPoint` stack **is** thread-local (`src/IR/Builder.cpp`: `static thread_local std::vector<ThreadContextEntry> stack;`). So this Machine DSL does *not* mirror it: `_builder_stack` is a plain Python module-global, which is strictly weaker.

So the accurate answer to "assumed by whom?": by this DSL layer — and it's a regression from the IR DSL, which handles concurrent tracing correctly via thread-local state. The right fix is to make `_builder_stack` a `threading.local()` so the Machine DSL matches the IR DSL rather than being a weaker special case. I'll do that (and reword the comment accordingly).

makslevental: Done — implemented the thread-local tracking in C++ to mirror the IR bindings rather than papering over it in Python. `MachineIRBuilder` now has `__enter__`/`__exit__` backed by a `thread_local std::vector<llvm::MachineIRBuilder *>` in Machine.cpp (9d8162f2, on the build-generic PR), plus a `current_machine_builder()` free function — the direct analogue of `with builder:`/`current_builder()`. machine.py (7024d242) drops its module-global `_builder_stack` and just uses `with MachineIRBuilder(mf):` + the C++ `current_machine_builder()`. So @machine_function tracing is now thread-safe in the same way the IR DSL is, and the comment you flagged is gone.

PR #584, opened 2026-08-17
[eudsl-llvmpy] Add generic-MIR control-flow primitives (blocks, branches, G_PHI)

makslevental:
Extend the MIR object model and MachineIRBuilder with the pieces control flow
needs: MachineFunction.create_block, MachineBasicBlock.add_successor, and the
builder's insert_block/set_block, build_icmp (G_ICMP, reusing ir.CmpPredicate),
build_br/build_brcond (G_BR/G_BRCOND), and build_phi (G_PHI from
(value, predecessor) pairs). Tested by hand-building an if-diamond CFG.

Fifth PR in the MIR bindings/DSL stack; the substrate for the control-flow DSL.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -20,6 +20,7 @@
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/CodeGen/TargetSubtargetInfo.h>
 #include <llvm/CodeGenTypes/LowLevelType.h>
+#include <llvm/IR/BasicBlock.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DerivedTypes.h>
 #include <llvm/IR/DiagnosticInfo.h>
@@ -97,6 +98,28 @@ void requireVRegOfType(llvm::MachineIRBuilder &b, llvm::Register reg,
                            " must be a virtual register of this "
                            "MachineFunction with the result type")
                               .c_str());
+}
+
+// A MachineBasicBlock / Register crosses the Python boundary as a bare handle
+// with no back-link to its function; passing one from a different function
+// builds corrupt MIR (cross-linked CFGs, out-of-bounds vreg indexes) that the
+// verifier -- gone under NDEBUG -- would otherwise catch. Guard provenance
+// against the builder's (or another block's) function.
+void requireSameFunction(const llvm::MachineFunction &mf,
+                         const llvm::MachineBasicBlock *mbb, const char *role) {
+  if (mbb->getParent() != &mf)
+    throw nb::value_error(
+        (std::string(role) + " belongs to a different MachineFunction")
+            .c_str());
+}
+
+void requireVReg(llvm::MachineIRBuilder &b, llvm::Register reg,
+                 const char *role) {
+  if (!b.getMF().getRegInfo().getType(reg).isValid())
+    throw nb::value_error(
+        (std::string(role) +
+         " must be a generic virtual register of this MachineFunction")
+            .c_str());
 }
 
 // The "current MachineIRBuilder" is tracked on a thread-local stack, mirroring
@@ -501,6 +524,9 @@ void populate_mir(nb::module_ &m) {
       .def(
           "add_successor",
           [](llvm::MachineBasicBlock &self, llvm::MachineBasicBlock *succ) {
+            // addSuccessor also calls succ->addPredecessor(this), so a
+            // cross-function successor would corrupt two functions' CFGs.
+            requireSameFunction(*self.getParent(), succ, "successor");
             self.addSuccessor(succ);
           },
           "successor"_a, "Add a CFG successor edge to another block.")
@@ -540,13 +566,16 @@ void populate_mir(nb::module_ &m) {
           "type"_a, "Create a new generic virtual register of the given LLT.")
       .def(
           "create_block",
-          [](llvm::MachineFunction &self) -> llvm::MachineBasicBlock * {
-            llvm::MachineBasicBlock *mbb = self.CreateMachineBasicBlock();
+          [](llvm::MachineFunction &self,
+             const llvm::BasicBlock *bb) -> llvm::MachineBasicBlock * {
+            llvm::MachineBasicBlock *mbb = self.CreateMachineBasicBlock(bb);
             self.push_back(mbb);
             return mbb;
           },
+          "basic_block"_a.none() = nb::none(),
           nb::rv_policy::reference_internal,
-          "Append a new, empty MachineBasicBlock to the function.")
+          "Append a new, empty MachineBasicBlock to the function, optionally "
+          "linked to an IR BasicBlock for debug info/naming.")
       .def("__str__",
            [](llvm::MachineFunction &self) { return eudsl::toString(self); });
 
@@ -823,6 +852,7 @@ void populate_mir(nb::module_ &m) {
       .def(
           "set_block",
           [](llvm::MachineIRBuilder &self, llvm::MachineBasicBlock *mbb) {
+            requireSameFunction(self.getMF(), mbb, "block");
             self.setMBB(*mbb);
           },
           "block"_a, "Insert subsequent instructions at the end of `block`.")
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## Comprehensive PR Review — #584 (Generic-MIR control-flow primitives)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`src/MIR/Machine.cpp` + `tests/mir/test_cf_primitives.py`). The new primitives are clean, thin wrappers, `create_block` establishes ownership correctly (`CreateMachineBasicBlock` + `push_back`, returned `reference_internal`), `build_phi` gets the operand interleave right and creates its own def register, and the docstrings are accurate. The recurring theme from earlier in this stack continues: **the shipped wheel is a release (`NDEBUG`) build, so the LLVM `assert`s these bindings delegate correctness to are compiled out** — turning user error into silently-malformed MIR rather than a Python exception. Two direct CLAUDE.md naming/API deviations are also worth addressing before merge.

> Review-only — nothing was changed. Line numbers are approximate to the post-PR `Machine.cpp`.

---

### 🟠 Important — CLAUDE.md compliance

**1. `insert_block` / `set_block` don't mirror the LLVM API.** (`Machine.cpp` ~402, ~409)
They wrap `MachineIRBuilder::getMBB()` / `setMBB()`. CLAUDE.md ("Mirror the LLVM C++ API") forbids renaming a method "because it reads as semantically equivalent," with `addSuccessor→add_successor` as the exemplar. There is no `getInsertBlock`/`setInsertBlock` on `MachineIRBuilder`, so `insert_block` is invented terminology — recommend `mbb` (prop) / `set_mbb` to track the actual spelling. (`add_successor`, `build_br/brcond/phi/icmp` all correctly mirror their LLVM names.)

**2. `create_block` hardcodes the IR `BasicBlock` association.** (`Machine.cpp` ~231)
`CreateMachineBasicBlock()` is called with no args, silently baking the optional `const BasicBlock *BB = nullptr` (which links the MBB to its IR block for debug info/naming) to null. Per CLAUDE.md "bind the full API; do not hardcode," this should be surfaced as an argument defaulting to `None`. (Null is fine as the *default* for hand-built MIR — the issue is that the choice isn't exposed.)

---

### 🔴 Release-build safety (asserts vanish under `NDEBUG`)

The `None`→`nullptr` case is *not* a concern — nanobind's type layer rejects `None` for these raw-pointer/Register args with a `TypeError` (only `std::optional` args accept none). But valid-but-wrong inputs are unguarded:

**3. Cross-`MachineFunction` handles accepted everywhere.** (`add_successor` ~199, `set_block` ~409, `build_br` ~423, `build_brcond` ~429, `build_icmp` ~415, `build_phi` ~435)
Every `MachineBasicBlock*` / `Register` crosses the boundary as a bare handle with no back-link to its owning function, and nothing checks provenance. Passing a block/register from a *different* function compiles fine and produces corrupt MIR. Notably: `set_block` with a foreign block redirects the insertion point while `getMRI()` still allocates from the original function (cross-linking two functions); `build_icmp`/`build_phi` `addUse` a vreg number that indexes a *different* function's register table (out-of-bounds → possible hard crash); and `add_successor` is worst — `addSuccessor` also calls `succ->addPredecessor(this)`, so one cross-function call mutates *two* functions' CFGs. This is the same unmitigated hazard flagged on the earlier `build_*` PRs; a cheap `getMF()`/`getParent()` same-function guard throwing `nb::value_error` would close it.

**4. `build_phi` builds structurally invalid MIR silently.** (`Machine.cpp` ~435-449) — two distinct gaps:
- **Empty `incomings`** → emits `%res = G_PHI` with a def and zero pairs, returning a live `Register` as if well-formed.
- **Incoming count vs. predecessor count is never checked** — a G_PHI needs exactly one `(value, pred)` per CFG predecessor of the insertion block; nothing reads `getMBB().predecessors()`, so too few/too many/wrong blocks all slide through. The test passes 2 entries matching 2 predecessors, masking the gap. Also (lower severity) the phi is inserted at the block *end* (builder default), but G_PHIs must lead the block — works only because `join_bb` is empty when built.

**5. `build_icmp` preconditions are all assert-gated.** (`Machine.cpp` ~415)
`buildICmp` asserts (a) result LLT is `s1`/vector-of-`s1`, (b) operand types consistent, (c) integer predicate. All compiled out in release. Note `CmpInst::Predicate` spans FCMP_* too, so a float predicate could be passed and encoded into a `G_ICMP` with no error. A non-`s1` result type or mismatched operands likewise slide through.

**6. Branch terminators don't maintain the CFG edge.** (`build_br` ~423, `build_brcond` ~429)
`build_br/brcond` emit the instruction but don't call `addSuccessor`, so the successor list and terminator operands are two independent sources of truth the user must hand-sync (the test sets both manually). This *faithfully mirrors LLVM's model* (they genuinely are separate), so it's not a binding bug — but it's an easy silent inconsistency, and a combined helper (branch that also wires the successor) alongside the raw ops would prevent drift. Flagged by both the type and silent-failure reviews as a design smell the binding *could* mitigate.

---

### 🟡 Suggestions — type/API shape

**7. `insert_block` (read-only property) vs `set_block` (method) is an incoherent get/set shape.** (`Machine.cpp` ~402 vs ~409)
A property you can read but not assign, whose writer is a differently-named verb, breaks the uniformity of the `build_*` family. Defensible LLVM-wise (getter pure, setter has side effects) but a read/write property or a matched `current_block`/`set_current_block` pair would read as one piece of mutable state. *(This overlaps with finding #1's naming fix.)*

**8. `build_phi`'s pair signature is the PR's best design choice** — `list[(Register, MachineBasicBlock)]` communicates "one value per edge" far better than parallel lists, and returning the def Register composes with the other `build_*` ops. The gap is purely enforcement (#4): the type invariant (incoming LLT == `ty`, arity == predecessors) is invisible and unchecked, leaving correctness to a verifier that never runs.

---

### 🧪 Tests — every binding invoked, but almost nothing structurally verified

Both tests use `assert_no_leaks()` and `triple=None` (host, portable for target-independent G_*), and `test_create_block_and_set_block` is genuine behavioral coverage (block count + `set_block`↔`insert_block` round-trip via `.number`). But the if-diamond test asserts only block count (`==4`), `phi.is_virtual`, and **opcode-name substring membership in `str(mf)`** — which is simultaneously brittle (coupled to the text dump) and weak (an opcode appearing anywhere passes, regardless of operands/targets/block). Unverified, in priority order:
- **G_PHI operand structure/order** (priority 8) — the one hand-rolled instruction, most likely to regress; swapping use/block order or dropping a pair would still pass. `MachineBasicBlock.instructions` + operand accessors make a structural check reachable without string parsing.
- **Branch targets** — `"G_BR"/"G_BRCOND" in text` can't distinguish a branch to the wrong block.
- **Successor edges** — `add_successor` is called 4× but the successor lists are never read back (zero behavioral assertion).
- **icmp predicate** — `SLT` is passed but never confirmed encoded (the enum→`CmpInst::Predicate` conversion is untested for correctness); only one predicate is exercised.
- **Multi-block `to_mir`/round-trip** of this CFG (G_PHI/branch serialization breaks silently), per-block instruction placement, and `build_phi` with empty incomings.

Swapping the substring assertions for structural checks over the API would close most of these at once.

---

### ✅ Handled well
- `create_block` correctly establishes ownership (`CreateMachineBasicBlock` + `push_back`, `reference_internal`).
- **`build_phi` operand ordering is correct** — `addDef(res)` then interleaved `addUse(reg); addMBB(mbb)` per pair matches the G_PHI layout exactly; it also creates its own def register internally rather than trusting the caller.
- `None` args are safely rejected by nanobind's type layer (`TypeError`), so there's no null-deref path.
- Brace style compliant (the multi-line `for` body is braced), imports at module top, `nanobind/stl/pair.h` + `<utility>` correctly added, `CmpInst::Predicate` reuse via `ir.ICmpPredicate` is consistent.
- All binding docstrings verified accurate against their implementations.

*(Comment nit: `test_cf_primitives.py:5-6` says these are "the pieces the @machine_function control-flow runtime is built on" — a forward reference to unlanded work (that runtime doesn't reference these primitives in this branch), which CLAUDE.md's "no forward-reference comments" rule discourages. Reword to describe what exists now.)*

---

makslevental:
Addressed the automated review. Changes are in the updated `cf-primitives` commit and cascaded up the rest of the stack (#585–#591).

**Release-build safety** (the primitives delegated correctness to LLVM asserts, which are compiled out under `NDEBUG` in the wheel, so bad input built corrupt MIR with no Python-level error):
- **#3 cross-function handles:** `add_successor`, `set_block`, `build_br`, `build_brcond`, `build_phi` now reject a `MachineBasicBlock` from a different `MachineFunction` (`requireSameFunction`); `build_brcond`/`build_icmp`/`build_phi` reject a `Register` that isn't a generic vreg of this function. `add_successor` was the worst case — it also calls `succ->addPredecessor(this)`, so a cross-function edge corrupted two functions' CFGs; that's now closed.
- **#4 `build_phi`:** rejects empty incomings (a def-only G_PHI is structurally invalid) and validates each incoming value's type + each predecessor block's provenance. I intentionally did **not** enforce arity == actual predecessor count: the upstack loop runtime (#586) builds header phis incrementally (`build_empty_phi` + `add_phi_incoming`) before the back-edge exists, so a strict arity check would be wrong there.
- **#5 `build_icmp`:** rejects a non-integer predicate (`CmpInst::Predicate` also spans `FCMP_*`), a non-`s1`/non-vector-of-`s1` result type, and mismatched operand types — all previously assert-only.

**CLAUDE.md**
- **#2 `create_block`:** surfaced the optional IR `BasicBlock` association (`CreateMachineBasicBlock`'s `const BasicBlock*`) as a `basic_block=None` arg instead of hardcoding null.
- Dropped the forward-reference in the test docstring.

**Tests:** replaced the brittle `"OPCODE" in str(mf)` substring checks with structural assertions — G_ICMP and G_PHI operand order verified by **register identity** (swapping the phi pairs or the icmp operands now fails a test), plus negative-path tests for every new guard (float predicate, non-s1 result, mismatched icmp operands, empty/mistyped phi incomings, cross-function block, cross-function register) and the `create_block` IR-BasicBlock arg. `src/MIR` stays at 100% line + function coverage.

**Not applied — with reasoning:**
- **#1 / #7 (rename `insert_block`/`set_block` → `mbb`/`set_mbb`):** `insert_block` is the established cross-builder convention — the IR `IRBuilder` binds `GetInsertBlock` as `insert_block` (`src/IR/Builder.cpp`), and the whole DSL (`cf.py`, `machine_cf.py`) plus IR/MIR tests use `b.insert_block` / `b.set_block` uniformly. Renaming only the MIR side would fracture that shared vocabulary, and the getter-property/setter-method asymmetry is inherited from the IR builder rather than introduced here. I'd rather keep MIR and IR consistent; happy to rename both in a dedicated cross-cutting change if you prefer.
- **#6 (a branch helper that also wires the successor):** the separate successor list / terminator operands faithfully mirror LLVM's model; left as a possible future convenience rather than changing the 1:1 bindings.
- **#8:** the pair signature is kept; the enforcement gap it noted is closed by #4.

Verified: PR #584's branch builds and its 67 `tests/mir` pass; the cumulative top of the stack builds and all 93 `tests/mir` pass.

PR #585, opened 2026-08-17
[eudsl-llvmpy] Add generic-MIR if/else control-flow runtime

makslevental:
Lower Python if/else in a @machine_function body to MachineBasicBlocks + G_PHI,
reusing the existing AST canonicalizer (llvm.ast) unchanged -- only the injected
runtime differs (MIRCanonicalizer). The MIR runtime (machine_cf._MIRIfOp) mirrors
the IR cf runtime: G_BRCOND cond->then plus a fall-through G_BR whose target is
repointed merge->else by else_ctx_manager (the MIR analogue of cond_br
set_successor), with yielded values becoming merge-block G_PHIs. Adds MachineValue
comparison operators (G_ICMP -> i1) and the supporting primitives (build_br
returns its instr, MachineInstr.set_branch_target, MachineBasicBlock.
replace_successor, MachineIRBuilder.machine_function).

Loops remain for a follow-up. Sixth PR in the MIR bindings/DSL stack.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -475,6 +475,16 @@ void populate_mir(nb::module_ &m) {
                    [](llvm::MachineInstr &self) { return self.mayStore(); })
       .def_prop_ro("is_debug_instr",
                    [](llvm::MachineInstr &self) { return self.isDebugInstr(); })
+      .def(
+          "set_branch_target",
+          [](llvm::MachineInstr &self, llvm::MachineBasicBlock *mbb) {
+            if (self.getNumOperands() == 0 || !self.getOperand(0).isMBB())
+              throw nb::value_error(
+                  "instruction has no branch-target (MBB) operand");
+            self.getOperand(0).setMBB(mbb);
+          },
+          "block"_a,
+          "Repoint a branch's target block (operand 0), e.g. a G_BR.")
       .def("__str__",
            [](llvm::MachineInstr &self) { return eudsl::toString(self); });
 
@@ -530,6 +540,13 @@ void populate_mir(nb::module_ &m) {
             self.addSuccessor(succ);
           },
           "successor"_a, "Add a CFG successor edge to another block.")
+      .def(
+          "replace_successor",
+          [](llvm::MachineBasicBlock &self, llvm::MachineBasicBlock *old,
+             llvm::MachineBasicBlock *replacement) {
+            self.replaceSuccessor(old, replacement);
+          },
+          "old"_a, "new"_a, "Replace a CFG successor edge with another block.")
       .def("__str__",
            [](llvm::MachineBasicBlock &self) { return eudsl::toString(self); });
 
@@ -856,6 +873,13 @@ void populate_mir(nb::module_ &m) {
             self.setMBB(*mbb);
           },
           "block"_a, "Insert subsequent instructions at the end of `block`.")
+      .def_prop_ro(
+          "machine_function",
+          [](llvm::MachineIRBuilder &self) -> llvm::MachineFunction * {
+            return &self.getMF();
+          },
+          nb::rv_policy::reference_internal,
+          "The MachineFunction this builder inserts into.")
       .def(
           "build_icmp",
           [](llvm::MachineIRBuilder &self, llvm::CmpInst::Predicate pred,
@@ -882,11 +906,14 @@ void populate_mir(nb::module_ &m) {
           "predicate"_a, "type"_a, "lhs"_a, "rhs"_a)
       .def(
           "build_br",
-          [](llvm::MachineIRBuilder &self, llvm::MachineBasicBlock *dest) {
+          [](llvm::MachineIRBuilder &self,
+             llvm::MachineBasicBlock *dest) -> llvm::MachineInstr * {
             requireSameFunction(self.getMF(), dest, "dest");
-            self.buildBr(*dest);
+            return self.buildBr(*dest).getInstr();
           },
-          "dest"_a, "Build an unconditional branch (G_BR) to `dest`.")
+          "dest"_a, nb::rv_policy::reference_internal,
+          "Build an unconditional branch (G_BR) to `dest`; returns the "
+          "instruction so its target can be repointed.")
       .def(
           "build_brcond",
           [](llvm::MachineIRBuilder &self, llvm::Register cond,
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -478,9 +478,10 @@ void populate_mir(nb::module_ &m) {
       .def(
           "set_branch_target",
           [](llvm::MachineInstr &self, llvm::MachineBasicBlock *mbb) {
-            if (self.getNumOperands() == 0 || !self.getOperand(0).isMBB())
+            if (self.getNumOperands() == 0 || !self.getOperand(0).isMBB()) {
               throw nb::value_error(
                   "instruction has no branch-target (MBB) operand");
+            }
             self.getOperand(0).setMBB(mbb);
           },
           "block"_a,
@@ -544,6 +545,13 @@ void populate_mir(nb::module_ &m) {
           "replace_successor",
           [](llvm::MachineBasicBlock &self, llvm::MachineBasicBlock *old,
              llvm::MachineBasicBlock *replacement) {
+            // replaceSuccessor only asserts `old` is a successor (gone under
+            // NDEBUG); without the check it walks past succ_end() and corrupts
+            // the CFG. Guard it, and reject a cross-function replacement.
+            if (!self.isSuccessor(old)) {
+              throw nb::value_error("`old` is not a successor of this block");
+            }
+            requireSameFunction(*self.getParent(), replacement, "new");
             self.replaceSuccessor(old, replacement);
           },
           "old"_a, "new"_a, "Replace a CFG successor edge with another block.")
--- a/projects/eudsl-llvmpy/src/llvm/dsl/machine.py
+++ b/projects/eudsl-llvmpy/src/llvm/dsl/machine.py
@@ -13,6 +13,7 @@ ints coerce to a G_CONSTANT of the other operand's type.
 
 import inspect
 
+from ..eudslllvm_ext.ir import ICmpPredicate
 from ..eudslllvm_ext.mir import (
     LLT,
     MachineIRBuilder,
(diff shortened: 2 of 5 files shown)


makslevental (commented):
## Comprehensive PR Review — #585 (Generic-MIR if/else control-flow runtime)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`Machine.cpp`, `dsl/machine.py`, new `dsl/machine_cf.py`, two test files). This is the most intricate PR in the stack, and the core lowering is **correct where it counts**: the general-code review independently verified that the diamond CFG stays consistent with its terminators (entry `{then, merge}` → repointed to `{then, else}` atomically via `replace_successor` + `set_branch_target`), that phi predecessors are captured correctly even under nested control flow (`b.insert_block` at each branch's trailing `yield_`), that `_if_stack` push/pop is balanced, and that the `__eq__`/`__ne__`-identity + `eq()`/`ne()` design faithfully mirrors `ArithValue`. Comments are accurate throughout. The findings below are about hazards the runtime leaves *unenforced* — several of which turn user mistakes into silent miscompiles rather than errors, continuing the release-build (`NDEBUG`) theme from earlier in the stack.

> Review-only — nothing was changed. `machine_cf.py` line numbers are from the new file.

---

### 🔴 High — silent correctness traps

**1. `if a == b:` silently lowers to a constant-false branch.** (`machine.py:88-98`)
`<`/`<=`/`>`/`>=` are overloaded to emit `G_ICMP`, but `==`/`!=` are deliberately kept as object *identity* (with `__hash__ = object.__hash__`), and value-equality is exposed only as named `eq()`/`ne()`. This faithfully mirrors `ArithValue` and is arguably forced by Python (overloading `__eq__` would break hashing/`in`/`== None` in traversal) — so the design is coherent. **But** in a `@machine_function` body the canonicalizer captures `if a < b:` and lowers it, whereas `if a == b:` is *not* overloaded: it evaluates Python identity → constant `False` for two distinct `MachineValue`s → a compile-time-false branch instead of a `G_ICMP EQ`. No guard, no warning; the user must remember `a.eq(b)`. This is sharper than for `ArithValue` because it lands in the most common construct. Worth an explicit docstring warning, or ideally a canonicalizer lint rejecting a bare `==`/`!=` as an `if` condition on a `MachineValue`.

**2. Loops appear not to be rejected despite the docstring promising it.** (`machine_cf.py` module docstring + `MIRCanonicalizer`)
The docstring says a `for`/`while` in a `@machine_function` body "is unsupported," and `MIRCanonicalizer.cst_transformers` lists `RejectUnsupportedJumps` but no loop transformer. The silent-failure review traced `RejectUnsupportedJumps` (`ast/cf_transformers.py:358-382`) and reports its `visit_While`/`visit_For` only walk for a nested `Return` and otherwise `generic_visit` — i.e. **it does not actually reject loops**. If that reading is correct, a `for i in range(n):` silently unrolls into straight-line MIR (wrong, no error) and a `while a < b:` hangs the trace in an infinite Python loop (`a < b` is a truthy `MachineValue`, no `__bool__`). *This one finding is from a single agent's reading of `RejectUnsupportedJumps` and wasn't cross-verified — please confirm — but if it holds, the "unsupported" contract is silently violated.* A test feeding a loop into a `@machine_function` body and asserting it raises would both settle this and lock the contract.

**3. `replace_successor` is unguarded → silent CFG/heap corruption in release.** (`Machine.cpp`, new `replace_successor` binding)
`MachineBasicBlock::replaceSuccessor` validates "old is a successor" only via `assert`, which is compiled out in the release wheel. If `old` isn't a successor, `OldI == succ_end()` and the code writes through / erases past the end iterator → UB, silently. The current in-tree caller (`else_ctx_manager`) always passes a real successor, so today's path is safe — but it's a public, unguarded footgun. Notably, the sibling `set_branch_target` in *this same PR* does exactly the right thing (guards operand 0 is an MBB, throws `nb::value_error`, has a negative test) — `replace_successor` should copy that pattern.

**4. Mismatched then/else branches fail silently rather than loudly.** (`machine_cf.py:82-90`)
Two sub-issues in `record_and_maybe_phi`'s `for tv, ev in zip(self.then_vals, self.else_vals)`:
- **Arity:** `zip` truncates to the shorter list, so if the branches yield different counts, extra carried values are silently dropped (surfacing later as an opaque unpack error). The general-code review judged this *low-risk* because the canonicalizer guarantees matching arity; the silent-failure review rated it HIGH as a latent footgun. Both agree the fix is a `len(then_vals) == len(else_vals)` check — cheap insurance regardless of severity. (The sibling `cf.py` indexes `else_vals[i]`, which would at least raise.)
- **Type:** the phi result type is taken from `tv.llt` with no check that `ev.llt == tv.llt`; divergent branch LLTs build type-corrupt MIR silently (verifier compiled out).

**5. Comparisons hardcode signed predicates with no unsigned path.** (`machine.py:70-86`)
`__lt__`/`<=`/`>`/`>=` always emit `SLT`/`SLE`/`SGT`/`SGE`; there's no way to request `ULT`/etc. A user comparing semantically-unsigned values silently gets signed compares (wrong results). The docstring says "default to signed," but with no override it's a baked-in assumption — brushing against CLAUDE.md's "bind the full API; do not hardcode assumptions." (This mirrors `ArithValue`'s integer defaults, so it's a consistency call, not a regression.)

---

### 🟠 Important

**6. Dropped LLT-match guard vs the mirrored `ArithValue`.** (`machine.py:45-50`, `70-76`)
`MachineValue._binary`/`_cmp` don't validate that operand LLTs match, whereas `ArithValue._binary`/`_cmp` (`values.py:49-50, 57-58`) raise `TypeError` on mismatch. Comparing/combining an `s32` with an `s64` `MachineValue` silently emits mismatched-operand MIR. This is the one runtime guard the mirror dropped — flagged independently by the code, type, and silent-failure reviews.

**7. CLAUDE.md C++ brace style.** (`Machine.cpp`, `set_branch_target` guard)
The `if (self.getNumOperands() == 0 || !self.getOperand(0).isMBB()) throw nb::value_error(...)` has a multi-line `throw` body without braces; CLAUDE.md requires braces when the body spans multiple lines (its example is exactly this). `clang-format` won't add them.

**8. Forward-reference comments.** (`machine_cf.py` module docstring; `MIRCanonicalizer` inline comment)
"Loops are not lowered here **yet** … unsupported **until the loop runtime lands**" and "# if/else only **for now**" both point at future work, which CLAUDE.md's "No forward-reference comments" rule bans ("activates once X lands" is the cited example). Reword to describe the present state without "yet"/"until … lands"/"for now".

**9. `yield_` outside an if body raises a bare `IndexError`.** (`machine_cf.py:130`)
`_if_stack[-1]` with an empty stack gives `IndexError: list index out of range` — since `_InjectMIRCFGlobals` permanently injects `yield_` into the traced function's globals, this is reachable. Contrast `current_machine_builder()`, which raises a clear `RuntimeError`. The sibling `cf.py` also carries a comment explaining why the stack is always populated; that reassurance was dropped here.

---

### 🟡 Suggestions — type/design

**10. `_MIRIfOp` maintains a dual source of truth for the CFG.** (`machine_cf.py:52-53, 63-64, 116-117`)
Unlike LLVM IR (where successors derive from the terminator), MIR tracks successor edges separately, so every edge change must mutate *both* the CFG edge and the branch instruction, by hand, at three sites (`__init__`, `_terminate_current`, `else_ctx_manager`). There's no encapsulated atomic "repoint the false edge" helper, so forgetting one half produces CFG/terminator disagreement that passes silently until a verifier. Encapsulating each edge mutation in one helper would remove a class of latent bugs. Relatedly, `_MIRIfOp` is a bag of mutable fields advanced by an implicit then-before-else protocol (`else_block is None` doubles as the no-else signal; out-of-order calls raise a cryptic `TypeError` on `zip(None, …)` rather than a guarded state check) — acceptable only because the canonicalizer generates the driving shape.

**Type-design ratings (delta as a whole):** encapsulation 6/10, invariant expression 6/10, usefulness 7/10, enforcement **4/10** — the weakest axis, driven by findings #1, #4, #6, and #10.

---

### 🧪 Tests — broad presence checks, shallow correctness

New tests cover the if/else diamond, if-without-else, comparison-produces-i1, all six predicates (exact `count == 6`), two-carried-values, and a good negative test for `set_branch_target`'s guard — all leak-checked and portable (`triple=None`). But nearly every if/else assertion is **opcode-name membership in `to_mir()` text + block count**, which would pass even if values were routed to the wrong branch. A structural limitation contributes: `MachineBasicBlock` exposes no `successors`/`predecessors` accessor and `MachineOperand` no MBB accessor, so CFG/target checks must currently go through the printed `.mir` text — but that text *does* contain the full `G_PHI %x, %bb.1, %y, %bb.2` lines, so a stronger assertion is available and just isn't used. Priority gaps:
- **PHI value/predecessor routing** (priority 9) — no test verifies the pairing; a then/else swap or wrong-predecessor bug passes. The two-carried-values test yields `a,b` then `b,a` (a classic swap) but asserts only `count == 2`.
- **Nested if/else** (priority 8) — the most subtle logic (moved-predecessor capture) has zero coverage.
- **Loop rejection** (priority 7) — untested; directly related to finding #2.
- **`if a == b:` identity behavior** — untested (finding #1).
- **Int-coerced / reflected comparison operands** (`a < 1`, `1 < a`) — untested; the reflected path silently changes the predicate (`1 < a` → `a.__gt__(1)` → SGT).
- **`replace_successor` / `set_branch_target` success-path effects** — the repointing is never asserted, only that it doesn't crash.

---

### ✅ Handled well
- **Diamond CFG is consistent with its terminators** in all cases (verified): entry successors match, `_terminate_current` adds the `pred→merge` edge alongside each `G_BR`, no stale `entry→merge` edge after repointing.
- **Phi predecessors are correct under nesting** — captured as `b.insert_block` at each branch's trailing `yield_`, so nested control flow that moved the builder still yields the true predecessor.
- **`set_branch_target` guard is correct** — for a `G_BRCOND` operand 0 is the condition register (not an MBB), so `isMBB()` is false and it correctly *rejects* rather than repointing the wrong operand; `G_BR` operand 0 is the target. Has a negative test.
- `_if_stack` push/pop is balanced across the two sequential context managers; body exceptions propagate loudly (no swallowing).
- `__eq__`/`__ne__` identity + `eq()`/`ne()` and the signed-predicate defaults faithfully mirror `ArithValue`; `build_br` returning `reference_internal MachineInstr*` is lifetime-safe.
- All comments/docstrings verified accurate against the code (only the two forward-reference phrasings are flagged).

---

makslevental:
Addressed the automated review. Changes are in the updated `cf-if` commit and cascaded up the rest of the stack (#586–#591).

**High — silent correctness traps**
- **#2 loops now rejected:** `MIRCanonicalizer` gains a `_RejectLoops` transformer, so a `for`/`while` in a `@machine_function` body raises `NotImplementedError` instead of silently unrolling (Python `range`) or hanging the trace (a truthy `MachineValue` condition). Confirmed the review's reading — `RejectUnsupportedJumps` only rejects nested `return`/`break`/`continue`, not loops. (The loop-runtime PR #586 replaces this rejection with real lowering.)
- **#3 `replace_successor` guarded:** it now checks `old` is actually a successor (`replaceSuccessor` only asserts it → walks past `succ_end()` under `NDEBUG`) and rejects a cross-function replacement, mirroring `set_branch_target`.
- **#4 mismatched then/else:** `record_and_maybe_phi` now validates the two branches yield the same **number** of values and matching **LLTs**, instead of `zip`-truncating / building type-corrupt phis.
- **#1 `if a == b:`:** documented that `==`/`!=` are Python identity (a compile-time `False`), so a value compare must use `a.eq(b)`/`a.ne(b)` — with a test pinning the identity behavior. (I kept the identity design, which mirrors `ArithValue` and is needed for hashability; a canonicalizer lint would risk false-positives on legitimate `== None` guards.)

**Important**
- **#6 dropped LLT-match guard:** `MachineValue._cmp` now raises on mismatched operand LLTs (matching `ArithValue._cmp` and the `_binary` guard already in place); a mismatched compare is also backstopped by `build_icmp`'s C++ operand-type check.
- **#7 brace style:** `set_branch_target`'s multi-line `throw` is braced.
- **#8 forward-references:** dropped "yet" / "until the loop runtime lands" / "for now" from `machine_cf.py`.
- **#9 `yield_` outside an if body:** now raises a clear `RuntimeError`, not a bare `IndexError`.

**Suggestions**
- **#5 unsigned path:** added `ult`/`ule`/`ugt`/`uge` named methods (the `< <= > >=` operators stay signed, mirroring `ArithValue`; Python has only one `<`, so unsigned is exposed by name).
- **#10 dual source of truth:** encapsulated the false-edge repoint (CFG successor + `G_BR` target) in `_MIRIfOp._repoint_false_edge` so the two halves can't drift. (The broader successor/terminator split is inherent to MIR and kept.)

**Tests:** G_PHI value routing verified **structurally** (register-identity swap check on the two-carried case, so a then/else swap now fails), nested if/else, for/while rejection, `==` identity, unsigned + reflected comparisons, mismatched-branch arity and type, mismatched comparison types, `yield_`-outside-body, and `replace_successor`'s guards (not-a-successor + cross-function). `src/MIR` stays at 100% line+function; `dsl/machine.py` and the if/else code in `dsl/machine_cf.py` at 100%.

Verified: PR #585's branch builds and its 85 `tests/mir` pass; the cumulative top builds and all `tests/mir` pass; full Python coverage clears the 99% gate.

PR #586, opened 2026-08-17
[eudsl-llvmpy] Add generic-MIR for/while loop lowering

makslevental:
Lower Python for/while in a @machine_function body to a preheader -> header
(phis) -> body -> exit MIR shape, mirroring llvm.dsl.cf._Loop with the same AST
transformers (ForToInline/WhileToInline, now enabled in MIRCanonicalizer). The
header's induction-variable and loop-carried G_PHIs are built empty at the block
top (build_empty_phi) so they precede the loop condition, seeded with the
preheader incoming, then given their back-edge incoming from the body's end
(MachineInstr.add_phi_incoming) -- the incremental-phi analogue of the IR
builder's phi + add_incoming. range_ (incl. single-arg and negative step) and
while_ factories + loop_yield complete the surface.

Seventh PR in the MIR bindings/DSL stack; completes Phase 2's control flow.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -490,6 +490,13 @@ void populate_mir(nb::module_ &m) {
           "add_phi_incoming",
           [](llvm::MachineInstr &self, llvm::Register reg,
              llvm::MachineBasicBlock *mbb) {
+            // Appending (value, block) operands only makes sense for a G_PHI;
+            // on any other instruction it silently grows it with junk operands
+            // (malformed MIR, no verifier under NDEBUG).
+            if (self.getOpcode() != llvm::TargetOpcode::G_PHI) {
+              throw nb::value_error(
+                  "add_phi_incoming requires a G_PHI instruction");
+            }
             llvm::MachineFunction &mf = *self.getMF();
             self.addOperand(mf,
                             llvm::MachineOperand::CreateReg(reg,
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -486,6 +486,18 @@ void populate_mir(nb::module_ &m) {
           },
           "block"_a,
           "Repoint a branch's target block (operand 0), e.g. a G_BR.")
+      .def(
+          "add_phi_incoming",
+          [](llvm::MachineInstr &self, llvm::Register reg,
+             llvm::MachineBasicBlock *mbb) {
+            llvm::MachineFunction &mf = *self.getMF();
+            self.addOperand(mf,
+                            llvm::MachineOperand::CreateReg(reg,
+                                                            /*isDef=*/false));
+            self.addOperand(mf, llvm::MachineOperand::CreateMBB(mbb));
+          },
+          "value"_a, "block"_a,
+          "Append a (value, predecessor-block) incoming pair to a G_PHI.")
       .def("__str__",
            [](llvm::MachineInstr &self) { return eudsl::toString(self); });
 
@@ -959,7 +971,22 @@ void populate_mir(nb::module_ &m) {
             return res;
           },
           "type"_a, "incomings"_a,
-          "Build a G_PHI from (value, predecessor-block) pairs.");
+          "Build a G_PHI from (value, predecessor-block) pairs.")
+      .def(
+          "build_empty_phi",
+          [](llvm::MachineIRBuilder &self,
+             llvm::LLT ty) -> llvm::MachineInstr * {
+            llvm::Register res =
+                self.getMRI()->createGenericVirtualRegister(ty);
+            llvm::MachineInstrBuilder phi =
+                self.buildInstr(llvm::TargetOpcode::G_PHI);
+            phi.addDef(res);
+            return phi.getInstr();
+          },
+          "type"_a, nb::rv_policy::reference_internal,
+          "Build a G_PHI with only its def (of type `type`); add incomings "
+          "later "
+          "with MachineInstr.add_phi_incoming. Its def is operand 0.");
 
   // The innermost MachineIRBuilder entered as a context manager, mirroring the
   // IR module's current_builder(). Raises when there is none.
--- a/projects/eudsl-llvmpy/src/llvm/dsl/machine_cf.py
+++ b/projects/eudsl-llvmpy/src/llvm/dsl/machine_cf.py
@@ -21,8 +21,9 @@ cond_br.set_successor. The values passed to yield_ become the merge block's
 G_PHIs, built on the else branch (when both incoming edges are known), so
 `r = yield_(...)` binds to the phi that `return r` uses.
 
-Only if/else is lowered here: a for/while in a @machine_function body is
-rejected (MIRCanonicalizer installs no loop transformers).
+Loops (for/while) lower the same way as llvm.dsl.cf._Loop: an inline
+preheader -> header(phis) -> body -> exit shape whose header G_PHIs carry the
+induction variable and loop-carried values.
 """
 
 from contextlib import contextmanager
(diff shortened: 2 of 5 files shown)


makslevental (commented):
## Comprehensive PR Review — #586 (Generic-MIR for/while loop lowering)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`Machine.cpp` + `dsl/machine_cf.py` + new `test_dsl_machine_loops.py`). This completes the control-flow stack, and the **core lowering is correct**: the general-code review independently verified the phi operand order, back-edge predecessor capture (`body_end = insert_block`, handling nested control flow), the induction compare direction (SLT ascending / SGT descending), witness/coerce inference, CFG/terminator consistency, and balanced `_loop_stack` discipline — all a faithful mirror of the already-merged IR sibling `cf.py::_Loop`. Comments are accurate and the prior forward-reference text ("until the loop runtime lands") is cleanly removed. This PR also resolves the loop-rejection concern raised on #585 by actually lowering loops.

**One important framing note up front:** the silent-failure review rated several unenforced-invariant issues as Critical/High, while the code review verified the lowering correct and classified the same issues as below-threshold *latent* notes — because the `ForToInline`/`WhileToInline` AST transformer controls the yield/arity contract, so they aren't reachable through the supported DSL surface, and most apply equally to the merged IR sibling (i.e. not regressions). Both are right about different things: the code is correct on the supported path; the concerns are about what happens under misuse or direct API use in a release (`NDEBUG`) wheel where the verifier is compiled out. Findings below are framed accordingly — treat them as **hardening**, with severity depending on how much you want the raw bindings to defend against out-of-surface use.

> Review-only — nothing was changed. `machine_cf.py`/`Machine.cpp` line numbers are approximate.

---

### 🔴 The one genuinely net-new footgun

**`add_phi_incoming` appends operands to *any* instruction with no G_PHI check.** (`Machine.cpp` ~187-197)
This is the standout finding — flagged by the code, type, and silent-failure reviews alike, and unlike the items below it is *new* in this PR (not inherited from the sibling). The binding does `self.addOperand(mf, CreateReg(reg, isDef=false))` + `CreateMBB(mbb)` with no check that `self.getOpcode() == TargetOpcode::G_PHI`. Called on a non-phi instruction (a `G_ADD`, `G_BR`, copy…) it silently grows that instruction with junk operands → malformed MIR, no error, surfacing later as an opaque crash in printing/codegen. Every sibling binding in this file that indexes operands (`operand`, `set_branch_target`) guards and throws `nb::value_error`/`index_error`; this one abandons that discipline. Recommend an opcode guard throwing `nb::value_error` — it closes the hole exactly where it's opened.

---

### 🟠 Unenforced phi-completeness (hardening; correct on the supported path)

These share a root cause: phis are built in two steps (`build_empty_phi` def-only → `add_phi_incoming` back-edge) with **no completeness check at either end**, relying on a MachineVerifier that's compiled out of the shipped wheel. The code review confirmed the *supported* DSL path always completes them; the exposure is misuse/direct-API.

**1. `build_empty_phi` can leave a def-only (incomplete) phi.** (`Machine.cpp` ~492-504) Correct only if every predecessor's incoming is later added; nothing enforces it. This is the enabling primitive for the paths below.

**2. `zip(carried_phis, next_carried)` silently drops back-edge incomings.** (`machine_cf.py` ~246) `next_carried` defaults to `[]` and is only filled by `loop_yield`. If `loop_yield` is never called (body carries values but doesn't yield) or yields a different count than `iter_args`, `zip` truncates silently — carried phis keep only their preheader incoming and never get the body→header back-edge, while line ~242 *does* add the CFG edge. Result: structurally malformed phi, no "you yielded 2 but the loop carries 3" message. An explicit `len(next_carried) == len(carried_phis)` check in `__exit__` would turn a silent miscompile into a loud error. *(Code review: not reachable via the supported DSL surface since the AST transform synthesizes `loop_yield`; silent-failure: HIGH as a latent footgun. A cheap assert satisfies both.)*

**3. Body-terminated block → possible double terminator.** (`machine_cf.py` ~240-244) `__exit__` unconditionally emits `build_br(header)` + `add_successor(header)` on the current block. If the body already left the builder in a terminated block (e.g. a nested `if` whose arms both `return`), this appends a second terminator / contradictory CFG edge. Worth a "not already terminated" check. *(Flagged by silent-failure; not contradicted but not confirmed by code review — worth verifying against a return-in-loop-body case.)*

---

### 🟡 Numeric / semantic silent-correctness (mostly shared with the IR sibling)

- **Signed comparison hardcoded** (`machine_cf.py` ~221 → `machine.py` ~78-88): the induction/`while` condition always uses `SLT`/`SGT`; an unsigned induction variable silently gets a signed compare (wrong trip count). Brushes CLAUDE.md's "bind the full API; do not hardcode assumptions." Mirrors `ArithValue`, so a consistency call — but flagged before on this stack.
- **`descending` misdetects non-int steps** (`machine_cf.py` ~220): `isinstance(self.step, int) and step < 0` — a negative step passed as a `MachineValue` or float isn't recognized, so a descending loop silently compiles with an ascending compare and never runs.
- **Float step silently truncated** (`_as_mv` ~138 / `MachineValue._coerce` ~42): `int(x)` truncates a float step/bound with no warning (compounds with the above).
- **`step == 0` → infinite loop** (`machine_cf.py` ~241): `next_iv = iv + 0` never advances; no guard, unlike Python's `range()` which raises. Shared with the sibling.
- **`while_` cond_fn result unchecked** (`machine_cf.py` ~223-224): a non-`MachineValue` return (folded bool / `None`) fails late on `.reg` with an opaque `AttributeError`; a non-i1 MachineValue builds a malformed `G_BRCOND` silently.
- **`loop_yield` outside a loop** (`machine_cf.py` ~259): bare `IndexError` on `_loop_stack[-1]` instead of a clear message (cf. `current_machine_builder`'s clean `RuntimeError`).

---

### 🟡 Type design

**`_MIRLoop` is a string-discriminated `for`/`while` union** (`machine_cf.py` ~145): all of `start/stop/step/iv_phi` (for) and `cond_fn` (while) always exist defaulting to `None`, with scattered `if self.kind == "for"` branches, so illegal combinations (`while` with `start`, `for` with `cond_fn`) are representable, and `iv_phi is None` doubles as the while-marker. `.results` is `()` before `__exit__` — a silent-wrong-answer if read early rather than an error. And the incremental-phi API makes "phi with a def but zero incomings" a first-class unmarked state (a regression in abstraction strength vs the atomic `build_phi`, though the capability is genuinely needed for back-edges). This all faithfully mirrors `_Loop`, so it's defensible as a consistency choice — the type review's suggestions (split into `for`/`while` subclasses; gate `.results` behind a `_finished` flag; assert carried arity) would apply to *both* files together. Ratings: encapsulation 4/10, invariant expression 3/10, usefulness 6/10, enforcement 3/10.

---

### 🧪 Tests — coarse; happy-path presence checks only

Four tests, all leak-checked, covering for/while, single- vs two-arg `range_`, and the ascending/descending predicate split (`intpred(slt)`/`intpred(sgt)` — the one place assertions actually discriminate a code branch). But every assertion is **substring membership in `to_mir()` + block count `== 4`**, so the SSA wiring that is the entire point goes unverified. Priority gaps:
- **PHI incoming structure never checked** (priority 10) — no test asserts each header phi has exactly two `(value, block)` incomings with preheader + body-end predecessors, or that the induction phi's seed is `start` and back-edge is `iv+step`. A dropped/swapped incoming passes.
- **Loop-carried values (`iter_args`/`.results`) never exercised** (priority 9) — the carried-phi back-edge routing (the `zip` at ~246) has no test; a regression wiring the wrong slot goes undetected.
- **Nested control flow (if-in-loop, loop-in-loop) untested** (priority 9) — the `body_end = insert_block` line exists precisely for this, yet every test has a straight-line body where `body_end == body`, so that branch never runs.
- **No test executes or even verifies** the built MIR — running the MachineVerifier on one built loop would catch exactly the malformed-phi defects above that text inspection can't.
- Untested: all-int-bounds `NotImplementedError`, `step==0`, explicit step >1, multi-carried `while` condition, direct C++ tests for `build_empty_phi`/`add_phi_incoming`.

---

### ✅ Handled well
- **The lowering is correct** (independently verified): phi operand order (`def, reg, mbb, …`), back-edge predecessor capture under nested control flow, SLT/SGT direction, witness inference, preheader→header→{body,exit}→header CFG with terminators matching successors, and balanced `_loop_stack` (popped before the exception check).
- Necessary new capability — deferred back-edge phi — is provided; the incremental API mirrors LLVM's IRBuilder `phi()`+`addIncoming()`.
- Exception path pops `_loop_stack` before returning `False`, keeping stack state consistent.
- All docstrings verified accurate; the prior forward-reference is removed; the `MIRCanonicalizer` pass-order comment usefully captures *why* loops precede if/else. (One low nit: the class docstring says "the header phis are the loop-carried live-outs" — only the carried phis, not the induction phi, become `.results`.)
- CLAUDE.md compliant: brace style, imports at top, `assert_no_leaks()` usage, no "Phase N" in code.

---

makslevental:
Addressed the automated review. Changes are in the updated `cf-loops` commit and cascaded up the rest of the stack (#587–#591).

**🔴 The net-new footgun**
- **`add_phi_incoming` now checks its instruction is a `G_PHI`.** On any other instruction it silently appended junk operands → malformed MIR; it now throws `nb::value_error`, matching the file's other guarded accessors. (Test at the C++ level.)

**🟠 Phi/loop completeness**
- **Carried-value arity:** `_MIRLoop.__exit__` now verifies the body yielded exactly as many carried values as the loop carries, instead of `zip`-truncating the back-edge incomings into a structurally-incomplete phi.
- **`build_empty_phi`:** left as the deliberate two-step primitive — completeness can't be enforced in C++ (no predecessor count at build time), and the DSL always completes it; the `add_phi_incoming` G_PHI guard + the arity check above backstop misuse.
- **Body double-terminator:** not reachable through the supported DSL (`RejectUnsupportedJumps` forbids an early `return` in a loop body, so the body can't self-terminate); guarding it cleanly would need a new "is-terminated" accessor, so I left it as a documented out-of-surface case.

**🟡 Numeric / semantic**
- **`range_` step:** rejects a non-int step (its sign picks the compare direction at trace time, so a `MachineValue`/float step can't work — and a float would silently truncate) and a zero step (infinite loop, like Python's `range()`). The `descending` detection is now simply `step < 0`, no longer misdetecting non-int steps.
- **`while_` condition:** validated to be an i1 `MachineValue` (was failing late on `.reg` or building a malformed `G_BRCOND`).
- **`loop_yield` outside a loop:** clear `RuntimeError` instead of a bare `IndexError`.
- **Signed comparison default:** kept (mirrors `ArithValue`/`cf.py`); unsigned is reachable via the `ult`/`ule`/`ugt`/`uge` methods added earlier in the stack.
- Docstring fix: only the loop-carried phis (not the induction phi) become `.results`.

**🟡 Type design (`_MIRLoop` union):** left as-is — it faithfully mirrors the merged IR sibling `cf.py::_Loop`, so splitting it into `for`/`while` subclasses is a cross-file consistency change better done to both together.

**Tests (the big gap — structural, not text):** induction/carried phi **incoming structure** (operand counts), loop-carried **`.results` routing** (register identity vs the header phi defs, so a wrong-slot back-edge fails), **if-in-loop** (exercises the moved back-edge predecessor — previously every body was straight-line), and negative paths for every new guard (`step==0`, non-int step, all-int-bounds, while-condition, `loop_yield`-outside, carried-arity mismatch), plus body-exception propagation + `_loop_stack` cleanup and C++-level `add_phi_incoming`/`build_empty_phi`. `src/MIR` stays at 100% line+function; `dsl/machine_cf.py` at 100%.

Verified: PR #586's branch builds and its 99 `tests/mir` pass; the cumulative top builds and all pass; full Python coverage clears the 99% gate.

PR #587, opened 2026-08-17
[eudsl-llvmpy] Add target-specific MIR construction (BuildMI + opcode lookup)

makslevental:
Build real target instructions (e.g. AArch64 ADDWrr) without the generated
target opcode enums, which are not installed with LLVM: MachineFunction.opcode
resolves an opcode number by mnemonic through the function's TargetInstrInfo
(and opcode_name inverts it). MachineIRBuilder.build_instr(opcode) is the
BuildMI analogue -- it inserts an empty instruction whose operands are then
appended with MachineInstr.add_def/add_use/add_imm/add_mbb.

Eighth PR in the MIR bindings/DSL stack; first half of Phase 3.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -122,6 +122,24 @@ void requireVReg(llvm::MachineIRBuilder &b, llvm::Register reg,
             .c_str());
 }
 
+// getInstrInfo() can be null for a target without one; real backends always
+// have it (these MachineFunctions come from create_machine_function/codegen
+// with a real subtarget), so this is purely defensive.
+const llvm::TargetInstrInfo &requireTII(const llvm::MachineFunction &mf) {
+  const llvm::TargetInstrInfo *tii = mf.getSubtarget().getInstrInfo();
+  if (!tii)
+    throw nb::value_error("target has no TargetInstrInfo"); // LCOV_EXCL_LINE
+  return *tii;
+}
+
+// MCInstrInfo::get/getName index an array bounded only by an assert (compiled
+// out under NDEBUG), so an out-of-range opcode would read out of bounds -- a
+// segfault or, worse, a silently-wrong name. Validate before indexing.
+void requireValidOpcode(const llvm::TargetInstrInfo &tii, unsigned opcode) {
+  if (opcode >= tii.getNumOpcodes())
+    throw nb::index_error("opcode number out of range");
+}
+
 // The "current MachineIRBuilder" is tracked on a thread-local stack, mirroring
 // the IR builder's `with builder:` / current_builder() mechanism (see the
 // thread_local ThreadContextEntry stack in IR/Builder.cpp). `with builder:`
@@ -651,10 +669,9 @@ void populate_mir(nb::module_ &m) {
       .def(
           "opcode",
           [](llvm::MachineFunction &self, const std::string &name) -> unsigned {
-            const llvm::TargetInstrInfo *tii =
-                self.getSubtarget().getInstrInfo();
-            for (unsigned i = 0, e = tii->getNumOpcodes(); i < e; ++i) {
-              if (tii->getName(i) == name)
+            const llvm::TargetInstrInfo &tii = requireTII(self);
+            for (unsigned i = 0, e = tii.getNumOpcodes(); i < e; ++i) {
+              if (tii.getName(i) == name)
                 return i;
             }
             throw nb::key_error(
@@ -665,7 +682,9 @@ void populate_mir(nb::module_ &m) {
       .def(
           "opcode_name",
           [](llvm::MachineFunction &self, unsigned opcode) {
-            return self.getSubtarget().getInstrInfo()->getName(opcode).str();
+            const llvm::TargetInstrInfo &tii = requireTII(self);
+            requireValidOpcode(tii, opcode);
+            return tii.getName(opcode).str();
           },
           "opcode"_a, "The mnemonic for a target opcode number.")
       .def("__str__",
@@ -1046,12 +1065,16 @@ void populate_mir(nb::module_ &m) {
           "build_instr",
           [](llvm::MachineIRBuilder &self,
              unsigned opcode) -> llvm::MachineInstr * {
+            requireValidOpcode(requireTII(self.getMF()), opcode);
             return self.buildInstr(opcode).getInstr();
           },
           "opcode"_a, nb::rv_policy::reference_internal,
           "Build and insert an empty instruction of the given opcode; append "
           "operands with MachineInstr.add_def/add_use/add_imm/add_mbb. The "
-          "BuildMI analogue for target-specific opcodes.");
+          "BuildMI analogue for target-specific opcodes. Like BuildMI, the "
+          "appended operands are not validated against the opcode's "
+          "MCInstrDesc (arity/kind/def-first order are the caller's "
+          "responsibility).");
 
   // The innermost MachineIRBuilder entered as a context manager, mirroring the
   // IR module's current_builder(). Raises when there is none.
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -505,6 +505,34 @@ void populate_mir(nb::module_ &m) {
           },
           "value"_a, "block"_a,
           "Append a (value, predecessor-block) incoming pair to a G_PHI.")
+      .def(
+          "add_def",
+          [](llvm::MachineInstr &self, llvm::Register reg) {
+            self.addOperand(*self.getMF(), llvm::MachineOperand::CreateReg(
+                                               reg, /*isDef=*/true));
+          },
+          "reg"_a, "Append a register def operand.")
+      .def(
+          "add_use",
+          [](llvm::MachineInstr &self, llvm::Register reg) {
+            self.addOperand(*self.getMF(), llvm::MachineOperand::CreateReg(
+                                               reg, /*isDef=*/false));
+          },
+          "reg"_a, "Append a register use operand.")
+      .def(
+          "add_imm",
+          [](llvm::MachineInstr &self, int64_t value) {
+            self.addOperand(*self.getMF(),
+                            llvm::MachineOperand::CreateImm(value));
+          },
+          "value"_a, "Append an immediate operand.")
+      .def(
+          "add_mbb",
+          [](llvm::MachineInstr &self, llvm::MachineBasicBlock *mbb) {
+            self.addOperand(*self.getMF(),
+                            llvm::MachineOperand::CreateMBB(mbb));
+          },
+          "block"_a, "Append a machine-basic-block operand.")
       .def("__str__",
            [](llvm::MachineInstr &self) { return eudsl::toString(self); });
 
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## Comprehensive PR Review — #587 (Target-specific MIR construction: BuildMI + opcode lookup)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`Machine.cpp` + new AArch64-gated `test_build_target.py`). This is a clean, well-documented mechanical mirror of LLVM's `BuildMI` + opcode-by-name lookup, and it fills a real need — target opcode enums aren't installed with LLVM, so resolving them by name at runtime through `TargetInstrInfo` is the right approach (the test docstring explains this well). Docstrings are all accurate and there are no forward-reference/"Phase N" leaks into the code. The findings converge on one theme: **two of the new entry points take a raw integer straight to an assert-only bounds check that's compiled out of the shipped `NDEBUG` wheel** — and notably, the name→number path already demonstrates the correct guarded pattern that the number→name path omits.

> Review-only — nothing was changed. Line numbers are approximate to the PR-head `Machine.cpp`.

---

### 🔴 High — release-build OOB (raw Python int → compiled-out assert)

Both were flagged independently by the code review and silent-failure review.

**1. `MachineFunction.opcode_name(number)` has no bounds check.** (`Machine.cpp` ~311-315)
`getInstrInfo()->getName(opcode).str()` — `MCInstrInfo::getName` guards `Opcode < NumOpcodes` only with an `assert`. In the release wheel, a user-supplied out-of-range `opcode` (any Python int) indexes `InstrNameIndices[]` out of bounds → a garbage offset into `InstrNameData`, then `StringRef`'s `strlen` runs from a wild pointer. Outcome is either a segfault **or — worse and quieter — a heap over-read copied verbatim into a Python `str` and returned as a normal value**, indistinguishable from success. `mf.opcode_name(999999)` is the repro.

**2. `MachineIRBuilder.build_instr(opcode)` passes an arbitrary opcode through unchecked.** (`Machine.cpp` ~555-559)
`buildInstr(opcode)` → `TII.get(opcode)`, same `assert(Opcode < NumOpcodes)`-then-index pattern. An out-of-range opcode does an OOB read of `Desc[]` and the resulting `MachineInstr` permanently stores a pointer into garbage — latent, detonating later in `num_operands`/printing/passes. This chains with the existing `MachineInstr.opcode_name` property (which the tests read right after `build_instr`), so a bad opcode corrupts both the build and the readback path. In the happy path callers pass `mf.opcode(...)` (validated), but the raw `unsigned` parameter is directly reachable from Python and untested.

**The fix pattern already exists in this same file:** `opcode(name)` correctly bounds its loop by `getNumOpcodes()` and throws `nb::key_error` on a miss (and `MachineInstr.operand`/`MachineOperand.reg`/`imm` all guard-and-throw). #1 and #2 just need the same `< getNumOpcodes()` check before calling `getName`/`buildInstr`. The asymmetry is the sharp part: the name→number direction validates and raises a clean `KeyError`; the inverse number→name direction is undefined behavior — a user reasonably assuming both directions validate gets very different treatment.

---

### 🟠 Medium

**3. `getInstrInfo()` dereferenced without a null check.** (`Machine.cpp` ~297-315, both `opcode` and `opcode_name`)
`getInstrInfo()` has a base default returning `nullptr`; `tii->getNumOpcodes()`/`getName()` would null-deref. Not reachable today (real backends always have TII, and these `MachineFunction`s come from `create_machine_function`/codegen with a real subtarget), so defensive — but cheap to harden.

**4. `add_def`/`add_use`/`add_imm`/`add_mbb` append operands with no arity/kind/order check.** (`Machine.cpp` ~200-227)
A user can `add_def` to an instruction with no def slot, append a use where an immediate is required, or append the wrong operand count for a fixed-arity target opcode → silently malformed MIR (verifier not run, asserts compiled out). This *deliberately mirrors* LLVM's `BuildMI`/`addOperand`, which is likewise caller-trusted — so it's consistent with the model rather than a novel defect, and consistent with the prior PR's `add_phi_incoming`. The severity note is that it hands an unchecked primitive to a Python audience with no compiler guardrails; worth documenting the silent-corruption exposure. Confirmed safe: `add_mbb(None)` fails loudly (nanobind rejects `None` for a non-`.none()` pointer arg with `TypeError`); LLVM's operand kinds themselves are correctly constructed (`CreateReg` def/use bit, `CreateImm`, `CreateMBB`).

---

### 🟡 Type/API design

**5. `build_instr` makes a half-built instruction a first-class value.** (`Machine.cpp` ~555)
C++'s `BuildMI` keeps the incomplete instruction bottled in a `MachineInstrBuilder` temporary consumed in one expression; this binding hands back a fully-materialized, already-inserted `MachineInstr*` with **zero operands**, relying on later `add_*` calls to complete it — and nothing ties the appended operands to the opcode's `MCInstrDesc` (arity/kind/def-first ordering all unconsulted, though reachable via the same TII path already used for `opcode`). This is the `build_empty_phi` concern from the prior PR, slightly worse (that one at least pre-installs the def). Contrast the high-tier typed builders (`build_add`/`build_phi`) which return a def `Register` and can't be left half-built — the two tiers are individually coherent but co-located as peers with no signposting that one returns a finished value and the other a shell you must finish.

**6. `opcode_name` naming collision and home.** `MachineInstr.opcode_name` (existing property, no args, "my own mnemonic") vs. the new `MachineFunction.opcode_name(number)` (method, "mnemonic of an arbitrary opcode") — same identifier, disambiguated only by receiver. The `MachineFunction` pair is really a bidirectional table; naming the reverse `opcode_number`/`opcode_name`, or a matched pair, would read more coherently. Also, `opcode`/`opcode_name` both immediately delegate to `getSubtarget().getInstrInfo()` — the opcode namespace belongs to the subtarget/TII, not the `MachineFunction` (which is just a convenient handle); if a `Subtarget`/`TII` type is ever bound, these would want to move (a breaking change). Ratings: encapsulation 6/10, invariant expression 3/10, usefulness 6/10, enforcement 3/10.

**7. Cross-function opcode provenance (known sharp edge).** An `unsigned` opcode and a `Register` cross the boundary as bare untyped values. An opcode from function A is meaningful in function B's `build_instr` **iff both share a target** (opcode numbering is per-target); different triples silently disagree with no detection. Inherent to mirroring LLVM's untyped representation — noted, not a defect introduced here.

---

### 🧪 Tests — happy-path construction only; the OOB paths are untested

Five tests, all leak-checked: opcode round-trip, unknown-name `KeyError` (good positive+negative on the *name* path), and building ADDWrr / MOVi32imm / B. Gaps:
- **Out-of-range number to `opcode_name(number)` / `build_instr(opcode)` is untested** (priority 7) — exactly the two HIGH findings above; the number path is never exercised with an independent or out-of-range value, only with one derived from `opcode()`.
- **No round-trip / verifier / serialization** of the built instruction (priority 6) — tests assert only operand count + `is_def`/`is_use`/`imm` flags; nothing runs `to_mir()`/`parse_mir`, the verifier, or even `str(mi)`. Notably `test_build_branch_with_mbb_operand` adds the MBB operand but never registers `target` as a CFG successor, leaving a verifier-inconsistent function that a verifier assertion would catch.
- **Operand identity/order not asserted beyond flags** (priority 5) — the two ADDWrr uses (`x` before `y`) — the whole point of an operand appender — aren't checked to carry the right registers in the right order; `operand(2)` and the MOVi32imm def are unverified.
- **AArch64 gating** collapses all five tests to the arm64 CI leg (skipped on x86 ubuntu/windows) — same known trade-off as the earlier inspect/roundtrip tests; the mechanisms under test are target-agnostic, so a break in the appenders would only surface on arm64.

---

### ✅ Handled well
- `opcode(name)` is the model to copy: bounded loop, throws a specific `nb::key_error("no target opcode named '<name>'")`, positive+negative tested.
- Runtime name-resolution design is correct and well-justified (target opcode enums aren't shipped in installed LLVM headers) — a genuine need, cleanly solved.
- `add_mbb(None)` fails loudly; operand-kind construction (`CreateReg` def/use, `CreateImm`, `CreateMBB`) is correct.
- The `add_def`/`add_use` split (vs a raw `add_reg(is_def=...)`) reads well.
- CLAUDE.md compliant: LLVM API naming (`addOperand`/`Create*`/`buildInstr`/`getName`), brace style correct (the multi-line lambda bodies are single statements; the `opcode` `for` is correctly braced), imports at top, `assert_no_leaks()` throughout, no forward-reference/"Phase N" comments in the source.
- All docstrings verified accurate; `build_instr`'s and the test module's docstrings are exemplary "why" documentation.

---

makslevental:
Addressed the automated review. Changes are in the updated `target-buildmi` commit and cascaded up the rest of the stack (#588–#591).

**🔴 Release-build OOB (the two HIGH findings)** — the number→name direction took a raw Python int straight to an assert-only bounds check that's compiled out of the `NDEBUG` wheel, unlike the name→number direction which already validated:
- **`opcode_name(number)`** and **`build_instr(opcode)`** now validate the opcode is in range (`requireValidOpcode`) before it reaches `MCInstrInfo::get`/`getName`. Out-of-range now raises `IndexError` (matching the guarded operand/reg accessors) instead of reading out of bounds → a segfault or a silently-wrong name / garbage `MachineInstr`.

**🟠 Medium**
- **#3** factored a `requireTII()` helper that also null-checks `getInstrInfo()` (defensive — real backends always have it), shared across `opcode`/`opcode_name`/`build_instr`.
- **#4** `add_def`/`add_use`/`add_imm`/`add_mbb` stay unchecked by design — they deliberately mirror LLVM's caller-trusted `BuildMI`/`addOperand` (consistent with `add_phi_incoming`) — but `build_instr`'s docstring now states that appended operands aren't validated against the opcode's `MCInstrDesc` (arity/kind/def-first order are the caller's responsibility).

**🟡 Type/API design**
- **#5** the half-built-instruction nature of `build_instr` is now signposted in its docstring (the `add_*` completion contract). Kept the two builder tiers (typed `build_add`/`build_phi` vs raw `build_instr`) as-is.
- **#6/#7** kept `MachineFunction.opcode`/`opcode_name` names — the collision with `MachineInstr.opcode_name` is receiver-disambiguated and both mirror LLVM's `getName`; renaming (or moving to an unbound `Subtarget`/`TII` type) is churn for a naming nit. Cross-function opcode provenance is inherent to LLVM's untyped model; `build_instr`'s new in-range check now rejects an opcode that isn't valid for this function's target.

**Tests:** out-of-range `opcode_name(999999)` and `build_instr(999999)` raise `IndexError` (the two HIGH findings); operand **identity and order** asserted for `ADDWrr` (def `d`, uses `x` then `y`) and `MOVi32imm` (def `d`); `str(mi)` checked; and the branch test now registers its target as a CFG successor so the built function stays verifier-consistent. `src/MIR` stays at 100% line + function coverage.

Verified: PR #587's branch builds and its 106 `tests/mir` pass; the cumulative top builds and all pass; full Python coverage clears the 99% gate.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -505,6 +523,34 @@ void populate_mir(nb::module_ &m) {
...
+          [](llvm::MachineInstr &self, llvm::Register reg) {
+            self.addOperand(*self.getMF(), llvm::MachineOperand::CreateReg(
+                                               reg, /*isDef=*/true));
+          },
+          "reg"_a, "Append a register def operand.")
+      .def(
+          "add_use",
+          [](llvm::MachineInstr &self, llvm::Register reg) {
+            self.addOperand(*self.getMF(), llvm::MachineOperand::CreateReg(
+                                               reg, /*isDef=*/false));
```
makslevental, line 537: why did you hardcode `isDef` here?

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -505,6 +523,34 @@ void populate_mir(nb::module_ &m) {
           },
           "value"_a, "block"_a,
           "Append a (value, predecessor-block) incoming pair to a G_PHI.")
+      .def(
+          "add_def",
+          [](llvm::MachineInstr &self, llvm::Register reg) {
+            self.addOperand(*self.getMF(), llvm::MachineOperand::CreateReg(
+                                               reg, /*isDef=*/true));
```
makslevental, line 530: why did you hardcode `isDef` here?

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: It's not a hidden default — this is the `add_def` method, so appending a *def* operand is the whole point of it. It mirrors LLVM's `MachineInstrBuilder::addDef` (MachineInstrBuilder.h:218), the def-specific spelling that sits alongside `addUse`; the def-ness is encoded in the method name rather than passed as a flag. The fully-parameterized binding that *does* surface `is_def` (plus is_kill/is_dead/is_undef/sub_reg/…) is `add_reg`, added in the build-selected PR (#588) up the stack — `add_def`/`add_use` are the two common-case shortcuts.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Same as `add_def`: this is `add_use`, so a *use* operand is its purpose and `isDef=false` is the method's identity, not a hidden assumption. It mirrors LLVM's `MachineInstrBuilder::addUse` (MachineInstrBuilder.h:225 — which even asserts against accidentally *defining* a register through `addUse`). The flag-exposing `add_reg` (with `is_def` and the rest) is added in #588.

PR #588, opened 2026-08-17
[eudsl-llvmpy] Run the GlobalISel pipeline + expose the machine verifier

makslevental:
run_codegen_to_mir gains global_isel=: when set it runs the GlobalISel lowering
pipeline (IRTranslator -> Legalizer -> RegBankSelect -> InstructionSelect) rather
than SelectionDAG ISel, so the retained MIR is fully selected target MIR
(DisableWithDiag keeps a legalization fallback from aborting). MachineFunction.
verify runs the machine verifier (non-aborting), returning whether the MIR is
well-formed.

Running individual GlobalISel passes on DSL-built generic MIR is intentionally
not exposed: the pass factories and an MMI-injecting pass ctor are absent from
the installed LLVM headers, so lowering from IR via the pipeline is the route.

Ninth PR in the MIR bindings/DSL stack; completes Phase 3.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Target.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Target.cpp
@@ -76,6 +76,12 @@ void populate_target(nb::module_ &m) {
                    [](llvm::TargetMachine &self) {
                      return self.createDataLayout().getStringRepresentation();
                    })
+      .def_prop_ro(
+          "global_isel",
+          [](llvm::TargetMachine &self) -> bool {
+            return self.Options.EnableGlobalISel;
+          },
+          "Whether instruction selection uses GlobalISel (vs SelectionDAG).")
       .def(
           "emit_assembly",
           [](llvm::TargetMachine &self, eudsl::Module &mod) {
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -32,6 +32,7 @@
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/Support/MemoryBuffer.h>
 #include <llvm/Target/TargetMachine.h>
+#include <llvm/Target/TargetOptions.h>
 
 #include <nanobind/stl/pair.h>
 
@@ -687,6 +688,15 @@ void populate_mir(nb::module_ &m) {
             return tii.getName(opcode).str();
           },
           "opcode"_a, "The mnemonic for a target opcode number.")
+      .def(
+          "verify",
+          [](llvm::MachineFunction &self) {
+            std::string buf;
+            llvm::raw_string_ostream os(buf);
+            return self.verify(/*p=*/nullptr, /*Banner=*/nullptr, &os,
+                               /*AbortOnError=*/false);
+          },
+          "Run the machine verifier; returns True if no problems were found.")
       .def("__str__",
            [](llvm::MachineFunction &self) { return eudsl::toString(self); });
 
@@ -736,14 +746,22 @@ void populate_mir(nb::module_ &m) {
   // passes are added -- not the object-emission pipeline (addPassesToEmitFile),
   // which would append FreeMachineFunctionPass -- so the MachineFunctions are
   // retained for inspection, the state `llc -stop-after=finalize-isel`
-  // produces.
+  // produces. With global_isel=True the ISel passes are the GlobalISel pipeline
+  // (IRTranslator -> Legalizer -> RegBankSelect -> InstructionSelect), so the
+  // retained MIR is fully selected target MIR rather than SelectionDAG output;
+  // DisableWithDiag keeps a legalization failure from aborting the process.
   m.def(
       "run_codegen_to_mir",
-      [](eudsl::Module &mod, llvm::TargetMachine &tm) {
+      [](eudsl::Module &mod, llvm::TargetMachine &tm, bool globalISel) {
         std::shared_ptr<llvm::LLVMContext> ctxKeepAlive =
             mod.context().shared();
         std::unique_ptr<llvm::Module> module = mod.take();
         module->setDataLayout(tm.createDataLayout());
+
+        if (globalISel) {
+          tm.setGlobalISel(true);
+          tm.setGlobalISelAbort(llvm::GlobalISelAbortMode::DisableWithDiag);
+        }
 
         auto pm = std::make_unique<llvm::legacy::PassManager>();
         llvm::TargetPassConfig *tpc = tm.createPassConfig(*pm);
@@ -778,7 +796,8 @@ void populate_mir(nb::module_ &m) {
                                      std::move(pm), /*ownedMmi=*/nullptr,
                                      &mmiwp->getMMI()};
       },
-      "module"_a, "target_machine"_a, nb::keep_alive<0, 2>());
+      "module"_a, "target_machine"_a, "global_isel"_a = false,
+      nb::keep_alive<0, 2>());
 
   // Parse .mir text into a MachineModuleInfo. Mirrors run_codegen_to_mir's
   // result but builds the MachineFunctions by deserialization rather than
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -691,12 +691,21 @@ void populate_mir(nb::module_ &m) {
       .def(
           "verify",
           [](llvm::MachineFunction &self) {
-            std::string buf;
-            llvm::raw_string_ostream os(buf);
-            return self.verify(/*p=*/nullptr, /*Banner=*/nullptr, &os,
-                               /*AbortOnError=*/false);
+            return self.verify(/*p=*/nullptr, /*Banner=*/nullptr,
+                               /*OS=*/nullptr, /*AbortOnError=*/false);
           },
           "Run the machine verifier; returns True if no problems were found.")
+      .def(
+          "verify_diagnostic",
+          [](llvm::MachineFunction &self) {
+            std::string buf;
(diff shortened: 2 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #588 (Run the GlobalISel pipeline + expose the machine verifier)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`Machine.cpp` + new AArch64-gated `test_mir_passes.py`). Both capabilities are genuinely valuable and the right primitives for this inspection module — a GlobalISel-selected MIR path and a non-aborting verifier — with good LLVM-API naming fidelity (`verify` mirrors `MachineFunction::verify`; `global_isel` mirrors `setGlobalISel`), a backward-compatible default, and both verify branches (True/False) tested. But the review converged strongly on **two design issues that reinforce each other**: the `global_isel` flag permanently mutates the shared `TargetMachine`, and a GlobalISel failure is invisible to the caller — while `verify()`, the only tool that could detect it, throws away the diagnostic explaining what's wrong.

> Review-only — nothing was changed. Line numbers are approximate to the PR-head `Machine.cpp`.

---

### 🔴 High — `global_isel` permanently mutates the shared TargetMachine (flagged by 4 of 5 agents)

**`run_codegen_to_mir` sets `tm.setGlobalISel(true)` / `setGlobalISelAbort(DisableWithDiag)` and never restores them.** (`Machine.cpp` ~371-374)
The `tm` is caller-owned, pinned via `keep_alive<0,2>`, and explicitly designed for reuse across calls. These setters mutate that shared object in place, and the `global_isel=False` path does *nothing* to reset them — so the mode is sticky/monotonic:
```python
run_codegen_to_mir(mod1, tm, global_isel=True)   # flips tm into GISel + DisableWithDiag
run_codegen_to_mir(mod2, tm)                       # default False — but tm is STILL GISel
```
The second call silently runs GlobalISel (and with abort disabled) despite `global_isel=False`. This is a parameter whose *apparent* scope (one call) doesn't match its *actual* scope (the `tm` forever) — and two distinct pieces of persistent state (selection mode + abort mode) ride on one transient-looking bool. Every test constructs a fresh `TargetMachine`, so nothing exercises reuse and the leak is uncaught. Suggested directions: expose GlobalISel as a property on the `TargetMachine` binding (matches LLVM's model, makes `run_codegen_to_mir` a pure consumer), or RAII save/restore around the run, or set both flags explicitly on *every* call (including `False`) so behavior is a pure function of the argument.

---

### 🔴 High — GlobalISel failure is invisible; MIR may not actually be "fully selected"

**`DisableWithDiag` + no diagnostic handler + ignored `pm->run` = silent incomplete selection.** (`Machine.cpp` ~374, ~258)
`DisableWithDiag` is by definition the "don't abort" mode: when the GlobalISel pipeline (IRTranslator→Legalizer→RegBankSelect→InstructionSelect) can't legalize/select an operation, it emits a diagnostic and continues — leaving residual generic `G_*` ops or falling back. That diagnostic goes to the `LLVMContext` handler, and eudsl installs none, so it lands on stderr (often invisible under a Python/notebook host — the same swallowed-diagnostic pattern flagged throughout this stack). `run_codegen_to_mir` inspects no failure state and returns the MMI unconditionally, so **an incompletely-selected function is handed back as success** with no programmatic signal. The only way to detect it is a separate `verify()` call — which (next finding) won't tell you *why*.

This is corroborated by the comment review: the updated block comment claims "**the retained MIR is fully selected target MIR** … DisableWithDiag keeps a legalization failure from aborting" — the two clauses are in tension. "Fully selected" is only true on success; the very mode cited is what makes it conditional. (The "keeps a failure from aborting" clause itself is accurate — though it's equally true of plain `Disable`; noting the *diagnostic* is what distinguishes `DisableWithDiag`.) Recommend softening the comment to "fully target-selected *when GlobalISel succeeds*," and, more substantively, giving the caller a real failure channel (a capturing diagnostic handler that raises/attaches text, `Enable` mode, or a post-run scan for residual `isPreISelGenericOpcode` instructions when `global_isel=True`).

---

### 🟠 Important — `verify()` captures the diagnostic, then discards it

**`Machine.cpp` ~319-325:**
```cpp
std::string buf;
llvm::raw_string_ostream os(buf);
return self.verify(nullptr, nullptr, &os, /*AbortOnError=*/false);
```
The lambda wires up `os` over `buf` *specifically to capture the verifier's explanation* of why the function is ill-formed — then drops `buf` and returns only a bare `bool`. A caller who gets `False` (as in `test_incomplete_generic_function_does_not_verify`) learns *that* the MIR is malformed but nothing about *why*, even though the "why" was already collected and thrown away. This is inconsistent with how the rest of the file surfaces failure (parse/codegen throw with messages; bad operands throw `nb::value_error`). Returning a bare bool for a *query* ("is this well-formed?") is defensible — I would **not** make it throw — but the captured detail should be surfaced: return the diagnostic string (empty ⇒ well-formed), or a `(bool, str)` pair (the file already includes `nanobind/stl/pair.h`), or add a sibling that returns the text. As written the buffer plumbing is dead weight — either use it or pass `nullptr`. This also compounds the finding above: `verify()` is the only way to detect the silently-incomplete GISel MIR, and even it only says "something's wrong," never what.

---

### 🧪 Tests — happy path solid; the new interaction/failure surface is thin

Four tests, all leak-checked: GISel selects real target opcodes (`ADDWrr`/`RET_ReallyLR` via `opcode_name` — good contract-level assertions), GISel- and SelectionDAG-selected MIR both verify True, and an incomplete DSL-built generic function verifies False (both verify branches covered, default SelectionDAG path pinned). Gaps:
- **TargetMachine mode-leak untested** (priority 9) — every test builds a fresh `tm`, so the sticky-state bug above can't be caught. A test reusing one `tm` for `global_isel=True` then `False` and asserting the second call is SelectionDAG would catch it.
- **GISel legalization-failure / `DisableWithDiag` fallback untested** (priority 7) — the entire reason that mode exists is unexercised; a regression flipping it to `Enable` (aborting) would pass all current tests. Feed GISel an un-legalizable construct and assert it returns without aborting (and `verify()` is False).
- **`verify()` not tested on `parse_mir`-produced MIR** (priority 6) — a distinct construction path; a `to_mir → parse_mir → verify` round-trip of GISel-selected MIR would also confirm the headline output survives serialization.
- **AArch64-gated** — all four skip on x86 legs. Reasonable for the opcode assertions, but `verify()` is *target-agnostic* and gets zero coverage on non-AArch64 runners; one non-gated `verify()` test on trivially-built/parsed MIR would give the binding baseline coverage everywhere.
- The False-verifying test can't distinguish "caught the intended undefined-vreg problem" from "failed for another reason" — inherent to the bare-bool API (ties to the finding above).

---

### ✅ Handled well
- `verify()` correctly forces `AbortOnError=false` so it can't crash the process; naming mirrors `MachineFunction::verify`; docstring ("returns True if no problems were found") is accurate and doesn't overclaim about diagnostics.
- `global_isel=False` default is backward compatible — existing 2-arg call sites and the SelectionDAG path keep prior behavior.
- Both verify outcomes are tested; opcode assertions are behavioral (mnemonics, not enum values).
- `<llvm/Target/TargetOptions.h>` correctly added at the top include block; brace style and imports compliant.
- **No forward-reference/"Phase N" comments** in the code or test docstrings (the "Ninth PR / completes Phase 3" framing stayed in the PR body); the test docstring's rationale for not exposing individual GlobalISel passes accurately describes the current state.

---

makslevental:
Addressed the automated review. Changes are in the updated `mir-passes` commit and cascaded up the rest of the stack (#589–#591).

**🔴 `global_isel` permanently mutated the shared TargetMachine**
- `run_codegen_to_mir` now sets both flags (`setGlobalISel` + the matching abort mode) on **every** call, so selection is a pure function of the argument and a `global_isel=True` call no longer leaves the reused `tm` in GlobalISel mode for a later default call. This also fixes a latent bug: AArch64's `TargetMachine` defaults `EnableGlobalISel` to **true**, so the old `global_isel=False` path (which set nothing) silently ran GlobalISel — it now genuinely runs SelectionDAG.
- Exposed `TargetMachine.global_isel` (read-only, mirrors the option) so the mode is observable and the non-sticky behavior is testable — this is the "expose GlobalISel on the TargetMachine binding" direction the review suggested, in read-only form (I kept the `global_isel=` parameter + per-call set rather than the full pure-consumer refactor, which the review also endorsed).

**🔴 GlobalISel failure was invisible**
- `DisableWithDiag` emits its fallback as a *warning* (not the `DS_Error` the diagnostic capture catches) and leaves residual generic ops instead of aborting. `run_codegen_to_mir` now scans the selected MIR and raises if any `G_*` instruction remains, so an incompletely-selected function is reported rather than returned as "fully selected." Softened the comment to "fully target-selected *when selection succeeds*." (Didn't use `Enable`/aborting mode — it would crash the process; the scan is the non-aborting programmatic signal.)

**🟠 `verify()` captured then discarded the diagnostic**
- `verify()` no longer allocates a report buffer it throws away (passes `nullptr`), and a new **`verify_diagnostic()`** returns the verifier's explanation (`""` when well-formed) — surfacing the "why" that the bare bool dropped. I kept `verify()` → `bool` rather than changing its return type, because `test_build_selected.py` (#589) depends on the bool.

**Tests:** the `tm` mode is not sticky across reuse (the priority-9 gap), `verify_diagnostic` empty vs. non-empty (asserting the actual reason text), GISel-selected MIR survives a `to_mir → parse_mir → verify` round-trip, and GISel skips a declared (body-less) function in the residual scan. `src/IR` + `src/MIR` stay at 100% line + function coverage.

Verified: PR #588's branch builds and its 114 `tests/mir` pass; `emit-jit` (where `run_codegen_to_mir` merges with the object-emission path) builds and its 121 pass; the cumulative top builds and all pass; full Python coverage clears the 99% gate.

PR #589, opened 2026-08-17
[eudsl-llvmpy] Build already-selected target MIR by hand (Route B foundation)

makslevental:
Add the primitives to construct fully-selected target MIR directly, sidestepping
GlobalISel (whose selection failure falls back to SelectionDAG, i.e. to the IR,
which hand-built MIR lacks): MachineFunction.reg_class / physreg (by-name lookup
via TargetRegisterInfo, like the opcode lookup), create_vreg(reg_class),
MachineBasicBlock.add_livein, MachineInstr.add_reg (exposing the full
MachineOperand register flag set -- def/use, implicit, kill, dead, undef,
early-clobber, sub-register, renamable), and MachineFunction.set_property over a
MachineFunctionProperty enum. A hand-built, fully-selected AArch64 add verifies()
True -- unlike incomplete generic MIR.

Tenth PR in the MIR stack; foundation for lowering hand-built MIR to machine code
without instruction selection (emit + JIT to follow).

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -142,6 +142,46 @@ void requireValidOpcode(const llvm::TargetInstrInfo &tii, unsigned opcode) {
     throw nb::index_error("opcode number out of range");
 }
 
+// getRegisterInfo() can be null for a target without one; real backends always
+// have it (these MachineFunctions come from create_machine_function/codegen
+// with a real subtarget), so this is purely defensive (mirrors requireTII).
+const llvm::TargetRegisterInfo &requireTRI(const llvm::MachineFunction &mf) {
+  const llvm::TargetRegisterInfo *tri = mf.getSubtarget().getRegisterInfo();
+  if (!tri)
+    throw nb::value_error("target has no TargetRegisterInfo"); // LCOV_EXCL_LINE
+  return *tri;
+}
+
+// Shared register-operand construction for add_def/add_use/add_reg, so the
+// "same operation with defaults" invariant lives in one place rather than in
+// three parallel CreateReg calls. Rejects flags that LLVM only permits on the
+// opposite operand kind (kill on a use, dead on a def) and out-of-range
+// subregister indices -- all guarded by asserts in LLVM that vanish under
+// NDEBUG, so an otherwise-silent corrupt operand would result.
+void addRegOperand(llvm::MachineInstr &mi, llvm::Register reg, bool isDef,
+                   bool isImp, bool isKill, bool isDead, bool isUndef,
+                   bool isEarlyClobber, unsigned subReg, bool isDebug,
+                   bool isInternalRead, bool isRenamable) {
+  if (isKill && isDef)
+    throw nb::value_error(
+        "is_kill is only valid on a use operand (is_def=False)");
+  if (isDead && !isDef)
+    throw nb::value_error(
+        "is_dead is only valid on a def operand (is_def=True)");
+  if (isRenamable && !reg.isPhysical())
+    throw nb::value_error("is_renamable is only valid on a physical register");
+  llvm::MachineFunction &mf = *mi.getMF();
+  if (subReg) {
+    const llvm::TargetRegisterInfo &tri = requireTRI(mf);
+    if (subReg >= tri.getNumSubRegIndices())
+      throw nb::index_error("sub_reg index out of range");
+  }
+  mi.addOperand(mf,
+                llvm::MachineOperand::CreateReg(
+                    reg, isDef, isImp, isKill, isDead, isUndef, isEarlyClobber,
+                    subReg, isDebug, isInternalRead, isRenamable));
+}
+
 // The "current MachineIRBuilder" is tracked on a thread-local stack, mirroring
 // the IR builder's `with builder:` / current_builder() mechanism (see the
 // thread_local ThreadContextEntry stack in IR/Builder.cpp). `with builder:`
@@ -287,10 +327,12 @@ void populate_mir(nb::module_ &m) {
   // already-selected MIR.
   nb::class_<llvm::TargetRegisterClass>(m, "TargetRegisterClass");
 
-  // MachineFunctionProperties::Property -- the pipeline-stage flags a
-  // MachineFunction carries (set with MachineFunction.set_property). Building
-  // already-selected MIR by hand means setting these to match what the
-  // corresponding codegen stage would have produced (e.g. IsSSA).
+  // MachineFunctionProperties::Property -- the flags a MachineFunction carries
+  // (set with MachineFunction.set_property). Some mark progress through the
+  // codegen pipeline (Legalized, RegBankSelected, Selected) and some are
+  // descriptive invariants of the current body (IsSSA, NoPHIs, TracksLiveness,
+  // NoVRegs). Building already-selected MIR by hand means setting these to
+  // match what the corresponding codegen stage would have produced.
   nb::enum_<llvm::MachineFunctionProperties::Property>(
       m, "MachineFunctionProperty")
       .value("IsSSA", llvm::MachineFunctionProperties::Property::IsSSA)
@@ -318,6 +360,37 @@ void populate_mir(nb::module_ &m) {
       .def_prop_ro("is_use",
                    [](llvm::MachineOperand &self) {
                      return self.isReg() && self.isUse();
+                   })
+      .def_prop_ro("is_implicit",
+                   [](llvm::MachineOperand &self) {
+                     return self.isReg() && self.isImplicit();
+                   })
+      .def_prop_ro("is_kill",
+                   [](llvm::MachineOperand &self) {
+                     return self.isReg() && self.isKill();
+                   })
+      .def_prop_ro("is_dead",
+                   [](llvm::MachineOperand &self) {
+                     return self.isReg() && self.isDead();
+                   })
+      .def_prop_ro("is_undef",
+                   [](llvm::MachineOperand &self) {
+                     return self.isReg() && self.isUndef();
+                   })
+      .def_prop_ro("is_early_clobber",
+                   [](llvm::MachineOperand &self) {
+                     return self.isReg() && self.isEarlyClobber();
+                   })
+      .def_prop_ro("is_renamable",
+                   [](llvm::MachineOperand &self) {
+                     // isRenamable() asserts a physical register (renamable is
+                     // a post-RA physreg concept), so guard on that too.
+                     return self.isReg() && self.getReg().isPhysical() &&
+                            self.isRenamable();
+                   })
+      .def_prop_ro("sub_reg",
+                   [](llvm::MachineOperand &self) -> unsigned {
+                     return self.isReg() ? self.getSubReg() : 0;
                    })
       .def_prop_ro("reg",
                    [](llvm::MachineOperand &self) -> llvm::Register {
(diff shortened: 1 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #589 (Build already-selected target MIR by hand — Route B foundation)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`Machine.cpp` + new AArch64-gated `test_build_selected.py`). The primitives are genuinely useful and well-motivated — typed vregs, physregs, the full register-operand flag set, live-ins, and function properties are exactly what "build selected MIR by hand" needs, and the payoff (a hand-built AArch64 add that `verify()`s True) is real. The keyword-arg `add_reg` is also a genuine ergonomic win over LLVM's positional-bool `CreateReg`. But the review converged on a set of **silent-corruption paths that are invisible specifically in the release (`NDEBUG`) wheel** because their only guard is a compiled-out `assert`, plus a cluster of concrete CLAUDE.md violations that are easy fixes.

> Review-only — nothing was changed. Line numbers are approximate to the PR-head `Machine.cpp` / the new test file.

---

### 🔴 High — silent corruption (release-build `assert`s compiled out)

**1. `add_livein(reg)` silently corrupts the livein list when given a virtual register.** (`Machine.cpp` ~304-307)
`self.addLiveIn(reg.asMCReg())` — `Register::asMCReg()` only `assert`s the register is physical. In the release wheel that assert is gone, so a virtual register (very plausibly passed — the arg type is `llvm::Register`, which accepts any register, and this PR also adds `create_vreg`) is truncated into a garbage `MCRegister` and recorded as a bogus physical livein. No exception, no signal at the call site; it surfaces (if at all) far downstream. The docstring already promises "physical register" — recommend a `if (!reg.isPhysical()) throw nb::value_error(...)` guard. Flagged by code + silent-failure.

**2. `set_property` lets the user assert unchecked invariants that the verifier then trusts.** (`Machine.cpp` ~375-379) — the central design concern, flagged by all of code/tests/types/silent-failure.
`getProperties().set(prop)` sets a `MachineFunctionProperties` flag with no check that the MIR actually satisfies it. This is load-bearing for the PR (the test hand-sets `IsSSA`/`TracksLiveness`/`NoPHIs` so `verify()` passes) — but the `MachineVerifier` does not re-derive these; it *trusts the flags* and chooses which invariant classes to enforce based on them. Concretely, liveness checks (which would catch #1's bogus livein) run **only when `TracksLiveness` is set**. So a user asserting the wrong property set can make `verify()` return True on malformed MIR — an invariant inversion where the property is meant to *reflect* the MIR's state but this API lets it *declare* it. It's arguably inherent to building selected MIR by hand (there's no codegen stage to earn the flags), so the *capability* is justified — but the *shape* amplifies risk: write-only (no `reset`, no getter), and nothing warns the caller these are unchecked promises. Recommend at minimum a docstring warning ("you are asserting, not requesting — the verifier trusts you") and a `properties` getter for symmetry.

**3. `add_reg` builds contradictory/out-of-range operands silently.** (`Machine.cpp` ~237-256)
Raw booleans forwarded with zero validation: `is_kill` is only valid on a use and `is_dead` only on a def, so `add_reg(v, is_def=True, is_kill=True)` builds a nonsensical operand; `sub_reg` is an arbitrary unsigned with no range check against the reg class's subregister space. LLVM asserts all of this — compiled out in release — so the bad operand is constructed silently and detonates downstream. Since the whole feature is hand-building MIR, these are exactly the mistakes an inexperienced user will make.

**4. `reg_class` / `physreg` deref `getRegisterInfo()` with no null check.** (`Machine.cpp` ~340, ~357)
Same unguarded-deref pattern as the prior PR's `opcode()`. Null TRI → interpreter segfault rather than a Python exception. Not reachable on real targets (defensive), but cheap to harden and inconsistent with the `key_error`/bounds-check discipline elsewhere in the file. *(The not-found paths themselves are done right — specific `key_error`, and `physreg` correctly scans from `i=1` to skip `NoRegister`.)*

---

### 🟠 Important — CLAUDE.md compliance (concrete, easy)

**5. Function-local `import pytest`.** (`test_build_selected.py` ~87, inside `test_unknown_reg_class_and_physreg_raise`) — `pytest` is already imported at module top (~19). CLAUDE.md: "Every import goes at module top. No function-local imports." Both redundant and a rule violation. Flagged by 3 of 5 agents.

**6. `add_reg` hardcodes `isDebug`/`isInternalRead` to false.** (`Machine.cpp` ~244-245) — the binding exposes every other `CreateReg` flag (including `isRenamable`, which comes *after* these two) but bakes these two. CLAUDE.md "bind the full API; do not hardcode": surface them as args, or note the deliberate cut. (Also: the docstring's "full MachineOperand register flag set" slightly overstates given these two are omitted.)

**7. Multi-line `if` condition without braces.** (`Machine.cpp` ~343-345, in `reg_class`) — the `name == tri->getRegClassName(...)` condition wraps across two lines; CLAUDE.md requires braces when the condition spans multiple lines. (The `physreg` loop is fine — single-line.)

---

### 🟡 Suggestions — type/API design

- **Three ways to append a register operand** — `add_def(reg)`, `add_use(reg, implicit)`, and `add_reg(reg, is_def=..., ...)` — where `add_reg` strictly subsumes the other two. The convenience/full-control layering is defensible, but each inlines its *own* `CreateReg` call with independently-hardcoded defaults, so the "same operation with defaults" invariant lives only in naming, not delegation, and the paths can drift. Consider having `add_def`/`add_use` forward to a single shared operand-construction path.
- **`reg_class` returns `reference_internal`** though `TargetRegisterClass` instances are target-static/immortal — this installs a spurious keep-alive tying the `MachineFunction`'s lifetime to the handle. Plain `rv_policy::reference` models "borrowed, outlives everything" more accurately.
- **`physreg`/`reg_class` homed on `MachineFunction`** are really subtarget/TRI queries (same as `opcode()` last PR) — consistent with precedent, and the natural migration cluster if a `Subtarget`/`TargetRegisterInfo` type is ever bound.
- **`create_vreg` with a reg class from a different-target function** is silently wrong (reg-class pointers are target-static; cross-target isn't checked). Narrow footgun. (`None` is safely rejected by nanobind.)
- **`verify()` still returns a bare bool with the diagnostic discarded** (inherited from the prior PR) — this directly undercuts the hand-build workflow: a failed build gives `False` with no reason. Worth revisiting alongside #2.
- Ratings (type review): encapsulation 6/10, invariant expression 5/10, usefulness 8/10, enforcement 4/10.

---

### 🧪 Tests — happy path covered; the new flag/property surface is assertion-light

Three tests, all leak-checked: reg_class/physreg lookup + `create_vreg().is_virtual`, `KeyError` negatives for both lookups, and an end-to-end hand-built AArch64 add with the expected opcode sequence and `verify() is True`. Gaps:
- **No negative `verify()` test** (priority 8) — nothing builds the same MIR *without* the properties/liveins and asserts `verify() is False`, so `set_property`/`add_livein` are never shown to be load-bearing; both could be no-ops and the suite would pass. This is the highest-value addition and the only observability channel these write-only APIs have.
- **`add_reg`'s flag matrix is 6/9 untested** (priority 7) — only `is_def`, plain use, and `implicit` are exercised; `is_kill`/`is_dead`/`is_undef`/`is_early_clobber`/`sub_reg`/`is_renamable` are never set or asserted, so an argument-position mis-wiring in the 9-arg lambda (e.g. the `isRenamable` that sits after two skipped bools) would be invisible.
- **`add_livein` with a virtual register** (the #1 misuse) isn't tested to raise; **operand structure** (which operand defs/uses which reg) isn't asserted beyond the opcode sequence; **liveins/properties** are never read back.
- **AArch64-gated** — the whole module skips on x86 legs, but `reg_class`/`physreg`/`create_vreg`/`set_property`/`add_reg` are target-agnostic C++ and get zero coverage there; a target-neutral smoke test would help. No `to_mir` round-trip of the hand-built selected MIR either.

---

### ✅ Handled well
- Lookups fail loudly with specific `nb::key_error` messages; `physreg` correctly skips `NoRegister` (i=1); `verify()` correctly forces `AbortOnError=false` (no interpreter abort).
- `add_use`'s new `implicit` param defaults to false — backward compatible.
- `TargetRegisterClass` as an opaque token is the right default; the inline `/*isDebug=*/false, /*isInternalRead=*/false` annotations at the `CreateReg` call site are exemplary (they pin the flag ordering for future maintainers).
- `create_vreg(None)` is safely rejected by nanobind (`TypeError`).
- Comments/docstrings are otherwise accurate (the `add_reg` flag list maps 1:1 to the exposed args; the `add_livein` "physical register" wording correctly reflects the precondition; the GlobalISel-fallback rationale is correct).

*(Comment nits: `add_use`'s docstring wasn't updated for the new `implicit` param; the `MachineFunctionProperty` enum comment calls all seven values "pipeline-stage flags" when several — `TracksLiveness`/`NoPHIs`/`NoVRegs` — are descriptive invariants; and the test docstring's "substrate for lowering to machine code without instruction selection" is a borderline forward-reference to the not-yet-landed emit/JIT work, which CLAUDE.md discourages.)*

---

makslevental:
Addressed the review; changes are in a follow-up commit on this branch and cascaded up the stack (#590, #591).

**🔴 Silent-corruption paths (assert-only, invisible in the NDEBUG wheel)**
1. `add_livein(reg)` — now rejects a virtual register with `nb::value_error` before `asMCReg()` (which only asserts physicality, so a vreg was truncated into a bogus `MCRegister` livein).
2. `set_property` — kept (it's inherent to hand-building selected MIR: there's no codegen stage to earn the flags), but its docstring now spells out that it *asserts, not requests* — the verifier trusts these flags and picks which invariant classes to enforce from them (e.g. liveness only when `TracksLiveness`). Added a `has_property()` read side for symmetry.
3. `add_reg` — now rejects `is_kill` on a def, `is_dead` on a use, `is_renamable` on a virtual register, and out-of-range `sub_reg` (`getNumSubRegIndices()` bound). All were LLVM asserts.
4. `reg_class` / `physreg` — route the `getRegisterInfo()` deref through a new `requireTRI()` null-guard, mirroring `requireTII()`.

**🟠 CLAUDE.md**
5. Dropped the function-local `import pytest`.
6. `add_reg` now surfaces `is_debug` / `is_internal_read` (were hardcoded false) — bind the full flag set.
7. Braced the multi-line `reg_class` condition.

**🟡 Design**
- `add_def`/`add_use`/`add_reg` now funnel through one shared `addRegOperand()`, so the "same op with defaults" invariant lives in one place instead of three parallel `CreateReg` calls.
- `reg_class` returns `rv_policy::reference` (the class is target-static/immortal — no spurious keep-alive).
- Exposed the `MachineOperand` register flags (`is_implicit`/`is_kill`/`is_dead`/`is_undef`/`is_early_clobber`/`is_renamable`/`sub_reg`) so operand wiring is observable. `is_renamable` is physical-guarded (`isRenamable()` asserts a physreg).
- `verify()` returning bare bool: already resolved on the base branch — `verify_diagnostic()` (added in #588) returns the reason string; the new negative-verify test uses it.
- `create_vreg` cross-target: left as-is — reg-class pointers are target-static with no cheap ownership check; narrow footgun, noted.

**Comments**
- Fixed the `MachineFunctionProperty` enum comment (not all values are pipeline-stage flags — `IsSSA`/`NoPHIs`/`TracksLiveness`/`NoVRegs` are descriptive invariants), updated `add_use`'s docstring for `implicit`, and dropped the forward-referencing test docstring.

**🧪 Tests**
- Negative `verify()` without live-ins (shows `add_livein` is load-bearing) + non-empty `verify_diagnostic()`.
- Full `add_reg` flag-matrix round-trip (physical-register variant here, virtual-register variant target-neutral), read back via the new operand accessors — catches an argument-position mis-wire in the 11-arg builder.
- Contradictory-flag / out-of-range-`sub_reg` / virtual-`is_renamable` rejections; `add_livein`-virtual rejection; `has_property` read-back; operand-structure assertions on the hand-built add; `to_mir`→`parse_mir` round-trip of the selected MIR.
- New `test_build_selected_generic.py` covers the target-agnostic guards on non-AArch64 CI legs (the AArch64 module skips there).

Verified: PR branch builds + full suite green; `emit-jit` (the `MachineModuleInfo`/`run_codegen_to_mir` merge point) builds + 128 MIR tests; top of stack builds + 576 pass at 99.72% Python coverage; `src/IR` + `src/MIR` at 100% line+function C++ coverage (`Machine.cpp` 777/777 lines, 98/98 functions).

PR #590, opened 2026-08-17
[eudsl-llvmpy] Emit object from selected MIR + JIT-execute it (Route B capstone)

makslevental:
MachineModuleInfo.emit_object runs only the back half of codegen (register
allocation, prologue/epilogue, object emission) over the hand-built,
already-selected MachineFunctions -- via TargetMachine.addPassesToEmitFile with
-start-after=finalize-isel so no instruction selection runs -- and returns the
relocatable object bytes. LLJIT.add_object loads such an object into ORC so the
DSL-built function can be called. create_machine_function now builds into a
MachineModuleInfoWrapperPass (with a defined IR stub) so emit_object can hand it
to the pipeline, and freezes reserved registers first.

This closes Route B: a MachineFunction assembled instruction-by-instruction in
Python lowers to native code with no GlobalISel dependence -- verified by
JIT-executing add(2,3)==5 on an AArch64 host (object emission itself is
host-independent). Eleventh PR in the MIR stack.

Alternative recorded for the generic-MIR route: lowering DSL-built *generic*
(G_*) MIR needs the target's post-legalizer passes, reachable only from IR via
run_codegen_to_mir(global_isel=True) or, upstream, a programmatic
TargetPassConfig::setStartStopPasses; GlobalISel's SelectionDAG fallback needs
IR, which hand-built MIR lacks. Route B sidesteps all of that.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/JIT.cpp
+++ b/projects/eudsl-llvmpy/src/IR/JIT.cpp
@@ -40,6 +40,19 @@ void populate_jit(nb::module_ &m) {
           },
           "module"_a)
       .def(
+          "add_object",
+          [](llvm::orc::LLJIT &self, nb::bytes obj) {
+            std::unique_ptr<llvm::MemoryBuffer> memBuf =
+                llvm::MemoryBuffer::getMemBufferCopy(
+                    llvm::StringRef(static_cast<const char *>(obj.data()),
+                                    obj.size()),
+                    "<jit-obj>");
+            eudsl::unwrap(self.addObjectFile(std::move(memBuf)));
+          },
+          "obj"_a,
+          "Add a relocatable object file (e.g. from "
+          "MachineModuleInfo.emit_object) to the JIT.")
+      .def(
           "lookup",
           [](llvm::orc::LLJIT &self, const std::string &name) {
             llvm::orc::ExecutorAddr addr = eudsl::unwrap(self.lookup(name));
--- a/projects/eudsl-llvmpy/src/IR/JIT.cpp
+++ b/projects/eudsl-llvmpy/src/IR/JIT.cpp
@@ -11,6 +11,7 @@
 #include <llvm/ExecutionEngine/Orc/ThreadSafeModule.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/Module.h>
+#include <llvm/Object/ObjectFile.h>
 #include <llvm/Support/MemoryBuffer.h>
 
 #include <memory>
@@ -47,6 +48,14 @@ void populate_jit(nb::module_ &m) {
                     llvm::StringRef(static_cast<const char *>(obj.data()),
                                     obj.size()),
                     "<jit-obj>");
+            // ORC's addObjectFile is lazy: it records the buffer and defers
+            // parsing to materialization (a later lookup), so bad bytes would
+            // otherwise surface as an opaque error attributed to the wrong
+            // operation. Parse eagerly here so add_object itself raises on a
+            // non-object / truncated / wrong-format buffer.
+            eudsl::unwrap(llvm::object::ObjectFile::createObjectFile(
+                              memBuf->getMemBufferRef())
+                              .takeError());
             eudsl::unwrap(self.addObjectFile(std::move(memBuf)));
           },
           "obj"_a,
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -29,9 +29,12 @@
 #include <llvm/IR/Function.h>
 #include <llvm/IR/GlobalValue.h>
 #include <llvm/IR/InstrTypes.h>
+#include <llvm/IR/Instructions.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/LegacyPassManager.h>
+#include <llvm/Support/CommandLine.h>
 #include <llvm/Support/MemoryBuffer.h>
+#include <llvm/Support/raw_ostream.h>
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
 
@@ -196,21 +199,25 @@ static std::vector<llvm::MachineIRBuilder *> &machineBuilderStack() {
   return stack;
 }
 
-// Owns everything the inspected MachineFunctions transitively depend on, so
-// none of it is freed out from under them: the LLVMContext (pinned by
-// `ctxKeepAlive`), the IR Module (whose Functions the MachineFunctions
-// reference), and the MachineModuleInfo that owns the MachineFunctions. The MMI
-// is owned two ways depending on how it was built: the codegen path
-// (run_codegen_to_mir) leaves it inside a MachineModuleInfoWrapperPass owned by
-// `pm`; the parse path (parse_mir) owns it directly via `ownedMmi`. `mmi` is
-// the query handle either way. The TargetMachine is a separate Python object
-// kept alive by a keep_alive.
+// Owns everything the MachineFunctions transitively depend on, so none of it is
+// freed out from under them: the LLVMContext (pinned by `ctxKeepAlive`), the IR
+// Module (whose Functions the MachineFunctions reference), and the
+// MachineModuleInfo that owns the MachineFunctions. The MMI is owned one of
+// three ways depending on how it was built: run_codegen_to_mir leaves it in a
+// MachineModuleInfoWrapperPass owned by the (already-run) `pm`; parse_mir owns
+// it directly via `ownedMmi`; create_machine_function owns a
+// MachineModuleInfoWrapperPass via `mmiwp` (not yet in a PassManager, so
+// emit_object can hand it to one). `mmi` is the query handle. `tm` is borrowed
+// (kept alive by a keep_alive on the factory).
 struct MachineModuleInfo {
   std::shared_ptr<llvm::LLVMContext> ctxKeepAlive;
   std::unique_ptr<llvm::Module> module;
-  std::unique_ptr<llvm::legacy::PassManager> pm;     // codegen path
+  std::unique_ptr<llvm::legacy::PassManager> pm;     // codegen/emit path
   std::unique_ptr<llvm::MachineModuleInfo> ownedMmi; // parse path
+  std::unique_ptr<llvm::MachineModuleInfoWrapperPass> mmiwp; // build path
   llvm::MachineModuleInfo *mmi;
+  llvm::TargetMachine *tm = nullptr;
+  bool emitted = false; // emit_object runs a pass pipeline once
 
   // Print the whole module as .mir text: the IR block, then each function's
   // machine-level block, matching what `llc -stop-after=finalize-isel` emits. A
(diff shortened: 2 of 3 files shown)


makslevental (commented):
## Comprehensive PR Review — #590 (Emit object from selected MIR + JIT-execute it — Route B capstone)

Ran a five-aspect review (general code, tests, error handling, type design, comments) over the delta against the base branch (`JIT.cpp`, `Machine.cpp`, new `test_emit_jit.py`). This closes Route B, and the mechanism is impressive: a MachineFunction assembled instruction-by-instruction in Python lowers to native code and JIT-executes `add(2,3)==5` with no GlobalISel dependence. The general-code review independently traced the trickiest part — the now-three-way ownership struct and the `mmiwp.release()`→`pm` transition — and **confirmed it's correct** (destruction order safe, no UAF/leak, aggregate-init order matches field order in all three factories, `-start-after` timing and `freezeReservedRegs` placement both right, the defined-stub/MF-body divergence sound under `-start-after=finalize-isel`). The `add_object`/`emit_object` byte-pipe API is clean. The findings cluster around one theme the whole stack has been building toward: **`emit_object` is the first consumer of the prior PRs' unchecked MIR-construction primitives, and it runs codegen with the verifier disabled and codegen diagnostics unhandled** — so malformed hand-built MIR becomes a silent bad object or a process abort rather than a catchable error.

> Review-only — nothing was changed. Line numbers are approximate to the PR-head files.

---

### 🔴 High — malformed hand-built MIR fails silently or fatally (not as a Python error)

**1. `emit_object` runs the pipeline with `DisableVerify=true`.** (`Machine.cpp` ~491)
`addPassesToEmitFile(..., /*DisableVerify=*/true, ...)` suppresses the `MachineVerifierPass` — the *one* check that would catch the malformed MIR the prior PRs' unchecked primitives can produce (`set_property` asserting `IsSSA`/`TracksLiveness` that don't hold, `add_reg` contradictory flags, `add_livein` on a vreg, undefined vregs). Since the release wheel is `NDEBUG`, the in-pass backup asserts are also gone. So emission proceeds over invalid MIR and either muddles through to a **garbage object returned as success**, or a codegen pass hits an impossible state (finding #2). Consider verifying first, or exposing a `disable_verify` argument rather than hardcoding the unsafe default.

**2. `pm->run` result ignored + no diagnostic/fatal handler installed.** (`Machine.cpp` ~500)
A codegen-pass failure (e.g. regalloc on malformed MIR) surfaces either as `report_fatal_error` → stderr + `abort()` (killing the interpreter; the RAII restore and nanobind unwinding don't run), or a `DS_Error` context diagnostic → stderr + `exit(1)`. Neither is a catchable Python exception — the same swallowed-diagnostic pattern flagged throughout this stack, now on a path where hand-built input can genuinely fail. Combined with #1, this is the central hole: malformed input yields silent corruption or sudden death, never an actionable error.

**3. `add_object` accepts arbitrary bytes; failure is deferred to `lookup`.** (`JIT.cpp` ~42-53)
`getMemBufferCopy` copies the bytes with no validation that they're an object file, and ORC's `addObjectFile` is *lazy* — it records the buffer and returns success; parsing/materialization happens later. So `add_object(b"garbage")`, a truncated object, or a wrong-arch object returns "success," and the failure re-emerges at a distant `lookup` as an opaque materialization error (routed through ORC's handler, not the eager `unwrap`). Users get an error attributed to the wrong operation. *(The `unwrap` itself is correct — it just only catches eager errors, which lazy loading essentially never populates.)*

---

### 🟠 Important

**4. Process-global `-start-after` cl::opt mutation.** (`Machine.cpp` ~467-481)
`emit_object` mutates the registered `start-after` command-line option to `"finalize-isel"` and RAII-restores it. Timing and the restore-on-C++-throw are both correct (verified by code review), and the GIL serializes Python callers today — but it's a process-global with no locking (concurrent/nested/free-threaded codegen would silently clobber each other's start point) and it isn't restored on a `report_fatal_error` abort (finding #2's path). It's also inherently fragile (programmatically mutating an llc CLI option); the comment acknowledges this. Not a bug for current single-threaded use, but worth a guard/note. *(The two `throw` sites here are correct RuntimeErrors.)*

**5. CLAUDE.md C++ brace style — two multi-line `if`/`throw` bodies without braces.** (`Machine.cpp` ~489-491 and ~457-460)
The `addPassesToEmitFile` `if` has a 3-line condition *and* a 2-line body; the `requires create_machine_function` guard has a multi-line `throw` body. CLAUDE.md requires braces when the condition or body spans multiple lines. *(The `LCOV_EXCL_LINE` throws match a pre-existing un-braced convention in the file, so only the non-excluded new instances are flagged.)*

**6. Forward-reference: `(see #589)` PR-number reference.** (`test_emit_jit.py` ~39, `_build_selected_add` docstring)
CLAUDE.md's "no forward-reference comments" rule explicitly forbids task/PR numbers because they go stale on squash-merge. Reword to describe what the helper builds, or reference the stable concept without the number.

---

### 🟡 Type/API design & minor

- **The MachineModuleInfo struct is now a three-way ownership union** (`pm` / `ownedMmi` / `mmiwp`) plus `tm` (borrowed) and `emitted` (phase flag), expressed as flat public fields discriminated only by null-checks and a load-bearing comment — and `emit_object` produces an intentional *two-owner terminal state* (`pm` owns the released `mmiwp`, `mmi` still valid). The code is *correct* (both code review and comment review verified the comment matches all three factories), but the type review argues it has crossed the threshold where distinct per-path types or a `std::variant` + explicit `Phase` enum would make "emit_object only on the build path," the one-shot, and the post-emit state compile-time facts instead of runtime throws + prose. Invariant-expression rated 2/10 (regressed as the union grew); the runtime guards (`emitted`, `mmiwp && tm`) are a genuine, well-placed enforcement improvement. Worth considering before the struct grows further.
- **`emitted` as a mutable bool** on a public struct is the weakest coherent way to express "consume once"; a consuming signature or a distinct `EmittedObject` return type would record the transition in the type system.
- **`emit_object` returns bytes with no non-empty/real-object guarantee** (`Machine.cpp` ~502) — the *test* checks ELF magic, but the binding contract doesn't, so a truncated/empty buffer could be returned as success and fed into #3.
- **`add_object` returns `None`** — consistent with the sibling `add_module`, so not a defect; but neither exposes an ORC `ResourceTracker` handle, so a loaded object can't be individually removed/re-scoped. A symmetric upgrade for both methods if unloading ever matters.
- Minor comment imprecision: `emit_object`'s leading comment says "asm/object emission" but the code hardcodes `ObjectFile` only.

---

### 🧪 Tests — solid guards; the capstone claim is only verified on arm64

Four tests, all leak-checked (the JIT test correctly `del j` before `assert_no_leaks()`): ELF-magic emission, double-emit raises, the create_machine_function-required guard, and JIT-execute `add(2,3)==5`/`add(40,2)`/`add(-5,5)`. Dual gating (module-level AArch64-backend + host-level `_IS_AARCH64`) is the right layering. Gaps:
- **Emitted-object correctness is verified only on AArch64 hosts** (priority 7) — the execution test is host-gated, so x86 legs only check `len>0` + ELF magic, which would pass for an object with no defined `add` symbol / empty `.text`. A host-independent symbol-table assertion (defined `add` symbol exists) would guard the whole point on every leg.
- **The `-start-after` global restore is untested** (priority 7) — nothing proves a `run_codegen_to_mir` after an `emit_object` still runs full ISel; a restore regression would silently contaminate later codegen in the same process.
- **`add_object` with malformed bytes** isn't tested to raise (host-independent, cheap); the `parse_mir` path also hits the requires-create_machine_function guard but only `run_codegen_to_mir` is tested; post-emission MMI validity (`to_mir`/`machine_function` after `emit_object`) isn't exercised; no multi-function/cross-call object.

---

### ✅ Handled well
- **3-path ownership + the `mmiwp.release()`→`pm` transition is correct** — destruction order safe (no UAF), no leak from overwriting the null build-path `pm`, aggregate-init order matches field order in all three factories.
- `-start-after` set/restore timing is right (set before `addPassesToEmitFile`, restored on throw); `freezeReservedRegs` correctly runs after the pipeline is built but before `pm->run`; the defined-IR-stub vs hand-built-MF-body divergence is sound under `-start-after=finalize-isel`.
- `add_object` correctly copies the transient Python bytes into an owned `MemoryBuffer` (self-contained after load); `unwrap` propagates eager errors; the `emitted` and `mmiwp && tm` guards raise clean `RuntimeError`s with stable messages (well-tested via `pytest.raises(match=...)`).
- The struct ownership comment is exemplary "why" documentation and stays synchronized with all three call sites; the emit_object RAII-restore and freeze-reserved-regs comments explain genuinely non-local requirements.

---

makslevental:
Addressed the review; changes are in a follow-up commit on this branch and cascaded up the stack (#591).

**🔴 Malformed hand-built MIR was silent/fatal, not a Python error**
1. **`emit_object` ran with `DisableVerify=true`** — it now **verifies every MachineFunction up front** and raises `RuntimeError` (`withDetail`) with the verifier's report if the MIR is malformed, *before* building the emission pipeline. The pipeline keeps `DisableVerify=true` since we've already verified.
2. **`pm->run` diagnostics were unhandled** — wrapped it in a `ScopedDiagnosticCapture` so a back-half codegen `DS_Error` surfaces as an exception instead of stderr+`exit`, and added an empty-buffer guard. (Both unreachable from well-formed MIR once the eager verify is in place, so `LCOV_EXCL`.)
3. **`add_object` deferred failure to `lookup`** — ORC's `addObjectFile` is lazy, so bad bytes returned "success" and blew up later at an unrelated `lookup`. It now parses eagerly (`ObjectFile::createObjectFile`) so `add_object` itself raises on a non-object / truncated / wrong-format buffer.

**🟠 Important**
4. **Process-global `-start-after`** — documented that the set/restore relies on the GIL serializing callers (no lock; no concurrent/nested/free-threaded codegen). Kept the mechanism (timing + RAII restore were confirmed correct); a lock isn't warranted for current single-threaded use.
5. **CLAUDE.md braces** — braced the multi-line `create_machine_function`-required guard, and braced the two `LCOV_EXCL` throw blocks while wrapping them in `LCOV_EXCL_START/STOP` (matching the existing `addISelPasses` convention) so the closing brace stays out of the coverage denominator.
6. **Forward reference** — dropped the `(see #589)` in the test helper docstring.

**🟡 Minor**
- Fixed the `emit_object` comment ("asm/object emission" → "object emission"; `ObjectFile` only).
- `emit_object` now guarantees a non-empty buffer (raises otherwise), closing the truncated/empty-object hole that fed into #3.

**Pushed back / not done:** the `std::variant` + `Phase`-enum redesign of the 3-way `MachineModuleInfo` ownership struct (and an `EmittedObject` return type for consume-once) — the review itself rated the code *correct* and the runtime guards well-placed; a type-system rewrite is a larger refactor better done deliberately than folded into a review pass, especially as later PRs don't extend this struct. `ResourceTracker`-based unloading for `add_object`/`add_module` is YAGNI (nothing unloads today).

**🧪 Tests:** reject malformed MIR (physregs without live-ins → `failed verification`); `add_object(b"...")` non-object bytes raises; `-start-after` restore (a `run_codegen_to_mir` after `emit_object` still runs full ISel — verifies + non-empty body); and a host-independent symbol-table assertion (`b"add\x00" in obj`) so every leg — not just AArch64 hosts — checks the object actually defines `add`.

**Verified:** emit-jit branch builds + full suite green; top of stack builds + 579 pass at 99.72% Python coverage; `src/IR` + `src/MIR` at 100% line+function C++ coverage (`Machine.cpp` 840/840 lines, 100/100 functions). Cascade to #591 was clean (no conflicts); force-push rewrote #591.

PR #601, opened 2026-08-23
[eudsl-llvmpy] Extract shared PassPipelineEnv scaffold in Passes.cpp

makslevental:
run_passes and run_default_pipeline built the PassBuilder + the four
analysis managers + StandardInstrumentations identically. Pull that setup
into a PassPipelineEnv struct (declared so member construction order gives
the lifetimes ModulePassManager::run needs) so the two copies cannot drift
and later Python-pass entry points can reuse it. Pure refactor.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -10,6 +10,37 @@
 #include <llvm/Passes/PassBuilder.h>
 #include <llvm/Passes/StandardInstrumentations.h>
 
+namespace {
+// Bundles the PassBuilder with the four analysis managers and instrumentation,
+// with the lifetimes ModulePassManager::run requires: the managers and the
+// StandardInstrumentations must outlive the run that references them, and the
+// PassBuilder holds a pointer to `pic`. run_passes and run_default_pipeline
+// build this identically, so keeping it in one place stops the two copies from
+// drifting. Member declaration order is the construction order: `pic` before
+// `pb` (which captures &pic), `si` before its registerCallbacks(pic).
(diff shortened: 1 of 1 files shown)


makslevental (commented):
## Automated review — focus: correctness, honesty, coverage

**Verdict: no substantive issues found.** Clean, behavior-preserving extraction of the shared `PassBuilder` + analysis-manager scaffold.

Verified:
- **Member construction order (the main hazard) is correct.** Declaration order in `PassPipelineEnv` is `pic, si, pb, lam, fam, cgam, mam`; C++ constructs in declaration order, so `pic` is fully constructed before `pb(..., &pic)` captures its address, and `si` before `si.registerCallbacks(pic)` in the body. The init-list order (`si` then `pb`) is harmless.
- **Lifetimes vs `ModulePassManager::run` are correct.** `env` is declared before the local `mpm` at both call sites, so the managers/instrumentation outlive `mpm.run(mod.get(), env.mam)`. Reverse destruction frees `mpm` first, then `pb` before `pic` — no dangling pointer at teardown.
- **Behavior-preserving.** One benign ordering change: `pb` is now constructed (member init) before `si.registerCallbacks(pic)` runs in the body, whereas the original registered callbacks before constructing `pb`. Inert — `PassBuilder`'s ctor only stores the `PassInstrumentationCallbacks*`; callbacks are still registered into the same `pic` before any `run`.
- **CLAUDE.md compliance** OK (anon-namespace placement, brace rules, no LLVM-name renaming).

**Coverage/tests:** no test change warranted — a pure refactor with identical observable behavior, adequately guarded by `tests/passmanager/test_passes.py`.

PR #602, opened 2026-08-23
[eudsl-llvmpy] Add module-level Python passes (run_python_pass)

makslevental:
Introduce PyModulePass, a new-PM module pass whose body is a Python
callable. The new PassManager is concept-based, so the pass only needs a
run(Module&, ModuleAnalysisManager&) plus PassInfoMixin -- no LLVM base
class. run_python_pass builds the shared PassPipelineEnv, adds the pass,
and runs it in place; the callback receives the same Module wrapper the
caller passed (rv_policy::reference), and a truthy return marks the IR
mutated (invalidate analyses) while None/falsy preserves all.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -5,12 +5,36 @@
 #include "IR/Common.h"
 #include "IR/Ownership.h"
 
+#include <llvm/ADT/Any.h>
 #include <llvm/IR/PassManager.h>
 #include <llvm/Passes/OptimizationLevel.h>
 #include <llvm/Passes/PassBuilder.h>
 #include <llvm/Passes/StandardInstrumentations.h>
 
+#include <exception>
+
 namespace {
+// A Python exception raised inside a pass callback is stashed here instead of
+// thrown across llvm::ModulePassManager::run, which libLLVMCore compiles with
+// -fno-exceptions (LLVM_ENABLE_EH OFF): unwinding through those frames skips
+// their destructors (e.g. leaves the thread-local PrettyStackTraceHead dangling
+// and leaks the run's PreservedAnalyses) and std::terminates on targets built
+// without asynchronous unwind tables. runPipeline resets the slot, runs the
+// pipeline, and re-raises afterward, so the throw only crosses our own
+// -fexceptions frames. Thread-local so concurrent pipelines don't clobber it.
+thread_local std::exception_ptr pendingPassError;
+
+void runPipeline(llvm::ModulePassManager &mpm, llvm::Module &m,
+                 llvm::ModuleAnalysisManager &mam) {
+  pendingPassError = nullptr;
+  mpm.run(m, mam);
+  if (pendingPassError) {
+    std::exception_ptr err = pendingPassError;
+    pendingPassError = nullptr;
+    std::rethrow_exception(err);
+  }
+}
+
 // Bundles the PassBuilder with the four analysis managers and instrumentation,
 // with the lifetimes ModulePassManager::run requires: the managers and the
 // StandardInstrumentations must outlive the run that references them, and the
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## Automated review (consolidated) — implementation, tests, honesty, exception mechanism

Covers `run_python_pass_on_module` + `PyModulePass`, the capture/rethrow exception mechanism, and the pipeline-skip fix. Updated to reflect fixes made since the first pass.

**Implementation — sound.**
- Same-wrapper identity via `nb::cast(mod, rv_policy::reference)` (guarded by `assert seen[0] is mod`); lifetimes correct (the module is kept alive for the synchronous call; the `nb::callable` moves into `PassModel` with refcount preserved).
- `nb::gil_scoped_acquire` in `run()` is the correct defensive choice for the `FREE_THREADED` build.
- Return mapping `PyObject_IsTrue` → `none()`/`all()` matches the None/falsy = unchanged convention.
- Reentrancy/nesting is safe; `exception_ptr` + GIL handling correct; `catch(...)` breadth appropriate; CLAUDE.md clean.

**Exception mechanism — sound, and the gap raised in re-review is now fixed.**
- ✅ **FIXED — "a stashed error didn't stop the rest of the pipeline."** `PassPipelineEnv` now registers a `ShouldRunOptionalPassCallback` returning `!pendingPassError`, so once a callback stashes an error every subsequent *optional* pass is skipped and the pipeline winds down to `runPipeline`'s re-raise instead of running builtins on half-mutated IR (composition test in #605).
- Residual (inherent, documented): `isRequired()` passes still run (can't be skipped without throwing); the throwing pass's own partial mutation isn't rolled back.
- Low-severity edges still open: (a) `nb::gil_scoped_acquire` is constructed *before* the `try`, so a throw from its construction wouldn't be stashed (very low probability); (b) if a genuine C++ exception propagates out of `mpm.run()` while an error is already stashed, the stashed one is dropped in favor of it (an error still surfaces).

**Naming:** `run_python_pass` → `run_python_pass_on_module` (paired with `run_python_pass_on_function`, #604).

**Tests / coverage:** the module tests here are honest (they fail on a no-op; the `is mod` identity check is strong). Exception-propagation tests live in #603; the composition/skip + error-through-`run_passes` test in #605. Remaining minor gap: the truthy→`none()` (mutated) return branch isn't directly asserted (a fixture mutates but returns `None`).

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -39,6 +39,36 @@ struct PassPipelineEnv {
...
+      : mod(mod), callback(std::move(callback)) {}
+
+  llvm::PreservedAnalyses run(llvm::Module &, llvm::ModuleAnalysisManager &) {
+    nb::gil_scoped_acquire gil;
+    nb::object res = callback(nb::cast(mod, nb::rv_policy::reference));
+    // Convention: None or any falsy return means "IR unchanged" (preserve all
+    // analyses); any truthy return means the pass mutated the IR.
+    int truthy = PyObject_IsTrue(res.ptr());
+    if (truthy < 0)
+      throw nb::python_error();
```
makslevental: needs a message here

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -95,4 +125,19 @@ void populate_passes(nb::module_ &m) {
       "module"_a, "level"_a, "tuning"_a = nb::none(), "debug"_a = false,
       "verify_each"_a = false,
       "Run the default optimization pipeline at the given level.");
+
+  m.def(
+      "run_python_pass",
+      [](eudsl::Module &mod, nb::callable callback) {
+        PassPipelineEnv env(mod.get().getContext(),
+                            llvm::PipelineTuningOptions(), /*debug=*/false,
+                            /*verifyEach=*/false);
```
makslevental: these shouldn't be hardcoded (we have bindings for `PipelineTuningOptions` so that can be passed in as well)

makslevental (commented):

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -37,6 +61,49 @@ struct PassPipelineEnv {
...
+
+  llvm::PreservedAnalyses run(llvm::Module &, llvm::ModuleAnalysisManager &) {
+    nb::gil_scoped_acquire gil;
+    try {
+      nb::object res = callback(nb::cast(mod, nb::rv_policy::reference));
+      // Convention: None or any falsy return means "IR unchanged" (preserve all
+      // analyses); any truthy return means the pass mutated the IR.
+      int truthy = PyObject_IsTrue(res.ptr());
+      if (truthy < 0)
+        throw nb::python_error();
```
makslevental: this needs a message

projects/eudsl-llvmpy/src/IR/Passes.cpp
makslevental: Done in 236c27a4a0f8404d9f9a4c341da9699d238dec22 — the `PyObject_IsTrue(<0)` branch now raises via `nb::raise_from(err, PyExc_ValueError, ...)` with the message "could not evaluate the truthiness of a Python module pass's return value", chaining the original error as `__cause__`. Test asserts both the message and the chained cause.

projects/eudsl-llvmpy/src/IR/Passes.cpp
makslevental: Done in 236c27a4a0f8404d9f9a4c341da9699d238dec22 — the `PyObject_IsTrue(<0)` branch now raises via `nb::raise_from(err, PyExc_ValueError, ...)` with the message "could not evaluate the truthiness of a Python module pass's return value", chaining the original error as `__cause__`. Test asserts both the message and the chained cause.

projects/eudsl-llvmpy/src/IR/Passes.cpp
makslevental: Done in 236c27a4a0f8404d9f9a4c341da9699d238dec22 — `run_python_pass_on_module` now takes `tuning`/`debug`/`verify_each` (defaulting to `None`/`False`), mirroring `run_passes`/`run_default_pipeline`, instead of hardcoding `PipelineTuningOptions()`.

makslevental:
## Follow-up on the consolidated review — remaining items addressed (b7f859ddd)

Working through the "remaining/open" items from the [consolidated review comment](https://github.com/llvm/eudsl/pull/602#pullrequestreview-5003323168):

**Inline comments — fixed:**
- `PyObject_IsTrue(<0)` now raises via `nb::raise_from(err, PyExc_ValueError, ...)` with a descriptive message ("could not evaluate the truthiness of a Python module pass's return value"), chaining the original error as `__cause__`. Asserted in `test_exception_from_unboolable_return_propagates`.
- `run_python_pass_on_module` no longer hardcodes the pipeline options — it takes `tuning`/`debug`/`verify_each` mirroring `run_passes`/`run_default_pipeline`. Covered by `test_module_pass_forwards_tuning_and_flags`.

**Low-severity exception edges:**
- **(a) GIL guard outside the `try`** — this is deliberate and left as-is: the `catch` stashes via `std::current_exception()` (which for an `nb::python_error` touches Python refcounts and so must run while the GIL is held), and `PyGILState_Ensure` does not throw. Added a comment documenting exactly this so it isn't "fixed" into a regression later.
- **(b) a C++ exception from `mpm.run()` dropping a stashed Python error** — on closer inspection the premise is false: LLVM is built `-fno-exceptions`, and the only `-fexceptions` frames below `runPipeline` are the trampolines, which always `catch(...)` internally and return normally. No C++ exception can unwind into `runPipeline`, so there is nothing to guard. Left `runPipeline` in its simple form (no dead `LCOV_EXCL` guard).

**Test/coverage gap — closed:**
- Added `test_module_pass_can_mutate_ir_and_report_changed`, whose callback returns a truthy value, exercising the truthy → `PreservedAnalyses::none()` path that no prior fixture hit (all returned `None`).

**Verification:** C++ coverage on this branch's own state is 100% (Passes.cpp 106/106 lines, 9/9 functions; Values.cpp 339/339, 60/60); `tests/passmanager/` is 17 passed.

The only items left are the two **inherent** residuals already noted in the review, which aren't code-fixable: `isRequired()` passes still run after a callback stashes an error, and a throwing pass's own partial mutation isn't rolled back.

PR #603, opened 2026-08-23
[eudsl-llvmpy] Guard error semantics of run_python_pass

makslevental:
Add tests pinning the behavior when a Python module pass raises: the
exception propagates out of run_python_pass, the module stays usable, and
the callable is released even on the error path (weakref check). Also pin
that an "unchanged" (falsy-returning) pass can run repeatedly.

No production change was needed: nanobind unwinds a raised Python
exception cleanly through ModulePassManager::run on this LLVM build, so
the explicit capture/rethrow anticipated in the plan is unnecessary here.

### DIFF
--- a/projects/eudsl-llvmpy/tests/passmanager/test_python_passes.py
+++ b/projects/eudsl-llvmpy/tests/passmanager/test_python_passes.py
@@ -3,6 +3,9 @@
 #  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 from textwrap import dedent
 
+import gc
+import weakref
+
 import pytest
 
 import llvm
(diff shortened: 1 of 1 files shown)


makslevental (commented):
## Automated review (consolidated) — test honesty + mechanism coverage

Test-only PR pinning the error semantics of `run_python_pass_on_module` against the capture/rethrow trampoline. Updated to reflect the fixes made since the first pass.

**Honesty: clean.** Every comment accurately scopes what its test proves.
- `test_exception_from_unboolable_return_propagates` genuinely exercises the `PyObject_IsTrue(<0) -> throw` branch (the lambda returns successfully; the error comes only from `__bool__` raising). `match="no truth value"` is tight.
- `test_none_returning_pass_runs_repeatedly_without_error` is honest — the comment correctly disclaims analysis preservation (fresh managers per call).
- The `weakref` test's fragility note is accurate (release is via `mpm` teardown; depends on the raised exception's traceback not outliving the `pytest.raises` block).

**✅ Corrected:** the earlier "no production change needed" claim is gone — the trampoline now captures the exception and `runPipeline` re-raises it after `mpm.run()` (production change in #602/#604), and these tests validate that path.

**Inherent limitation (worth stating):** the exception tests pin the *error semantics*, not the *portability hardening* — the observable contract (a `ValueError` reaches Python, module still usable) is identical whether captured-and-rethrown or thrown directly; the leak gate only counts `Context`/`Module`, so it can't observe the prevented `PreservedAnalyses`/`PrettyStackTraceHead` leak. Acceptable (best available proxy).

**Coverage gaps in the first-error-wins logic:**
- ✅ **Closed:** error-path through `run_passes` (a raising pass composed with a builtin) is now covered by `test_raising_pass_skips_rest_of_pipeline` (#605), which also asserts the builtin is skipped.
- Still open (minor): no test runs two module-level Python passes (`"p1,p2"`) to assert the module-pass early-skip; and the function-pass first-error-wins is exercised but not asserted (#604 uses identical per-function messages). Both deferrable.

makslevental (commented):

makslevental:
## Follow-up on the consolidated review — remaining items addressed

Working through the two "still open (minor)" coverage gaps from the [consolidated review comment](https://github.com/llvm/eudsl/pull/603#pullrequestreview-5003334126):

- **Function-pass first-error-wins not asserted** — already resolved. `test_exception_in_function_pass_propagates` (#604) now uses distinct per-function messages (`raise ValueError("boom in " + fn.name)`) and asserts both `match="boom in f"` and `assert seen == ["f"]`, so it pins that the first function's error propagates and the second function is skipped. The review note predates that change.

- **No test runs two module-level Python passes to assert the module-pass early-skip** — closed in #605 (1e3daf219), where the `register_python_pass` API needed for a named Python-pass pipeline actually exists (it isn't available at this PR's stack position). Added `test_raising_module_pass_skips_later_python_pass`, which composes two registered module passes as `"p1,p2"` and asserts p2's callback never runs after p1 raises — pinning that the ShouldRun callback gates a later *Python* pass, not just builtins (the prior skip test only composed with `instcombine`).

Everything else in the review was already marked clean or ✅ corrected. The stated **inherent limitation** stands and isn't code-fixable: these tests pin the error *semantics* (a `ValueError` reaches Python, module still usable), not the portability hardening, since the leak gate only counts `Context`/`Module` and can't observe the prevented `PreservedAnalyses`/`PrettyStackTraceHead` leak.

**Verification** (at #605's own state, where the new test lives): C++ coverage 100% (Passes.cpp 188/188 lines, 18/18 functions; Values.cpp 339/339, 60/60); `tests/passmanager/test_python_passes.py` 19 passed.

PR #604, opened 2026-08-23
[eudsl-llvmpy] Add function-level Python passes (run_python_function_pass)

makslevental:
PyFunctionPass mirrors PyModulePass at function scope; run_python_function_pass
wraps it in a ModuleToFunctionPassAdaptor so it runs once per defined
function, handing the callback the llvm::Function. A separate entry point
(rather than a mode arg on run_python_pass) keeps the two distinct callback
signatures explicit.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -119,6 +119,33 @@ struct PyModulePass : llvm::PassInfoMixin<PyModulePass> {
     }
   }
 };
+
+// The per-function analogue of PyModulePass. Wrapped in a
+// ModuleToFunctionPassAdaptor it runs once per defined function; the callback
+// receives the llvm::Function (bound directly, so nanobind hands back its
+// wrapper). Same truthy-return / GIL contract as PyModulePass.
+struct PyFunctionPass : llvm::PassInfoMixin<PyFunctionPass> {
+  nb::callable callback;
+
+  explicit PyFunctionPass(nb::callable callback)
+      : callback(std::move(callback)) {}
+
+  llvm::PreservedAnalyses run(llvm::Function &f,
+                              llvm::FunctionAnalysisManager &) {
+    nb::gil_scoped_acquire gil;
+    try {
+      nb::object res = callback(nb::cast(&f, nb::rv_policy::reference));
+      int truthy = PyObject_IsTrue(res.ptr());
+      if (truthy < 0)
+        throw nb::python_error();
+      return truthy ? llvm::PreservedAnalyses::none()
+                    : llvm::PreservedAnalyses::all();
+    } catch (...) {
+      pendingPassError = std::current_exception();
+      return llvm::PreservedAnalyses::all();
+    }
+  }
+};
 } // namespace
 
 void populate_passes(nb::module_ &m) {
@@ -193,4 +220,20 @@ void populate_passes(nb::module_ &m) {
       "Run a Python callable as a module pass over the module in place. The "
       "callable receives the Module; return a truthy value if it mutated the "
       "IR (so analyses are invalidated), None/falsy otherwise.");
+
+  m.def(
+      "run_python_pass_on_function",
+      [](eudsl::Module &mod, nb::callable callback) {
+        PassPipelineEnv env(mod.get().getContext(),
+                            llvm::PipelineTuningOptions(), /*debug=*/false,
+                            /*verifyEach=*/false);
+        llvm::ModulePassManager mpm;
+        mpm.addPass(llvm::createModuleToFunctionPassAdaptor(
+            PyFunctionPass(std::move(callback))));
+        runPipeline(mpm, mod.get(), env.mam);
+      },
+      "module"_a, "callback"_a,
+      "Run a Python callable as a function pass over each defined function in "
+      "the module in place. The callable receives a Function; return a truthy "
+      "value if it mutated the function, None/falsy otherwise.");
 }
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## Automated review (consolidated) — function trampoline: implementation, tests, exception mechanism

Covers `run_python_pass_on_function` + `PyFunctionPass` and its exception hardening. Updated to reflect fixes since the first pass.

**Implementation — sound.**
- `createModuleToFunctionPassAdaptor` runs the single `PyFunctionPass` (one `nb::callable`) once per *defined* function (declarations skipped). `nb::cast(&f, rv_policy::reference)` hands back the correct `Function` wrapper. Renaming during iteration is safe (`setName` touches the value symbol table, not the module's function list).
- A separate `run_python_pass_on_function` entry point (vs a mode arg) keeps the distinct callback signatures explicit — reasonable and consistent with `run_python_pass_on_module`.

**Exception hardening — correctly mirrors the module pass.** `PyFunctionPass::run` has the same early-skip + `try/catch(...)` stash; `run_python_pass_on_function` routes through `runPipeline`. Multi-function unwinding is clean: every `run()` returns normally (the catch-all converts a raise into stash + `all()`), so no C++ exception traverses the adaptor's extra `-fno-exceptions` frames — `runPipeline` re-raises afterward.

**Naming:** `run_python_function_pass` → `run_python_pass_on_function` (paired with `run_python_pass_on_module`, #602).

**Tests / coverage:**
- `test_function_pass_runs_once_per_function` is load-bearing (`sorted(names) == ["f","g"]` catches double-invocation and module-wide invocation). Mutation test proves write-through on both functions.
- `test_exception_in_function_pass_propagates` validates propagation but **does not assert first-error-wins** — both functions raise the same message, so it would still pass if the early-skip were removed and the 2nd function re-entered. To lock it in: record invocations or use distinct per-function messages.
- Other gaps (deferrable): declaration-skipping isn't tested (add a `declare` and assert it's not visited); the adaptor isn't pinned (both tests would pass with a hand-rolled C++ loop bypassing the adaptor).

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -69,6 +69,28 @@ struct PyModulePass : llvm::PassInfoMixin<PyModulePass> {
...
+  explicit PyFunctionPass(nb::callable callback)
+      : callback(std::move(callback)) {}
+
+  llvm::PreservedAnalyses run(llvm::Function &f,
+                              llvm::FunctionAnalysisManager &) {
+    nb::gil_scoped_acquire gil;
+    nb::object res = callback(nb::cast(&f, nb::rv_policy::reference));
+    int truthy = PyObject_IsTrue(res.ptr());
+    if (truthy < 0)
+      throw nb::python_error();
```
makslevental: needs a message

projects/eudsl-llvmpy/src/IR/Passes.cpp
```diff
@@ -140,4 +162,20 @@ void populate_passes(nb::module_ &m) {
       "Run a Python callable as a module pass over the module in place. The "
       "callable receives the Module; return a truthy value if it mutated the "
       "IR (so analyses are invalidated), None/falsy otherwise.");
+
+  m.def(
+      "run_python_function_pass",
+      [](eudsl::Module &mod, nb::callable callback) {
+        PassPipelineEnv env(mod.get().getContext(),
+                            llvm::PipelineTuningOptions(), /*debug=*/false,
+                            /*verifyEach=*/false);
```
makslevental: these shouldn't be hardcoded (we have bindings for `PipelineTuningOptions` so it can be passed in as well)

makslevental (commented):

projects/eudsl-llvmpy/src/IR/Passes.cpp
makslevental: Done in 312931445 — `PyFunctionPass::run` now raises via `nb::raise_from(err, PyExc_ValueError, ...)` with "could not evaluate the truthiness of a Python function pass's return value", chaining the original error as `__cause__` (mirrors PyModulePass). `test_function_pass_unboolable_return_propagates` asserts the message and the chained cause.

projects/eudsl-llvmpy/src/IR/Passes.cpp
makslevental: Done in 312931445 — `run_python_pass_on_function` now takes `tuning`/`debug`/`verify_each` (defaulting to `None`/`False`), mirroring the other entry points, instead of hardcoding `PipelineTuningOptions()`. Covered by `test_function_pass_forwards_tuning_and_flags`.

makslevental:
## Follow-up on the review — remaining items addressed (312931445)

**Inline comments — fixed:**
- `PyFunctionPass::run` now raises the truthiness failure via `nb::raise_from(err, PyExc_ValueError, ...)` with a descriptive message, chaining the original as `__cause__` (mirrors the module pass). Asserted in `test_function_pass_unboolable_return_propagates`.
- `run_python_pass_on_function` no longer hardcodes the pipeline options — it takes `tuning`/`debug`/`verify_each`, mirroring the other entry points. Covered by `test_function_pass_forwards_tuning_and_flags`.

**Consolidated review items:**
- **first-error-wins not asserted** — resolved: `test_exception_in_function_pass_propagates` uses distinct per-function messages (`raise ValueError("boom in " + fn.name)`) and asserts `match="boom in f"` plus `assert seen == ["f"]`, so the early-skip is now load-bearing (it would fail if the 2nd function re-entered).
- **declaration-skipping not tested** — closed: added `test_function_pass_skips_declarations`, which parses a module with a bare `declare` plus one `define` and asserts the callback sees only `["f"]`. This also exercises the adaptor's "defined functions only" behavior.
- **adaptor isn't pinned** — partially addressed by the decl-skipping test (a hand-rolled loop over all functions would visit the declaration and fail it). Fully distinguishing the specific adaptor from any correct per-defined-function driver isn't observable through the bindings, so this is the best available proxy.

**Verification** (at this branch's own state): C++ coverage 100% (Passes.cpp 144/144 lines, 12/12 functions; Values.cpp 339/339, 60/60); `tests/passmanager/test_python_passes.py` 14 passed.

PR #605, opened 2026-08-23
[eudsl-llvmpy] Register Python passes by name for textual pipelines

makslevental:
register_python_pass(name, callback) records a callable in a static
registry (mirroring the caster registry, cleared at atexit). PassPipelineEnv
now takes the Module and installs a registerPipelineParsingCallback that
resolves a registered name to a PyModulePass, so a registered pass can be
named inside a run_passes pipeline (e.g. "my-pass,instcombine"). The
callback captures the current module and fires only during
parsePassPipeline, so it stays inert for the default-pipeline paths.

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -315,10 +315,15 @@ void populate_passes(nb::module_ &m) {
       "pipeline. on=PassKind.MODULE (default) names a module pass directly "
       "(e.g. \"my-pass,instcombine\"); on=PassKind.FUNCTION names a function "
       "pass, invoked inside a function(...) pipeline (e.g. "
-      "\"function(my-pass)\"). Pick a name that does not collide with a "
-      "builtin pass. The callable follows the run_python_pass_on_module / "
+      "\"function(my-pass)\"). Registering the same name again replaces the "
+      "callable. Pick a name that does not collide with a builtin pass. The "
+      "callable follows the run_python_pass_on_module / "
       "run_python_pass_on_function contract (truthy return marks the IR "
-      "mutated).");
+      "mutated). Caveats: the registry is not synchronized, so register from a "
+      "single thread (e.g. at import) before running pipelines concurrently; "
+      "and a callable that closes over a Module or Context keeps it alive "
+      "until "
+      "interpreter exit, so prefer callables that take the IR as an argument.");
 
   // Drop the registered callables while the interpreter is still up, so their
   // decref does not run after Python teardown (mirrors the caster registry).
--- a/projects/eudsl-llvmpy/src/IR/Passes.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Passes.cpp
@@ -12,6 +12,8 @@
 #include <llvm/Passes/StandardInstrumentations.h>
 
 #include <exception>
+#include <string>
+#include <unordered_map>
 
 namespace {
 // A Python exception raised inside a pass callback is stashed here instead of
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## Automated review (consolidated) — named registration: lifetime/thread-safety, tests, coverage

Covers `register_python_pass` and the pipeline-parsing callbacks. Updated for the current API (a bound `PassKind` enum, module + function registries) and the fixes made since the first pass.

**Current shape:** `register_python_pass(name, callback, on=PassKind.MODULE)` records the callable in one of two static registries keyed by kind (`PassKind` enum, not a string). `PassPipelineEnv` installs a module-level parsing callback (name used directly, e.g. `"my-pass,instcombine"`) and a function-level one (name used inside `function(...)`, e.g. `"function(my-pass)"`); both cleared at `atexit`.

**Implementation — sound.**
- The parsing callback's `modPtr` capture is safe (it's the `run_passes` reference arg, valid for the whole call; the callback fires only during `parsePassPipeline` in the same scope). `atexit` clear is correct (decref while interpreter alive). A registered name that shadows a builtin is simply unreachable (builtins resolve first) — the docstring says to avoid that.
- ⚠️ **Free-threaded data race (note):** the registries are unsynchronized `std::unordered_map`s; concurrent `register_python_pass` (write) + `run_passes` (read) races under the `FREE_THREADED` build. Mirrors the accepted `casterMap()` pattern; typical usage is register-at-import-then-run, so low severity — flagging.
- ⚠️ **Retained-callable caveat (note):** a registered callable that *closes over* a `Module`/`Context` stays alive until `atexit` (would surface as an `assert_no_leaks` failure). The tests avoid this; worth a docstring caveat.

**Tests / coverage:**
- `test_registered_pass_runs_by_name` proves the textual-pipeline path (not a direct call — an empty registry would make `parsePassPipeline` raise).
- `test_registered_pass_composes_with_builtin` proves both the Python pass and `instcombine` ran in one pipeline (`add %x, 0` is a reliable instcombine signal). It does **not** prove order, despite the comment — the two effects are independent.
- `test_registered_function_pass_runs_by_name` exercises `on=PassKind.FUNCTION` via `function(...)`; `test_register_python_pass_rejects_non_passkind` pins that `on` is the typed enum.
- ✅ **Added `test_raising_pass_skips_rest_of_pipeline`:** a raising registered pass before `instcombine` propagates the error out of `run_passes` **and** leaves the add-zero unfolded — validating both the error path through `runPipeline` and the ShouldRun pipeline-skip (fix in #602).
- Still open (deferrable): overwrite/last-write-wins semantics; two Python passes in one pipeline; truthy/falsy return via the named path. Residual note: the registry accumulates names globally for the rest of the suite (cleared only at `atexit`) — harmless now, but cross-test global state.

makslevental:
## Follow-up on the consolidated review — remaining items addressed (8b131f4c4)

No inline comments; working through the [consolidated review](https://github.com/llvm/eudsl/pull/605#pullrequestreview-5003360986).

**Test-honesty fix:**
- `test_registered_pass_composes_with_builtin` claimed "in order" but only proves *composition* (the rename and add-zero fold are independent effects). Reworded the comment to say exactly that, and added `test_registered_passes_run_in_pipeline_order`, which records each pass's name as it runs and asserts the textual order `["p1", "p2"]` — that one genuinely pins ordering.

**Deferrable test gaps — closed:**
- **overwrite / last-write-wins:** `test_register_python_pass_last_write_wins` registers the same name twice and asserts only the second callable runs.
- **two Python passes in one pipeline:** covered both by the new order test (happy path, both run) and the existing `test_raising_module_pass_skips_later_python_pass` (error path).
- **truthy/falsy via the named path:** `test_registered_pass_truthy_return_via_named_path` drives a truthy return through `run_passes` and asserts the mutation persists.

**⚠️ Notes — now documented in the `register_python_pass` docstring:**
- The registries are unsynchronized (mirrors the accepted `casterMap()` pattern), so the docstring says to register from a single thread (e.g. at import) before running pipelines concurrently.
- A callable that closes over a `Module`/`Context` keeps it alive until `atexit`; the docstring recommends callables that take the IR as an argument. Also documented last-write-wins on duplicate names.

**Not changed (accepted residuals):** the free-threaded race itself (design choice mirroring the caster registry; low severity for register-then-run usage) and the registry accumulating names globally for the rest of the suite (cleared only at `atexit`) — harmless, and the new tests use unique names to avoid cross-test interference.

**Verification** (at this branch's own state): C++ coverage 100% (Passes.cpp 202/202 lines, 18/18 functions; Values.cpp 339/339, 60/60); `tests/passmanager/` suite passed.

PR #606, opened 2026-08-23
[eudsl-llvmpy] Document Python-driven IR passes in the README

makslevental:
Add a "Python-driven passes" subsection covering run_python_pass,
run_python_function_pass, and register_python_pass: the truthy/falsy
changed-vs-unchanged return convention, exception propagation, and the
GIL/synchronous-execution note. Also list the new entry points in the
package-layout bullet for llvm.passmanager.

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -34,7 +34,9 @@ package top level:
 - `llvm.types` — the `Type` factories (`i1`…`i64`, `f16`/`f32`/`f64`, `ptr`,
   `void`, `function`, `struct`, `array`, `vector`) and `TypeID`.
 - `llvm.instructions` — free-function instruction emitters.
-- `llvm.passmanager` — `run_passes`, `run_default_pipeline`.
+- `llvm.passmanager` — `run_passes`, `run_default_pipeline`, and Python-driven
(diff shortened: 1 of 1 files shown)


makslevental (commented):
## Automated review — doc accuracy (updated for the current README)

**Verdict: accurate.** All claims match the implementation and tests; no dead links; the example is runnable. Refreshed for the current API after the renames + `PassKind`.

- Documents `run_python_pass_on_module` / `run_python_pass_on_function` (renamed one-shots) and `register_python_pass(..., on=PassKind.MODULE|FUNCTION)` — names spelled exactly as bound; the `function(...)` composition path is described.
- The truthy/None changed-vs-unchanged convention, "runs once per defined function", the builtin-name-collision caveat, exception propagation, and the GIL/synchronous note all match the code.
- No reference to the removed standalone doc (no dead links/anchors).
- ✅ The earlier nit — the "(MIR/machine passes are a separate follow-on)" forward-reference — has been removed; the section now describes only current scope ("these passes operate on LLVM IR at module and function level").
- Not overclaiming re #603: the README makes only the narrow, test-backed propagation claim, not a portability guarantee about unwinding through LLVM's `-fno-exceptions` frames.

PR #608, opened 2026-08-23
[eudsl-llvmpy] Bind object-level IR mutation/inspection APIs

makslevental:
Spike that came out of trying to write real transform passes in Python (a vectorizer and an inliner, in the two PRs stacked on this one). Those passes need a handful of object-level methods that already exist in LLVM C++ but weren't bound; this PR binds them (each a small passthrough):

- **Instruction:** `opcode`, `opcode_name` (distinguish add/mul/call/ret/…), `clone`, `erase_from_parent`, `remove_from_parent`, `move_before`, `insert_before`, `insert_after`.
- **User:** `set_operand`.
- **BasicBlock:** `split_basic_block`, `erase_from_parent`, `remove_from_parent`.

Notes:
- `insert_before`/`insert_after`/`move_before` use the iterator overloads (the `Instruction*` ones are deprecated in this LLVM).
- `clone` returns an unattached instruction (`rv_policy::reference`); the caller must insert it. `erase_from_parent` leaves the Python handle dangling (documented) — same contract as the freed C++ pointer.
- **`getContext` intentionally not bound:** the passes create types from existing `Value`s and use a no-context `IRBuilder` / the contextual current context (`currentOr`, as `BasicBlock.create` already does), so a per-value `getContext` is unnecessary here — and it would need a separate decision on reconciling a non-owning `LLVMContext&` with the leak-tracked `Context` wrapper.

Tested in `tests/ir/test_object_mutation.py` (opcode discrimination, clone+insert, move, erase, remove+reinsert, set_operand rewire, split_basic_block, block erase).

### DIFF
--- a/projects/eudsl-llvmpy/src/IR/Values.cpp
+++ b/projects/eudsl-llvmpy/src/IR/Values.cpp
@@ -9,6 +9,7 @@
 #include <llvm/IR/Argument.h>
 #include <llvm/IR/Attributes.h>
 #include <llvm/IR/BasicBlock.h>
+#include <llvm/IR/CFG.h>
 #include <llvm/IR/Constant.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DerivedTypes.h>
@@ -135,6 +136,7 @@ void populate_values(nb::module_ &m) {
       .def_prop_ro("num_operands", &llvm::User::getNumOperands)
       .def("operand", &llvm::User::getOperand, "index"_a,
            nb::rv_policy::reference_internal)
+      .def("set_operand", &llvm::User::setOperand, "index"_a, "value"_a)
       .def_prop_ro(
           "operands",
           [](llvm::User &self) {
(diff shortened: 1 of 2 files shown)


makslevental:
Expanded the toolkit (per review) so arbitrary Python passes have what they need at the object level, all 1:1 passthroughs to existing LLVM C++:

- **Instruction:** `comes_before` (program order), and the effect predicates `may_have_side_effects`, `may_read_or_write_memory`, `may_read_from_memory`, `may_write_to_memory` — on top of `opcode`/`opcode_name`, `clone`, `erase_from_parent`/`remove_from_parent`, `move_before`/`insert_before`/`insert_after`.
- **BasicBlock:** `predecessors`, `successors` (CFG navigation).
- **User:** `set_operand`.

Tests extended accordingly (`comes_before` ordering, the memory/side-effect predicates on load/add/store, and CFG preds/succs on a branchy function). `#609`'s vectorizer now uses `comes_before` + the purity predicates to perform a real legality check.

makslevental (commented):
## Automated review — bindings correctness, test honesty, coverage

### Implementation (code-reviewer): sound, no ≥80-confidence issues
- **`clone` with `rv_policy::reference` is the correct (safer) choice.** Inserted → the BB's ilist owns it; never inserted → a C++ leak but never a double-free (`take_ownership` would double-free after insertion). Docstring caveat is accurate.
- **`predecessors`/`successors` lifetime is correct** — `reference_internal` ties the returned list to `self` (kept alive by its function, which owns the blocks); valid on a container return.
- **`insert_before`/`insert_after`/`move_before` use the iterator overloads** — confirmed the `Instruction*` overloads are `LLVM_DEPRECATED` and these avoid them.
- CLAUDE.md compliant (snake_case mirrors of LLVM names, specific nb types, brace style); `splitBasicBlock` has no `Before` bool in this LLVM (separate `splitBasicBlockBefore`), so nothing is hardcoded away.

Low-severity notes:
- **`erase_from_parent` dangling handle**: documented, but nanobind keeps the freed pointer in its instance registry until GC — if LLVM reuses that address for a new Instruction/BasicBlock meanwhile, a *stale* wrapper (possibly wrong subtype) could be handed back. Inherent to binding LLVM's model; mitigated by the "don't touch after erase" contract.
- **`set_operand` is raw `&User::setOperand`**: an out-of-range index is `assert` in debug / OOB write (UB) in release, and there's no type check. Consistent with the already-bound `operand`/`getOperand`, but a one-line docstring noting the index precondition would match the care taken elsewhere.

### Tests (pr-test-analyzer): mostly honest & non-tautological; three real gaps
Confirmed each mutator's test fails on a no-op binding (clone→length delta, erase→shrink, insert_after vs before→index, move_before→reorder, split→2 blocks+`br`+moved insts, comes_before→both directions, CFG preds/succs→real edges). But:

1. **`opcode` (int) test passes for the wrong reason** — it only asserts `isinstance(int)` + `add.opcode != mul.opcode`; the comment claims "agrees with the name" but nothing checks that. A binding returning a *distinct-but-wrong* integer (even the address/hash) would still pass. Pin the `opcode`↔`opcode_name` agreement, or that two `add`s share an opcode while `add`/`mul` differ.
2. **`BasicBlock.remove_from_parent` is bound but has no test** — the only `remove_from_parent` test operates on an *Instruction*. (And when added, it must assert re-attachment/deletion itself — a detached-but-undeleted block is invisible to the leak gate.)
3. **`may_read_from_memory` has no False case** — only asserted True on `load`; a binding that always returns True (or is mis-wired to `mayReadOrWriteMemory`) would pass. Add `assert not add.may_read_from_memory`.

Minor: `clone` doesn't assert operands were copied (`dup.operand(0) is add.operand(0)`); `set_operand` has no bounds/type negative test (arguably can't without a binding-side guard); the `remove_from_parent` (Instruction) test proves detach but only *implies* non-deletion via a successful re-insert.

**Leak-gate note:** `assert_no_leaks()` counts only Context/Module — leaked Instruction/BasicBlock objects are invisible to it, so correctness of the erase/clone/remove tests rests entirely on the `len(...)` assertions (the tests do avoid real orphaning today).

makslevental:
## Follow-up on the review — remaining items addressed (94c151e63)

No inline comments; working through the [consolidated review](https://github.com/llvm/eudsl/pull/608#pullrequestreview-5003579944).

### The `erase_from_parent` dangling handle — **now fixed** (it was solvable)
The concern: after `inst.erase_from_parent()`, LLVM frees the C++ object, but nanobind keeps the wrapper (with its now-dangling pointer) in its instance registry, so a later attribute access is a use-after-free (and if LLVM reused the address, a wrong-subtype wrapper).

nanobind exposes the exact primitive to close this: `nb::inst_set_state(self, ready=false, destruct=false)`. Both `erase_from_parent` bindings (Instruction and BasicBlock) now free the object and then poison the wrapper. Any later use hits nanobind's cast guard and raises `TypeError` ("attempted to access an uninitialized instance") instead of dereferencing freed memory. `destruct=false` because the ilist owned the object and it is already deleted. Tests assert the poisoning.

**Honest caveat (why "fixed" not "eliminated"):** this poisons only *the* handle you called erase on. It does not defend against aliasing (two wrappers for the same instruction) or against `remove_from_parent`+leak. But the common footgun — erase then accidentally touch the same handle — goes from UB to a clean exception, which is the real win.

### Test gaps — closed
1. **`opcode` passed for the wrong reason** (only `isinstance(int)` + `add != mul`; a distinct-but-wrong integer would pass). Now pins `opcode`↔`opcode_name` agreement: a clone shares the integer (`add2.opcode == add.opcode`) while `add`/`mul` differ.
2. **`may_read_from_memory` had no False case.** Added `assert not add.may_read_from_memory` (pure add) and `assert not store.may_read_from_memory` (store writes but does not read) — rules out a binding hard-wired to True or mis-wired to `mayReadOrWriteMemory`.
3. **`BasicBlock.remove_from_parent` "bound but untested"** — it is in fact *not* bound (only `Instruction.remove_from_parent` is; the BB variant was dropped earlier as untestable/uncovered). The review note was stale, but the **test-module docstring still falsely listed it** — fixed the docstring.

### Minor
- `set_operand`: added a docstring noting the `index < num_operands` precondition (out of range is UB, matching LLVM's `setOperand`). A binding-side bounds check would change the mirror-LLVM contract, so left as documentation.

**Verification** (at this branch's own state): C++ coverage 100% (Values.cpp 437/437 lines, 77/77 functions; Passes.cpp 202/202, 18/18); `tests/ir/test_object_mutation.py` 12 passed. Also reran the downstack SLP-vectorizer + inliner example passes (which call `erase_from_parent`) — 18 passed, confirming they never touch a handle after erasing it.

PR #609, opened 2026-08-23
[eudsl-llvmpy] Minimal SLP vectorizer example (function pass)

makslevental:
A `run_python_pass_on_function` example that packs two independent scalar `add`s into one `<2 x T>` vector `add`, exercising the object-level bindings from #608 with no LLVM analyses (the independence check is done in Python):

- finds the adds via `opcode_name`,
- builds the operand vectors with `insert_element` (element type taken from the scalar op's `type`),
- replaces the scalar results with `extract_element`s via `replace_all_uses_with`,
- erases the now-dead scalar adds via `erase_from_parent`.

Tests assert the emitted `add <2 x i32>`, that both scalar `add i32`s are gone, and that the module still verifies.

Stacked on #608.

### DIFF
--- a/projects/eudsl-llvmpy/tests/passmanager/test_example_slp_vectorizer.py
+++ b/projects/eudsl-llvmpy/tests/passmanager/test_example_slp_vectorizer.py
@@ -8,12 +8,13 @@ single vector `add`, rewires the scalar uses to extractelements, and erases the
 dead scalar adds. No LLVM analyses are used -- the (trivial) legality check is
 done in Python."""
 
+import ctypes
 from textwrap import dedent
 
 import pytest
 
 import llvm
-from llvm import ir, types
+from llvm import ir, jit, types
 from llvm import instructions as I
 from llvm.testing import assert_no_leaks
 
@@ -60,6 +61,11 @@ def slp_vectorize(fn):
     # Legality (checked, not assumed): same element type; both pure (no memory
     # traffic or other side effects); and mutually independent -- neither
     # transitively feeds the other -- so a single vector op can replace both.
+    # The independence walk is the load-bearing check here and is exercised in
+    # both the direct and transitive tests. The purity guard is a general-
+    # legality illustration: candidates are integer `add`s, which are always
+    # pure, so for this candidate set it never fires -- a real SLP pass over a
+    # wider opcode set (e.g. loads/calls) would need it.
     if x.type is not y.type:
         return False
     if any(i.may_read_or_write_memory or i.may_have_side_effects for i in (x, y)):
@@ -68,7 +74,9 @@ def slp_vectorize(fn):
         return False
 
     # Insert after whichever add comes later, so all four scalar operands
-    # dominate the new vector op.
+    # dominate the new vector op. `adds` is in program order, so x precedes y
+    # here and `last` is always y; comes_before is used to keep the rule correct
+    # if candidate selection ever stops being ordered.
     last = y if x.comes_before(y) else x
     vec = types.vector(x.type, 2)  # element type taken from the scalar op
     b = ir.IRBuilder()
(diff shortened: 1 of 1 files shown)


makslevental:
The legality check is now actually performed (no longer assumed): same element type, both operands pure (`may_read_or_write_memory`/`may_have_side_effects`), and mutual data-independence via an operand-chain walk; the insertion point is ordered with `comes_before`. Added a test that dependent adds (`%s1 = add %s0, %q`) are left un-vectorized.

makslevental (commented):
## Automated review — test honesty, coverage, correctness

The three tests each guard a distinct branch against *removal* (no-op pass → `combines` fails; removed `len<2` guard → `noop_single_add` errors; removed independence check → `skips_dependent_adds` fails because the self-referential IR fails `verify()`). But they verify **structure only** (string shape + `verify()`), never semantics — and two legality branches are effectively dead.

### ⚠️ Honesty gap — `combines_two_adds` passes on a semantically-wrong vectorization (most important)
The asserts (`"add <2 x i32>" in text`, `"add i32" not in text`, `count("extractelement")==2`, `verify()`) are satisfied by *any* well-formed vector add with two lane extracts — **regardless of which operand lands in which lane or which extract feeds which store**. Concretely, swapping the two `extract_element` indices (so the store to `%pa` gets `a1+b1` instead of `a0+b0`) still contains `add <2 x i32>`, no `add i32`, exactly two `extractelement`, and passes `verify()` — the test goes green on a wrong transform. Unlike the inliner PR (which JIT-executes), there's no execution check. **Meaningful missing coverage:** JIT-run `@f` on known inputs and assert the stored results, which would catch lane/operand/wiring bugs that string+verify can't.

### Coverage gaps
- **`skips_dependent_adds` only exercises the *direct* dependence.** With `%s1 = add %s0, %q`, `_reaches` returns on the first operand comparison — the `stack`/`seen`/`op.parent is block` walk never runs. A multi-hop case (`%t = add %s0,…; %s1 = add %t,…`) would actually exercise the transitive walk (and catch a broken push). The check's *presence* is guarded; its *transitive correctness* isn't.
- **The purity guard (`may_read_or_write_memory`/`may_have_side_effects`) is vacuous here** — candidates are filtered to `opcode_name == "add"`, and integer adds never have memory/side effects, so the guard can't fire for the only instructions it sees. No test can trip it without changing candidate selection. Worth knowing the "skips impure candidates" story isn't really covered.
- **The `comes_before` else-branch is unreachable** — `adds` is collected in program order, so `x` always precedes `y` and `x.comes_before(y)` is always True. The "insert after whichever comes later" logic is only half-validated.

### Good
- `noop_single_add` correctly guards the `len(adds) < 2` early return. The `"add i32" not in text` negative is robust (`add <2 x i32>` doesn't contain `add i32`). `verify()` does reject the use-before-def IR a removed independence check would produce, which is what makes that removal case fail loudly. Leak-gate usage is correct.

Net: the combines test proves *a* vectorization happened and fails on a no-op, but not that it's *correct*; a JIT/execute assertion is the highest-value addition, plus a transitive-dependence case.

makslevental:
## Follow-up on the review — remaining items addressed (3aaf52132)

No inline comments; working through the [consolidated review](https://github.com/llvm/eudsl/pull/609#pullrequestreview-5003584758).

### ⚠️ Honesty gap — combines test passed on a semantically-wrong vectorization — **fixed**
This was the most important finding: string shape + `verify()` are satisfied by any well-formed vector add with two extracts, so a lane/operand swap would still go green. Added `test_slp_vectorizer_preserves_semantics`, which JIT-executes `@f` on known inputs and asserts the **stored results**: `*pa == a0+b0` (13) and `*pc == a1+b1` (27). A swapped vectorization would store 27 into `*pa`, so the test now fails on wrong wiring — exactly what the review asked for.

### Coverage gaps
- **Transitive dependence not exercised** — closed: `test_slp_vectorizer_skips_transitively_dependent_adds` uses `%s0 -> %s1(mul) -> %s2`, so `adds[0]`/`adds[1]` (= `%s0`/`%s2`) depend only *through* the intervening mul. That forces `_reaches` to push `%s1` and walk past the direct operands — the stack/seen path the direct-only test never hit.
- **Vacuous purity guard** — not hidden, now documented honestly: candidates are integer `add`s, which are always pure, so the guard can't fire for this candidate set. The comment now says it is a general-legality illustration (a real SLP pass over loads/calls would need it) rather than implying it's exercised. I did not contrive a test that fakes coverage of an unreachable branch.
- **`comes_before` else-branch unreachable** — same treatment: `adds` is in program order so `x` always precedes `y` and `last` is always `y`; the comment now states this and explains `comes_before` is kept so the rule stays correct if candidate selection ever stops being ordered.

**Verification** (at this branch's own state): `tests/passmanager/test_example_slp_vectorizer.py` 5 passed (incl. the JIT semantics + transitive tests); C++ coverage 100% (Values.cpp 437/437, 77/77; Passes.cpp 202/202, 18/18).

PR #610, opened 2026-08-23
[eudsl-llvmpy] Minimal inliner example (module pass)

makslevental:
A `run_python_pass_on_module` example that inlines calls to defined, single-block callees, exercising the object-level bindings from #608 (no LLVM analyses):

- clones each callee instruction before the call (`clone`),
- remaps operands from callee args/values to the caller's via a value map (`set_operand`),
- rewires the call result to the returned value (`replace_all_uses_with`),
- removes the call (`erase_from_parent`).

Includes a JIT test confirming semantics are preserved (`caller(5) == (5+3)*5+100`), a structural test (call removed, body spliced with args remapped), and that external (declaration) calls are left alone.

### DIFF
--- a/projects/eudsl-llvmpy/tests/passmanager/test_example_inliner.py
+++ b/projects/eudsl-llvmpy/tests/passmanager/test_example_inliner.py
@@ -1,0 +1,119 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""A minimal inliner written as a Python module pass, exercising the
+object-level bindings end to end: for a call to a single-block, defined callee
+it clones each callee instruction into the caller before the call, remaps
+operands (callee args -> actual args, callee values -> their clones) with
+set_operand, rewires the call result to the callee's returned value, and erases
+the call. No LLVM analyses; the (trivial) inlinability check is done in Python."""
+
+import ctypes
+from textwrap import dedent
+
+import pytest
+
+import llvm
+from llvm import ir, jit
+from llvm.testing import assert_no_leaks
+
+_SRC = dedent("""\
+    define i32 @callee(i32 %a, i32 %b) {
+      %t = add i32 %a, %b
+      %u = mul i32 %t, %a
+      ret i32 %u
+    }
+    define i32 @caller(i32 %x) {
+      %c = call i32 @callee(i32 %x, i32 3)
+      %d = add i32 %c, 100
+      ret i32 %d
+    }
+    """)
+
+
+def _inline_single_block(call, callee):
+    # Map callee formals to the actual call arguments.
+    vmap = {callee.arg(i): call.arg_operand(i) for i in range(callee.num_args)}
+    ret_val = None
+    for ci in callee.entry_block.instructions:
+        if ci.opcode_name == "ret":
+            ret_val = (
+                vmap.get(ci.operand(0), ci.operand(0)) if ci.num_operands else None
+            )
+            break
+        clone = ci.clone()
+        # Remap operands that refer to callee-local values; leave constants /
+        # globals (not in the map) pointing at the shared objects.
+        for k in range(clone.num_operands):
+            op = clone.operand(k)
+            if op in vmap:
+                clone.set_operand(k, vmap[op])
+        clone.insert_before(call)
+        vmap[ci] = clone
+    if ret_val is not None:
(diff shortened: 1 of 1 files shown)


makslevental (commented):
## Automated review — test honesty, coverage, correctness

**Strong overall:** the structural + JIT pair is genuinely complementary and would catch the likely inlining bugs. No bug found in the pass.

**Why the JIT check is honest:** `caller(5) == (5+3)*5+100 == 140` uniquely pins the operand remap because `%a` appears in both the `add` and the (order-sensitive) `mul`, and `%a→x` vs `%b→3` are distinct — plausible bugs give distinct results:
- swapped arg order (`%a→3, %b→x`) → `(x+3)*3+100 = 124 ≠ 140` ✓
- `mul` using `%b`(3) instead of `%a`(x) → `124 ≠ 140` ✓
- clone/operand left pointing at a callee value → cross-function ref → `verify()` raises ✓

It also correctly exercises the trickiest path — a clone referencing a *prior* clone (`%u = mul %t, %a` via `vmap[ci]=clone` ordering) — and a constant actual arg (`i32 3`) flowing through `set_operand`. `verify()` is a real gate (raises), and the JIT round-trip is real e2e, not a mock. `add` is commutative so the JIT alone doesn't pin add-operand order, but the structural `"add i32 %x, 3"` assert does — **neither test is complete alone; together they're solid.**

### Findings
- **A (medium, honesty): `leaves_external_calls_alone` doesn't prove the `callee.is_declaration` guard it claims.** A declaration has 0 basic blocks, so the next guard `len(basic_blocks) != 1` already excludes it — removing `or callee.is_declaration` fails no test. (It *does* genuinely pin the caller-side `if fn.is_declaration: continue` — removing that makes `@ext.entry_block` None → `AttributeError`.) Either note the redundancy or drop the misleading comment.
- **B (medium, coverage): the "operand not in `vmap`" branch is never exercised** — every operand here (`%a`,`%b`,`%t`) is in the map, so the "leave constants/globals pointing at shared objects" path the docstring emphasizes is dead. Add a callee with an immediate constant operand (e.g. `%t = add i32 %a, 7`) to cover it.
- **C (low): the multi-block `len(basic_blocks) != 1` skip is unverified** — no multi-block callee; removing the guard would clone a multi-block callee into broken IR uncaught. Cheap to add (assert the call survives).
- **D/E/F (low, deferrable): void callee / unused result (`ret_val is None` path); multiple calls, non-entry-block calls, recursion (the pass only scans `entry_block` — fine as "minimal", worth a doc note); the `changed` return value is never asserted.**

Exact "what breaks each test": remove `insert_before` or the `set_operand` loop or `erase_from_parent` → structural fails; any verifying-but-wrong remap → JIT `≠ 140`; remove caller-side `is_declaration` guard → external test errors.

makslevental:
## Follow-up on the review — remaining items addressed (6ecffbcdd)

No inline comments; working through the [consolidated review](https://github.com/llvm/eudsl/pull/610#pullrequestreview-5003590562).

**A (honesty) — `leaves_external_calls_alone` didn't prove the callee-side `is_declaration` guard.** Fixed the comment: the test genuinely pins the *caller-side* `if fn.is_declaration: continue` (removing it makes iterating `@ext`'s entry block fail), while the callee-side `is_declaration` check is redundant with the single-block guard (a declaration has 0 blocks). The comment now says exactly that instead of implying it proves the callee-side guard.

**B (coverage) — "operand not in vmap" branch was dead.** Added `test_inliner_keeps_constant_operands_through_remap`: the callee uses an immediate (`add %a, 7`), so that operand isn't in vmap and must stay pointing at the shared constant. JIT confirms `5 + 7 == 12`.

**C (coverage) — multi-block skip unverified.** Added `test_inliner_skips_multi_block_callee`: a two-block callee is left alone (`changed is False`, the call survives).

**D (coverage) — void callee / `ret_val is None` path.** Added `test_inliner_void_callee_with_unused_result`: a `ret void` callee has no result to rewire; the call is spliced away and JIT confirms the inlined store ran.

**F — misc.** The splice test now asserts `changed is True`; the pass docstring notes it only scans entry blocks (recursion / non-entry calls / multiple passes are out of scope by design).

Not added: the review's "add-operand order isn't pinned by JIT alone" point is already covered — the structural `"add i32 %x, 3"` assert pins it, and the review itself concluded the structural + JIT pair is complete together.

**Verification** (at this branch's own state): `tests/passmanager/test_example_inliner.py` 6 passed (incl. the three new coverage tests, two JIT-backed); C++ coverage 100% (Values.cpp 437/437, 77/77; Passes.cpp 202/202, 18/18).

Issue #593, opened 2026-08-19
[TODO] MIR stack: deferred review items (#579–#591)


makslevental:
## MIR stack — deferred review items tracker

Across the MIR-stack review pass (#579–#591) a number of reviewer recommendations were **not applied** in the review commits — either deliberately declined (they conflict with a project convention) or deferred as larger, cross-cutting work better done on its own than folded into a per-PR review pass. This issue collects them in one place so the maintainer can decide which to pursue. Each item notes its source PR and the reasoning given at the time.

None of these is a known bug — they are design/ergonomics/enforcement follow-ups. (Blocking correctness findings from the reviews were all fixed in the respective PRs.)

---

### 1. ~~`MachineModuleInfo` ownership-struct invariant enforcement~~ — ✅ done in #619
~~**From #580 (item 4) and #590 (type-design).** The struct is a flat public aggregate now discriminated by three ownership modes (`pm` / `ownedMmi` / `mmiwp`) plus `tm` (borrowed) and `emitted` (one-shot phase flag), enforced by runtime null-checks + a load-bearing comment.~~

- ~~#580 suggested renaming it (`MirModule`) + a private ctor + named factories to enforce the invariants and avoid the `llvm::MachineModuleInfo` shadow.~~
- ~~#590 suggested a `std::variant` over per-path types + an explicit `Phase` enum (and an `EmittedObject` return type for the consume-once `emit_object`) so "emit_object only on the build path", the one-shot, and the post-emit two-owner state become compile-time facts instead of runtime throws.~~

~~**Why deferred:** the code was verified correct at every branch; the rename/private-ctor fights the aggregate-init the upstack PRs use, and the variant/Phase rewrite is a deliberate refactor. Best done as one dedicated change now that the stack has stopped growing the struct.~~

### 2. ~~Typed `Register` (and richer builder surface)~~ — ✅ done in #621
~~**From #582 (items 7, 8).**~~
- ~~Wrap `Register` with its owning `MachineFunction` so cross-function misuse is caught structurally even when a foreign vreg's index collides with a same-typed vreg in the target function (today the C++ `requireVRegOfType` misses that collision case; the DSL `MachineValue` anchor added in #583 covers the DSL path, but the raw binding does not).~~
- ~~Arbitrary `DstOp` / multi-block builder control (a fuller `MachineIRBuilder` surface).~~

[...]

PR #619, opened 2026-08-24
[eudsl-llvmpy] MirModule: variant-typed ownership + named factories (#593 item 1)

makslevental:
Implements [issue #593](https://github.com/llvm/eudsl/issues/593) item 1 (**both** recommendations) for the MIR wrapper in `src/MIR/Machine.cpp`.

## Before

A flat `MachineModuleInfo` aggregate held the MMI three different ways — `pm` (codegen path), `ownedMmi` (parse path), `mmiwp` (build path) — of which only one was ever non-null, plus a borrowed `tm` and a one-shot `bool emitted`. Which mode an instance was in was answered by runtime null-checks over three pointers, and `emit_object`'s "build-path only" / one-shot contract was a pair of runtime guards held together by a load-bearing comment. The name also shadowed `llvm::MachineModuleInfo`, which it is not.

## After

- **rename + private ctor + named factories.** The wrapper is now `class MirModule` (no longer shadowing the LLVM type). It has a **private constructor** and is built only through three named static factories — `MirModule::codegen(...)`, `MirModule::parsed(...)`, `MirModule::building(...)` — one per construction path, so callers cannot construct an inconsistent path/ownership pairing.
- **variant + type-state emit.** Ownership is a `std::variant<CodegenOwned, ParseOwned, BuildOwned, EmittedOwned>`; each alternative carries exactly what that path owns and a `queryHandle()`, and `mmi()` recovers the query handle via `std::visit`. `emit_object` does `std::get_if<BuildOwned>` and, on success, **consumes** the `BuildOwned` into an `EmittedOwned`. So "build-path only" and the one-shot are the variant's active alternative rather than runtime flags, and the old post-emit "both `mmiwp` and `pm` live" state is now unrepresentable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,8 @@
 .mcp.json
 .coverage
 *.zip
+
+# Local LLVM/MLIR dev install (see mlir_wheel; used for building eudsl-llvmpy locally)
+/mlir_wheel/
+/mlir_wheel-*.whl
+/mlir_wheel-*.dist-info/
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -254,7 +254,7 @@ and **lower** built MIR to native code.
 The object model is bound as its own hierarchy (MIR is not part of the `Value`
 hierarchy): `MachineFunction`, `MachineBasicBlock`, `MachineInstr`,
 `MachineOperand`, `Register`, plus `LLT` (the generic low-level type) and
-`MachineFunctionProperty`. `MachineModuleInfo` owns the `MachineFunction`s and
+`MachineFunctionProperty`. `MirModule` owns the `MachineFunction`s and
 keeps everything they reference alive.
 
 > The examples below use AArch64 opcodes/registers/triples, so they need the
@@ -266,7 +266,7 @@ keeps everything they reference alive.
 ### Inspecting MIR from the compiler
 
 `run_codegen_to_mir` runs instruction selection on an IR module and hands back the
-`MachineModuleInfo` owning the resulting `MachineFunction`s. Pass
+`MirModule` owning the resulting `MachineFunction`s. Pass
 `global_isel=True` for the GlobalISel pipeline instead of SelectionDAG.
 `to_mir()` prints the `.mir` textual form and `parse_mir` reads it back;
 `MachineFunction.verify()` runs the machine verifier.
--- a/projects/eudsl-llvmpy/src/IR/JIT.cpp
+++ b/projects/eudsl-llvmpy/src/IR/JIT.cpp
@@ -60,7 +60,7 @@ void populate_jit(nb::module_ &m) {
           },
           "obj"_a,
           "Add a relocatable object file (e.g. from "
-          "MachineModuleInfo.emit_object) to the JIT.")
+          "MirModule.emit_object) to the JIT.")
       .def(
           "lookup",
           [](llvm::orc::LLJIT &self, const std::string &name) {
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -365,6 +365,11 @@ public:
     if (tm->addPassesToEmitFile(
             *pm, os, nullptr, llvm::CodeGenFileType::ObjectFile,
             /*DisableVerify=*/true, build->mmiwp.release())) {
+      // The wrapper has already been released into `pm`, so the BuildOwned's
+      // mmiwp is now null. Consume it into EmittedOwned before throwing so `pm`
+      // (and thus `info`) stays alive and a later query/retry reports a clean
+      // error instead of dereferencing the released mmiwp.
+      state_ = EmittedOwned{std::move(pm), info};
       throw std::runtime_error("target cannot emit an object file");
     }
     // LCOV_EXCL_STOP
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -43,6 +43,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <variant>
 #include <vector>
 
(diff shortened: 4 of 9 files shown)


makslevental:
## Review — focus: test coverage, correctness, honesty

Reviewed with three focused passes (test-coverage/honesty, correctness, silent-failure). Overall this is a faithful, clearer re-expression of the ownership logic and the 100% coverage is **genuine, not gamed** — but there is one real latent regression to flag.

### Important — `emit_object` failure path drops the original's exception-safety guard (latent UAF)

`projects/eudsl-llvmpy/src/MIR/Machine.cpp:365-369`

```cpp
if (tm->addPassesToEmitFile(*pm, os, nullptr, llvm::CodeGenFileType::ObjectFile,
                            /*DisableVerify=*/true, build->mmiwp.release())) {
  throw std::runtime_error("target cannot emit an object file");
}
```

`build->mmiwp.release()` clears the `unique_ptr` **before** the return value is checked, and the `state_ = EmittedOwned{...}` transition (line 388) only runs on success. So on this `throw`, `state_` stays `BuildOwned` with a **null** `mmiwp`, while the **local** `pm` (line 361) unwinds and frees the wrapper it already adopted (`addPassesToGenerateCode` does `PM.add(&MMIWP)` before it can fail).

Consequence: any subsequent access — a retry `emit_object()` (line 313, `&build->mmiwp->getMMI()`), or `to_mir()` / `machine_functions` routing through `mmi()` → `BuildOwned::queryHandle()` (line 230, `&mmiwp->getMMI()`) — dereferences the null/freed `mmiwp` → **null-deref / use-after-free**, not a catchable Python error.

Regression vs. `main`: the original stored `mmi` as a **separate cached raw pointer** and kept `pm` as a **member**, so after the same failure `self.mmi` stayed valid, a retry hit `!self.mmiwp` and threw a clean catchable error, and `to_mir()` still worked. The variant swapped that null-pointer guard for a *type* check (`std::get_if<BuildOwned>`) that says nothing about whether the `BuildOwned`'s `mmiwp` is still populated.

Reachability / honesty note: this branch is `// LCOV_EXCL_START -- AArch64 can always emit an object file`, so it is **currently unreachable** with the only target the suite exercises — no CI or normal-AArch64 impact today, and the exclusion itself is legitimate per the coverage policy. But `create_machine_function` takes a **user-supplied** `target_machine`; a target/config where `addPassesToEmitFile` returns `true` would hit this, and the `LCOV_EXCL` marker means the gate can never surface the regression. Worth fixing before another target is added (e.g. transition `state_` to `EmittedOwned`/a poisoned alternative *before* throwing, so later calls report a clean error instead of touching the freed wrapper).

### Suggestions (non-blocking)

- **`test_mir_module_is_queryable_after_emit` — one assertion isn't load-bearing.** `projects/eudsl-llvmpy/tests/mir/test_emit_jit.py:95-107`. The test is a genuine behavioral test (not a coverage touch): `assert list(mmi.machine_functions) == []` dereferences the pointer `EmittedOwned::queryHandle()` returns and verifies the post-emit MMI is still live (MFs freed by `FreeMachineFunctionPass`) — exactly the invariant `EmittedOwned{pm, mmi}` exists to preserve. But `assert "@add" in mmi.to_mir()` comes from `printMIR(os, *module_)` (the IR text), independent of the query handle, so it would still pass even if `queryHandle()` returned a wrong-but-non-null pointer. Consider asserting `mmi.machine_function("add")` now raises `KeyError` — it routes through the same handle and directly documents the freed-MF semantics.
- **Parse-path emit rejection not explicitly asserted.** `test_emit_object_requires_create_machine_function` (test_emit_jit.py:110) reaches the `!BuildOwned` branch only via a `CodegenOwned` module; a `ParseOwned` module hits the identical branch (so it's not a line gap), but the parse path's rejection isn't documented. Low value.

### Strengths / honesty verdict

- **Coverage is genuine.** All three construction paths + each alternative's `queryHandle()`, and `emit_object`'s three outcomes (success / "already emitted" / wrong-path), are exercised with **specific** assertions (`match="already emitted"`, `"create_machine_function"`, `"failed verification"`; ELF magic + `add\x00` + JIT-executed `add(2,3)==5`).
- **`LCOV_EXCL` markers are honest** — all guard truly-unreachable or defensive lines (unregistered `-start-after`; the AArch64 emit-failure `throw` body while the call itself is covered; post-`pm->run` diag/empty checks made unreachable by the eager `mf->verify()`), and are carried over from the original, not newly added to hide the refactor.
- The two-emit test verifies the actual **state transition** (BuildOwned consumed into EmittedOwned), not merely a re-thrown flag. Moving `pm` from a member to a local also genuinely improves exception-safety on the *early* (verify/registration) throws — the one gap is the `release()`-then-throw window above.
- Construction (`new MirModule(MirModule::codegen(...))` with private ctor + move-only members) and nanobind ownership are correct; the happy path and both user-facing error messages are behaviorally equivalent to `main`.

makslevental:
Addressed the review in `08401a2`:

- **emit_object UAF (Important):** on a failed `addPassesToEmitFile`, the wrapper has already been `release()`d into the local `pm`, so I now transition `state_ = EmittedOwned{std::move(pm), info}` **before** the throw. `pm` (and the captured `info`) stays alive, and a later query/retry reports a clean `"already emitted"` error instead of dereferencing the released (null) `mmiwp` / touching the freed wrapper. (Still `LCOV_EXCL` — AArch64 always emits — but no longer a crash on the user-target-reachable path.)
- **Strengthened `test_mir_module_is_queryable_after_emit`:** now also asserts `machine_function("add")` raises `has no MachineFunction` post-emit, which actually routes through the post-emit MMI handle (the `"@add" in to_mir()` check only reads the IR text).
- **Added `test_emit_object_rejects_parse_path`:** documents that a `parse_mir` (`ParseOwned`) result is rejected with the same message as the codegen path.

Verified locally (built against the LLVM dev install, system `llvm-profdata`/`llvm-cov`): full suite **647 passed / 3 xfailed**, and `src/IR` + `src/MIR` remain at **100% line + function coverage**.

PR #621, opened 2026-08-25
[eudsl-llvmpy] Typed Register (owning function) + generic build; drop DSL anchor (#593 item 2)

makslevental:
Implements both recommendations from [#593 item 2](https://github.com/llvm/eudsl/issues/593) (deferred from #582): a typed `Register` that knows its `MachineFunction`, and a fuller `MachineIRBuilder` surface — then refactors the DSL, which previously handled cross-function provenance in Python.

## Part A — typed `Register`

`Register` is bound from a `TypedRegister` value type that pairs the register with the `MachineFunction` that minted it (a physreg carries no owner, since physregs are target-static). The **invariant is enforced by construction**: the constructor is private and a single `owned(mf, reg)` factory *derives* the owner from `reg.isVirtual()`, so a vreg-with-no-owner or physreg-with-owner is unrepresentable (not merely unchecked).

A single shared guard, `requireOwnedVReg(mf, reg, role)`, rejects any vreg whose owner isn't the consuming builder's function, and is called from **every** consuming site (`build_add/sub/mul/copy/icmp/brcond/phi`, `add_reg/def/use`, `add_phi_incoming`, and the new `build`).

This closes the **index-collision hole** the old `getType`-only check missed: a vreg's numeric id is only an index into its own function's `MachineRegisterInfo`, so when `f` and `g` each mint `%0:s32`, feeding `g`'s `%0` into `f`'s builder used to pass (it found `f`'s own same-typed `%0`). It now raises. `MachineOperand.reg` tags the register with its owning function so it round-trips back into the builder.

Register **equality and hashing consider both id and owner** (`__eq__`/`__ne__` compare `(id, owner)`, `__hash__` combines them), so two same-id registers from different functions are distinct — they can safely coexist as `set`/`dict` keys, and a cross-function `==` is `False`.

## Part B — richer builder surface

Adds `MachineIRBuilder.build(opcode, dsts, srcs)`, the typed analogue of `buildInstr(Opc, ArrayRef<DstOp>, ArrayRef<SrcOp>)`: each dst is an `LLT` (mints a fresh generic vreg and defines it) or a `Register` (defines that existing vreg), each src is a `Register` use. Unlike the fixed single-def helpers it expresses multiple defs and writing into a caller-provided vreg; it returns the `MachineInstr` so minted defs are readable. The "Building target MIR" README example now uses `build(...)` for the COPY/ADDWrr instructions (keeping `build_instr` + `add_reg` only for `RET_ReallyLR`'s implicit operand, which `build`'s plain-use srcs can't express).

## DSL refactor

Removes the Python-side `_builder`/`_require_current` anchor from `MachineValue` (added in #583). The typed register now enforces cross-function provenance structurally in C++, so the anchor is redundant. Observable change: a value used across `@machine_function` bodies now raises `ValueError` ("belongs to a different MachineFunction") from the builder rather than a Python `RuntimeError`. A value built under one builder is still usable under a second builder of the *same* function (the owner check keys on the function, not the builder instance).

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -255,7 +255,10 @@ The object model is bound as its own hierarchy (MIR is not part of the `Value`
 hierarchy): `MachineFunction`, `MachineBasicBlock`, `MachineInstr`,
 `MachineOperand`, `Register`, plus `LLT` (the generic low-level type) and
 `MachineFunctionProperty`. `MirModule` owns the `MachineFunction`s and
-keeps everything they reference alive.
+keeps everything they reference alive. A `Register` carries the
+`MachineFunction` that minted it (a physical register carries none), so a vreg
+passed into a *different* function's builder is rejected — its numeric id would
+otherwise silently alias a same-typed register in that function.
 
 > The examples below use AArch64 opcodes/registers/triples, so they need the
 > AArch64 backend linked (`EUDSL_LLVMPY_TARGETS`); on other hosts
@@ -336,6 +339,9 @@ already-selected **target** MIR directly and run only the back half of codegen.
 `reg_class`/`physreg` resolve register classes and physical registers by name,
 `create_vreg` makes class-constrained vregs, `add_reg` appends operands with the
 full flag set (def/use, implicit, kill, …), and `set_property` marks the function.
+(`build(opcode, dsts, srcs)` is a typed one-shot alternative to
+`build_instr` + `add_reg`: each dst is an `LLT` to mint a fresh vreg for or a
+`Register` to define, each src is a `Register` use.)
 `emit_object()` then runs register allocation and emission (via
 `-start-after=finalize-isel`, so no instruction selection), and `LLJIT.add_object`
 loads the result:
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -361,9 +361,11 @@ with ir.Context() as ctx:
     mf.blocks[0].add_livein(w0); mf.blocks[0].add_livein(w1)
     v0, v1, v2 = (mf.create_vreg(gpr32) for _ in range(3))
     for dst, src in ((v0, w0), (v1, w1)):
-        c = b.build_instr(mf.opcode("COPY")); c.add_reg(dst, is_def=True); c.add_reg(src)
-    a = b.build_instr(mf.opcode("ADDWrr")); a.add_reg(v2, is_def=True); a.add_reg(v0); a.add_reg(v1)
-    c = b.build_instr(mf.opcode("COPY")); c.add_reg(w0, is_def=True); c.add_reg(v2)
+        b.build(mf.opcode("COPY"), [dst], [src])   # dst = COPY src
+    b.build(mf.opcode("ADDWrr"), [v2], [v0, v1])   # v2 = ADDWrr v0, v1
+    b.build(mf.opcode("COPY"), [w0], [v2])         # $w0 = COPY v2
+    # RET_ReallyLR takes an *implicit* use of $w0, which build's plain-use srcs
+    # can't express, so drop to build_instr + add_reg for that one operand.
     r = b.build_instr(mf.opcode("RET_ReallyLR")); r.add_reg(w0, implicit=True)
     for p in ("IsSSA", "TracksLiveness", "NoPHIs"):
         mf.set_property(getattr(mir.MachineFunctionProperty, p))
--- a/projects/eudsl-llvmpy/CLAUDE.md
+++ b/projects/eudsl-llvmpy/CLAUDE.md
@@ -4,6 +4,16 @@ Guidance for this project: the hand-written nanobind bindings over the LLVM
 C++ API, plus the DSL frontend built on them. These apply to every PR in the
 stack, so check new code against them before pushing.
 
+## Breaking changes are fine
+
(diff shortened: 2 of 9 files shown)


makslevental (commented):
## Review — PR review toolkit (code / tests / comments / silent-failure / type-design)

**No critical or merge-blocking issues.** Cross-checked findings from five focused passes: the `mf` pointer is only ever *compared*, never dereferenced (no UAF); every register-consuming binding routes through `requireOwnedVReg` (no unguarded site); the collision fix is real and behaviorally tested; dropping the DSL anchor introduced no silent failure (missing-builder still raises; cross-function still rejected). CLAUDE.md-compliant, 100% C++ coverage gate satisfied.

### Important

1. **Comment overclaims what the verifier catches** — `src/MIR/Machine.cpp:96-97`. The `TypedRegister` comment says a foreign vreg "builds corrupt MIR that the verifier — gone under NDEBUG — would otherwise catch." For the *collision* case (the very case this type exists to fix), the MIR is well-formed in the target function, so the verifier would **not** catch it in debug or release — contradicting the comment's own next sentence. Reword so the verifier is credited only with the detectable (out-of-bounds / wrong-type) cases, and make explicit that the same-id/same-type collision is invisible even to the verifier, which is *why* the structural owner check exists.

2. **`TypedRegister` invariant is convention-only** — `src/MIR/Machine.cpp:105-108`. It's a bare public aggregate constructed inline at ~10 sites; nothing enforces "vreg ⇒ non-null `mf`, physreg ⇒ null `mf`." Item 1 (#619) in this same issue applied private-ctor + named factories to the sibling `MirModule` struct in this same file for exactly this reason, so landing this as a raw aggregate is internally inconsistent. Recommend a private ctor + `vreg(mf, r)` / `phys(r)` factories (no variant needed); give the `MachineOperand.reg` accessor an explicit branch so the "virtual-but-no-parent" case is a conscious decision, not a silent fall-through.

3. **Test gaps on genuinely new paths.** Line coverage is 100%, but because `requireOwnedVReg` is shared, several call sites' rejection behavior and the new `build` paths are not exercised:
   - **Round-trip (highest value):** no test reads a register off `MachineOperand.reg` and feeds it back into a builder. This is the *only* direction where a wrong owner would wrongly **reject** a legitimate register (`src/MIR/Machine.cpp:619-632`).
   - **`build` foreign dst/src rejection:** the two guard calls in `build` are only line-covered via the shared helper; a guard-one-but-not-the-other regression would ship silently.
   - **`build` multiple-defs:** the docstring's headline differentiator (multiple defs) is only exercised single-dst.

### Suggestions

4. **`requireSameFunction` comment stale** — `src/MIR/Machine.cpp:142-143`: says "MachineBasicBlock … no back-link to its function," but the check is `mbb->getParent() != &mf` — it works *because* there is a back-link. Leftover from when the comment jointly covered `Register`.
5. **Document the equality/hash footgun** — cross-function registers compare/hash equal (id-only, intentional and matching `llvm::Register`), so keying a `set`/`dict` on registers from two functions silently collides. The owner guard only catches this at *build* time, not for container use. Worth a docstring caveat near `src/MIR/Machine.cpp:511`.
6. **Pin the "operand always attached" invariant** — a one-line comment at the `mi ? mi->getMF() : nullptr` in `MachineOperand.reg` (`src/MIR/Machine.cpp:628`) so a future detached-operand binding can't silently turn a same-function vreg into a spurious "different MachineFunction" rejection.
7. Cheap extra asserts on the existing collision fixture: foreign vreg into `build_icmp` / `build_phi` incoming, and a DSL same-function/different-builder acceptance test (the intentional relaxation from dropping the anchor is currently unlocked by any test).

### Strengths

- `test_cross_function_vreg_collision_rejected` forces `own.id == foreign.id`, then checks rejection across three consuming sites plus acceptance of the same-shaped local register — a true behavioral test, not a line-coverage stunt.
- `test_build_brcond_rejects_physical_register` correctly distinguishes the two failure grounds (physreg passes the owner guard on null owner, but is still rejected as a non-generic vreg).
- The cross-function guard is centralized in one helper with one message and applied uniformly; no consuming site was missed.

**Plan:** apply 1, 4, 5, 6 (small comment/doc fixes), do 2 (factories, matching the item-1 pattern), and add the round-trip + `build` rejection + multiple-defs tests from 3; then re-run the coverage gate + formatters.

projects/eudsl-llvmpy/README.md
```diff
@@ -336,6 +339,9 @@ already-selected **target** MIR directly and run only the back half of codegen.
 `reg_class`/`physreg` resolve register classes and physical registers by name,
 `create_vreg` makes class-constrained vregs, `add_reg` appends operands with the
 full flag set (def/use, implicit, kill, …), and `set_property` marks the function.
+(`build(opcode, dsts, srcs)` is a typed one-shot alternative to
+`build_instr` + `add_reg`: each dst is an `LLT` to mint a fresh vreg for or a
+`Register` to define, each src is a `Register` use.)
```
makslevental, line 344: you added this comment but then didn't add an example

makslevental: Fixed in bb8a78e — the example now uses `build(...)` for the COPY/ADDWrr instructions (kept `build_instr` + `add_reg` only for `RET_ReallyLR`'s implicit $w0 operand, which `build`'s plain-use srcs can't express). Ran the snippet verbatim: `add(2, 3) == 5`.

PR #623, opened 2026-08-25
[eudsl-llvmpy] Reject double-terminator + add MachineBasicBlock.is_terminated (#593 item 6b)

makslevental:
Closes item 6b of #593. That item noted the `build_empty_phi` two-step primitive
is deliberate (nothing to change) and that guarding a loop-body **double
terminator** cleanly "would need a new *is-terminated* `MachineBasicBlock`
accessor."

`MachineIRBuilder` only ever inserts at the **end** of its block — `set_block`
parks the insert point at `end()` and no mid-block insert point is exposed for
MIR — so `getMBB().back()` reliably reflects the block's last instruction. On
that basis:

- **`MachineBasicBlock.is_terminated`** — `True` when the block ends in a
  *barrier* terminator (`G_BR`, a return, …) so control cannot fall through. A
  block ending in only a conditional `G_BRCOND` is **not** terminated: it still
  falls through, which is exactly why the standard `build_brcond` + `build_br`
  pair is two terminators in one block. `isBarrier()` is what distinguishes the
  unconditional closers from the fall-through conditional.
- **Guard `build_br` / `build_brcond`** against appending a terminator to a
  block that already ends in a barrier — an unreachable double terminator past
  the barrier that the verifier (gone under `NDEBUG`) would otherwise catch.
  The `brcond`+`br` pattern, and the DSL `if`/loops built on it, still build
  fine (the block ends in `G_BRCOND`, not a barrier, when the `br` is appended).

Scope note: the generic `build`/`build_instr`/`add_*` path stays caller-trusted
(consistent with #593 item 4); only the high-level convenience terminator
builders are guarded. The DSL keeps rejecting early `return`/`break`/`continue`
inside control flow at the AST level (`RejectUnsupportedJumps`) — this guard is
the MIR-level backstop.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -189,25 +189,32 @@ void requireVReg(llvm::MachineIRBuilder &b, const TypedRegister &reg,
 // (G_BR, a return, ...) -- control cannot fall through off the end. A block
 // ending in only a conditional G_BRCOND is NOT terminated: it still falls
 // through, which is exactly why the standard `build_brcond` + `build_br` pair
-// is two terminators in one block. isBarrier() is what distinguishes the
-// unconditional closers from the fall-through conditional.
+// is two terminators in one block. We require *both* isTerminator() and
+// isBarrier(): isBarrier() distinguishes the unconditional closers from the
+// fall-through conditional, and pairing it with isTerminator() excludes the
+// rare barrier-but-not-terminator instruction (e.g. a target TRAP), which does
+// not close the block for branch purposes.
 bool blockEndsInBarrier(const llvm::MachineBasicBlock &mbb) {
-  return !mbb.empty() && mbb.back().isBarrier();
+  return !mbb.empty() && mbb.back().isTerminator() && mbb.back().isBarrier();
 }
 
 // Appending a terminator to a block that already ends in a barrier builds a
-// malformed block (a second, unreachable terminator past the barrier) that the
-// verifier -- gone under NDEBUG -- would otherwise catch. The builder only ever
-// inserts at the end of its block (setMBB parks the insert point at end(); no
-// mid-block insert point is exposed), so back() is reliably where the next
-// instruction lands. Reject the double-terminator here.
+// malformed block (a second, unreachable terminator past the barrier). The
+// machine verifier catches this, but only when run explicitly via verify():
+// the automatic verification and the build-helper asserts that would flag it
+// during construction are gone under NDEBUG, so in a release build the bad
+// block is otherwise built silently. The builder only ever inserts at the end
+// of its block (setMBB parks the insert point at end(); no mid-block insert
+// point is exposed), so back() is reliably where the next instruction lands.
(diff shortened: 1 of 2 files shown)


makslevental (commented):
## PR review toolkit — aggregated findings (#623, item 6b)

Ran code-reviewer, pr-test-analyzer, comment-analyzer, and a silent-failure pass. **No correctness bugs; the core logic verified correct** (barrier-vs-terminator predicate, the "builder only inserts at block end" invariant, and `is_terminated` semantics all check out against LLVM 22 `GenericOpcodes.td` + `MachineIRBuilder.h`). The items below are comment-precision and hardening suggestions — none block merge.

### Important
- **`requireNotTerminated` comment overstates NDEBUG (`src/MIR/Machine.cpp:~200`).** The comment says the double terminator is one "the verifier -- gone under NDEBUG -- would otherwise catch." The machine verifier is *not* compiled out in release — it's exposed here as the `verify()` binding and runs against the release wheel. What's actually NDEBUG-gated is the build-helper asserts and *automatic* verification; a malformed block slips through silently during normal building unless `verify()` is called explicitly. Suggest rewording to describe that gap. (Note: the same "verifier -- gone under NDEBUG" phrasing pre-exists elsewhere in this file, so this is really a file-wide wording nit surfacing on the new comment.)

### Suggestions
- **`is_terminated` / `blockEndsInBarrier` key off `isBarrier()` only (`Machine.cpp:194-196, 992-994`).** For the current generic-MIR surface this is exactly right (G_BR barrier, G_BRCOND not). The edge case: an instruction that is a *barrier but not a terminator* (e.g. a target `TRAP`) would make `is_terminated` report True and `requireNotTerminated` reject a still-needed terminator. Not reachable through today's bindings; if you want to be strict, `back().isTerminator() && back().isBarrier()` expresses the intent precisely.
- **Guard covers only the terminator builders (`build_br`/`build_brcond`).** Appending an ordinary `build_*` into an already-barrier-terminated block (instruction-after-terminator) is not caught. This is consistent with the deliberate #593-item-4 decision to keep `build_instr`/`add_*`/`build_*` caller-trusted, so it's a scope note rather than a defect.
- **Error-string parenthetical (`Machine.cpp:~204`).** `"...a second (unconditional) terminator"` can be misread as "a second *unconditional* one is disallowed (but a conditional would be fine)"; the guard also fires for `build_brcond`. Minor wording.
- **Test could pin the role prefix (`tests/mir/test_cf_primitives.py:126,129`).** The reject test matches `"second .*terminator"` only; asserting `"build_br "` / `"build_brcond "` would pin which builder is named (guards against a swapped role string). Low value.
- The `is_terminated` docstring's "G_BR **or a return**" is aspirational — no return builder exists in the MIR bindings yet, so only the G_BR path is testable today.

### Strengths
- Tests hit every new C++ line/branch (100% gate holds): empty / G_BRCOND-only / G_BR for the accessor, and both throw + pass-through directions of the guard.
- Comments explain *why* (the brcond+br two-terminator idiom, the isBarrier-not-isTerminator choice, the back()-is-reliable invariant) rather than restating what.
- Validate-before-mutate ordering is correct; no partial-mutation-then-throw path.

PR #624, opened 2026-08-25
[eudsl-llvmpy] Add MachineIRBuilder.branch / cond_branch convenience helpers (#593 item 6d)

makslevental:
Closes item 6d of #593 (the "branch helper that also wires the successor").

**Stacked on #623 (item 6b)** — base branch is `mbb-double-terminator-guard`.
Review/merge #623 first.

MIR tracks a block's CFG **successor list** and its **branch terminator
operands** separately, so emitting a branch and wiring the matching successor
edge are two steps that must stay in sync — forgetting the `add_successor`
produces MIR whose terminator points somewhere the CFG doesn't list, which the
verifier only catches later. Two additive helpers fold both:

- **`branch(dest)`** — `G_BR` to `dest` + the successor edge from the current
  block; returns the `G_BR` so its target can be repointed.
- **`cond_branch(cond, true_block, false_block)`** — `G_BRCOND` on `cond` to
  `true_block`, fall-through `G_BR` to `false_block`, and both successor edges,
  in one call. Returns the fall-through `G_BR` (the repointable one).

`branch`/`cond_branch` sit on top of the existing surface — `build_br`,
`build_brcond`, and `add_successor` are unchanged. The successor edge attaches
to `getMBB()`, the builder's current insert block, which is exactly the block
the terminator lands in.

Because this is stacked on 6b, both helpers route through the same
`requireNotTerminated` double-terminator guard as `build_br`/`build_brcond`, so
neither appends a terminator past a barrier (`test_branch_helpers_reject_second_terminator`).

Adopted at the call sites to show the ergonomics win:
- the if-diamond primitive test (`test_build_if_diamond_with_phi`): the
  `add_successor` × 2 + `build_brcond` + `build_br` block becomes one
  `cond_branch`, and each closing `build_br` + `add_successor` becomes `branch`.
- the DSL lowering in `machine_cf.py` (`_MIRIfOp`, `_MIRLoop`): the four-line
  brcond/br/two-successor blocks collapse to a single `cond_branch`, and the
  preheader/body back-edge `build_br` + `add_successor` pairs to `branch`.

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -329,7 +329,12 @@ For lower-level construction, `MachineIRBuilder` exposes the typed `build_*`
 helpers (`build_constant`, `build_add`, `build_br`, `build_brcond`, `build_phi`,
 …) and a generic `build_instr(opcode)` for any opcode; `MachineFunction.opcode`
 looks a target opcode up by mnemonic (the generated opcode enums are not in
-installed LLVM headers, so lookup is by name via `TargetInstrInfo`).
+installed LLVM headers, so lookup is by name via `TargetInstrInfo`). For
+control flow, `branch(dest)` and `cond_branch(cond, true_block, false_block)`
+fold the terminator emission and the CFG successor edge(s) into one call — MIR
+tracks the successor list and the branch operand separately, so these keep the
+two in sync (`cond_branch` returns the fall-through `G_BR` so its target can be
+repointed).
 
 ### Building target MIR and lowering to native code
 
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -1598,8 +1598,11 @@ void populate_mir(nb::module_ &m) {
           },
           "dest"_a, nb::rv_policy::reference_internal,
           "Unconditional branch to `dest` that also wires the CFG successor "
-          "edge from the current block (build_br + add_successor in one step); "
-          "returns the G_BR so its target can be repointed.")
+          "edge from the current block (add_successor + build_br in one step); "
+          "returns the G_BR so its target can be repointed. This helper owns "
+          "the "
+          "edge -- do not also call add_successor(dest) yourself, as neither "
+          "dedups and the edge would be wired twice.")
       .def(
           "cond_branch",
           [](llvm::MachineIRBuilder &self, TypedRegister cond,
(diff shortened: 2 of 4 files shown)


makslevental (commented):
## PR review toolkit — aggregated findings (#624, item 6d)

Ran code-reviewer, pr-test-analyzer, comment-analyzer, and a silent-failure pass over the incremental diff (stacked on #623). **No correctness bugs in the helpers or the DSL refactor** — all five rewritten `machine_cf.py` call sites were traced and `getMBB()` equals the intended predecessor block at each; operand ordering, the returned fall-through `G_BR`, and the guard routing are all correct. Items below are test-strength and hardening suggestions.

### Important
- **Tests don't pin branch *targets* — a routing swap would pass (`tests/mir/test_cf_primitives.py`, `test_branch_and_cond_branch_wire_successors`).** The assertions check the successor *set* (`{then, else}`), opcode *order* (`["G_BRCOND","G_BR"]`), and that the return is a `G_BR`. A regression that swapped targets — `buildBrCond(cond, false_block)` + `buildBr(true_block)` — produces the identical successor set, identical opcode order, and a `G_BR` return, so it survives all tests (and the DSL tests, which don't execute or assert terminator operands). Returning the `G_BRCOND` instead of the fall-through `G_BR` would likewise pass on opcode but break `_MIRIfOp._repoint_false_edge`. Recommend asserting the `operand(n).mbb.number` routing: G_BRCOND → `then_bb`, returned `false_br` → `else_bb`, `branch` return → `join_bb`.

### Suggestions
- **`cond_branch(cond, blk, blk)` adds a duplicate successor edge (`Machine.cpp:1600-1608`).** `addSuccessor` doesn't dedup, so equal true/false blocks wire the same edge twice → the "duplicate successor" malformed-CFG class. It's a degenerate input and *not a regression* (the raw two-step code did the same), so I'm flagging it as optional hardening rather than a bug — the helper is a natural place to reject `true_block == false_block` or dedup via `isSuccessor`.
- **Mixing the public `add_successor` with the helpers duplicates edges (`Machine.cpp:1588, 1606-1607`).** `blk.add_successor(dest)` then `b.branch(dest)` yields a doubled edge; neither dedups. The DSL refactor itself is clean (it dropped the manual `add_successor` calls). Optional: make the edge-adds idempotent, or document that the helpers own the edge.
- **`branch` docstring lists ops out of code order (`Machine.cpp:1592-1593`).** Says "build_br + add_successor" but the code adds the successor first; `cond_branch`'s docstring is in code order. Purely cosmetic.

### Strengths
- Cross-function rejection is tested per-operand for all four operands (branch `dest`; cond_branch `cond`/`true_block`/`false_block`), and double-terminator rejection is tested for both helpers — every guard call is covered (100% C++ gate holds).
- `branch`'s success assertion uses exact list equality on successors, proving *exactly one* edge is wired.
- DSL refactor is behavior-equivalent to the original two-step code and stays covered by the existing if/loop DSL tests.
- Docstrings/README verified accurate; no forward-reference comments.

Issue #600, opened 2026-08-23
[eudsl-llvmpy] Python-driven register allocators and instruction schedulers (MIR codegen)


makslevental:
## Context

`eudsl-llvmpy` already drives the codegen pass pipeline from Python: `llvm.mir.run_codegen_to_mir` builds a `TargetPassConfig` and runs ISel (`src/MIR/Machine.cpp:1060`), and `emit_object` runs the back half of codegen — **register allocation** + emission — over already-selected MIR (`src/MIR/Machine.cpp:950`, back-half at `979-1042`, using a set+restore of the process-global `-start-after` cl::opt at `982-1002`). Register allocation and instruction scheduling are the codegen stages *inside* that pipeline, but neither can currently be driven by user logic.

This tracks adding **Python-driven register allocators and instruction schedulers** — a custom allocator/scheduler whose core decisions are made by a Python callable.

Feasibility (already investigated): both are registerable and both can be Python-driven. It is **C++-only** (no `llvm-c` C API for either), which is fine — the extension links LLVM's C++ codegen libs and can subclass the interfaces + construct the registry objects directly. Registration points:
- **Reg alloc:** `RegisterRegAlloc::setDefault(&createFn)` (in-process, no CLI) (`RegAllocRegistry.h:54`); selected by `TargetPassConfig::createRegAllocPass` (`TargetPassConfig.cpp:1415`). The pass is a `MachineFunctionPass` filling `VirtRegMap`; idiomatic base `RegAllocBase` (`RegAllocBase.h`) with decision point `selectOrSplit(const LiveInterval&, SmallVectorImpl<Register>&)`.
- **Scheduler:** `MachineSchedRegistry` + `-misched=` (`MachineScheduler.h:173`, selected in `MachineScheduler.cpp:525`), or the `TargetMachine::createMachineScheduler` hook (`TargetMachine.h:167`). Custom strategy implements `MachineSchedStrategy` (`MachineScheduler.h:246`): `initialize`, `pickNode(bool&IsTopNode)`, `schedNode`, `releaseTopNode`, `releaseBottomNode`; wrap via `createSchedLive<Strategy>(C)`.

Binding shape follows MLIR's callback trampoline (`mlir/lib/Bindings/Python/Rewrite.cpp:87-118`) but simpler — hold the `nb::callable` as a member of the C++ subclass and override the virtual directly.

Companion findings doc: `REGALLOCA_ISCHED_PASSES.md`. Sibling issue for IR passes: #599.

[...]

PR #627, opened 2026-08-25
[eudsl-llvmpy] MIR: native trivial MachineScheduler strategy selectable by name (#600 item 1)

makslevental:
Adds a native, trivial-but-correct `MachineSchedStrategy` (first-ready, top-down) registered as `-misched=trivial`, selectable from Python via `emit_object(scheduler="trivial")`. No Python callback yet.

- New TUs `src/MIR/TrivialScheduler.cpp` + `src/MIR/PythonCodegen.cpp` (`populate_python_codegen` into the `llvm.mir` submodule).
- `ScopedDiagnosticCapture`/`withDetail` lifted from `Machine.cpp` into a shared `src/MIR/Diagnostics.h`.
- Selection sets the process-global `-misched` cl::opt under a set+restore RAII (mirrors the existing `-start-after` handling), GIL-serialized.
- Build now undefines `NDEBUG` (`-UNDEBUG`, gated on `LLVM_ENABLE_ASSERTIONS`, applied in both build paths) to match the assertions-on distro ABI — otherwise instantiating LLVM inline codegen templates (`createSchedLive<>`) corrupts codegen.

Test JIT-executes the scheduled `add(i32,i32)` and **proves the strategy actually ran** (a `pickNode` counter fires only when `scheduler="trivial"`, 0 for the default). 100% line+function C++ coverage.

`#600` item 1. Stacked on #626.

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -37,6 +37,18 @@ include_directories(${LLVM_INCLUDE_DIRS})
 link_directories(${LLVM_BUILD_LIBRARY_DIR})
 add_definitions(${LLVM_DEFINITIONS})
 
+# The prebuilt LLVM is built with assertions enabled, so its ScheduleDAG/SUnit
+# (and similar) classes carry extra "#ifndef NDEBUG" members. A Release build
+# here defines NDEBUG and would give those shared classes a smaller layout, so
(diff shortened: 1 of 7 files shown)


makslevental:
Folded into #628. The standalone native trivial MachineScheduler was removed; the scheduler track is now a single "Python-driven MachineScheduler via pickNode callable" PR (#628, items 1-2).

PR #632, opened 2026-08-26
[eudsl-llvmpy] MIR: native trivial RegAllocBase allocator via setDefault (#600 item 4)

makslevental:
Adds a native, trivial-but-correct register allocator — `class PyRegAlloc : MachineFunctionPass, RegAllocBase` (own TU `src/MIR/TrivialRegAlloc.cpp`), modeled on LLVM's `RABasic` — selectable via `emit_object(regalloc="eudsl-trivial")`. No Python callback yet.

- `selectOrSplit` = first non-interfering physreg (`AllocationOrder` + `Matrix->checkInterference == IK_Free`), else spill `VirtReg`. Full `getAnalysisUsage`/`INITIALIZE_PASS` wiring copied from `RABasic`.
- Selected by installing `RegisterRegAlloc::setDefault` under a save/restore RAII.
- **Latent crash fixed:** naively restoring `getDefault()` SIGSEGVs on a fresh process, because `getOptimizeRegAlloc` lazily sets the global default once via a `once_flag`; installing our override first makes that init a no-op, so a plain restore puts back `null`. We restore the `"default"` ctor when the saved value was null.

Tests: valid object emitted + JIT-executes correctly; **fail-if-no-op** (`selectOrSplit` counter fires only when selected); a high-pressure 40-vreg function forces a real spill (`spill_count > 0`). 100% line+function coverage.

`#600` item 4. Stacked on #629.

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -138,6 +138,7 @@ nanobind_add_module(eudslllvm_ext
   src/IR/Intrinsics.cpp
   src/MIR/Machine.cpp
   src/MIR/PythonCodegen.cpp
+  src/MIR/TrivialRegAlloc.cpp
   src/MIR/TrivialScheduler.cpp
 )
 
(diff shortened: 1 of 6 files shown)


makslevental:
Folded into #633. The standalone native trivial register allocator was removed; the register-allocator track is now a single "Python-driven register allocator via selectOrSplit callable" PR (#633, items 4-5).

PR #634, opened 2026-08-26
[eudsl-llvmpy] MIR: marshal LiveInterval + candidate physregs and propagate selectOrSplit exceptions (#600 item 6)

makslevental:
Makes the Python allocator callback **exception-safe** and gives it a richer view (the register-allocator analogue of #629).

- `selectOrSplit` reuses `eudsl::runCodegenPipeline`/`pendingCodegenError`: the trampoline catches and stashes a Python exception, returns a **legal** assignment-or-spill (first free physreg, else native spill — never an interfering/invalid reg, which would abort codegen under `-UNDEBUG`), and the wrapper re-raises after the pipeline. GIL outside the `try`; an early guard stops invoking Python once an error is pending.
- Illegal / out-of-candidate returns now raise a clear `ValueError` ("not one of the legal candidates") — tightening the native fallback the allocator used before this commit.
- The `select` callback signature becomes `(live_interval, candidates)`; binds `LiveInterval` read-only (`reg`, `weight`, `is_spillable`), all `const` accessors (so the `const_cast` for binding is safe).

Tests: a raising callback surfaces as a Python error without crashing and leak-clean; illegal return raises; a legal callback JIT-executes correctly; `LiveInterval` field reads. 100% line+function C++ coverage; clang-format clean.

`#600` item 6. Stacked on #637.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -1413,8 +1413,10 @@ void populate_mir(nb::module_ &m) {
           "mutually exclusive. `regalloc` selects a registered register "
           "allocator by name (e.g. \"eudsl-python\") for the run; an unknown "
           "name raises. `select` is a Python callable that drives the eudsl "
-          "allocator's selectOrSplit: it receives the legal candidate physreg "
-          "ids as a list[int] and returns an id to assign or None to spill. "
+          "allocator's selectOrSplit: it receives the vreg's LiveInterval and "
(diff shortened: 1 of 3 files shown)


makslevental:
Superseded by #644, which rebases this work onto the merged Python-subclassable scheduler design on `main` (dropping the old `pick=` scheduler code and reusing `Diagnostics.h`'s `pendingCodegenError`/`runCodegenPipeline`) and adds `register_regalloc(name, cls)` with an optional `priority` queue hook. Closing in favor of #644.

PR #637, opened 2026-08-26
[eudsl-llvmpy] MIR: Python-driven register allocator via selectOrSplit callable (#600 items 4-5)

makslevental:
Adds a Python-driven register allocator (native allocator + callback folded into one commit; lives in `src/MIR/PythonCodegen.cpp` alongside the scheduler). `class PyRegAlloc : MachineFunctionPass, RegAllocBase` (modeled on LLVM's `RABasic`, registered `regalloc="eudsl-python"`) makes its `selectOrSplit` decision either natively or via a user Python callable selected with `emit_object(select=<callable>)`.

- Native `selectOrSplit`: first `AllocationOrder` register with `Matrix->checkInterference == IK_Free`, else spill `VirtReg`; full `getAnalysisUsage`/`INITIALIZE_PASS` wiring copied from `RABasic`. The `select` callable receives the legal candidate physreg ids as a `list[int]` and returns an id to assign or `None` to spill (an out-of-set/wrong-type return falls back to the native policy here; raising is item 6, #634).
- Selection installs the allocator via `RegisterRegAlloc::setDefault` under a save/restore RAII. **Latent crash fixed:** naively restoring `getDefault()` SIGSEGVs on a fresh process — the process-global default is lazily set once via a `once_flag`, so installing our override first makes that init a no-op and a plain restore puts back `null`; we restore the `"default"` ctor when the saved value was null.
- `regalloc` (name) and `select` (callable) are mutually exclusive; a scheduler-side option (`scheduler`/`pick`) may be combined with an allocator-side one (`regalloc`/`select`) in a single `emit_object`.

Tests: a valid object is emitted and JIT-executes correctly; fail-if-no-op counters prove the allocator actually runs when selected (`select` count > 0) and its spill path runs under a high-register-pressure function (spill count > 0). 100% line+function C++ coverage.

`#600` items 4-5. Stacked on #629.

### DIFF
--- a/projects/eudsl-llvmpy/scripts/cpp_coverage.sh
+++ b/projects/eudsl-llvmpy/scripts/cpp_coverage.sh
@@ -67,9 +67,16 @@ if [[ -z "$SO" || ! -f "$SO" ]]; then
 fi
 
 echo ">> checking src/IR + src/MIR coverage (threshold=${THRESHOLD}%)"
+# RegAllocBase.h and AllocationOrder.h under src/MIR are verbatim upstream LLVM
+# private CodeGen headers, vendored only so TrivialRegAlloc.cpp can reach their
+# declarations (they are not installed with the LLVM distro). We do not author
+# them and exercise them only incidentally through the allocator, so their
+# inline accessors (getOrder, getOrderLimitEnd, the hint iterator, ...) are not
+# all reachable from our code; exclude them from our authored-code gate.
 "$PY" "${PROJ_DIR}/scripts/check_coverage.py" \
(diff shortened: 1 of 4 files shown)


makslevental:
Superseded by #644, which rebases this work onto the merged Python-subclassable scheduler design on `main` (dropping the old `pick=` scheduler code and reusing `Diagnostics.h`'s `pendingCodegenError`/`runCodegenPipeline`) and adds `register_regalloc(name, cls)` with an optional `priority` queue hook. Closing in favor of #644.

PR #641, opened 2026-08-26
[eudsl-llvmpy] MIR: Python-subclassable MachineSchedStrategy driving the pre-RA MachineScheduler (#600)

makslevental:
Replaces the `emit_object(pick=callable)` scheduler-callback approach with a
Python-**subclassable** `llvm::MachineSchedStrategy` (#600 items 1–3).

## What
- `mir.MachineSchedStrategy` binds the real `llvm::MachineSchedStrategy` (nanobind
  trampoline); Python overrides `initialize`, `get_policy`,
  `release_top_node`/`release_bottom_node`, `pick_node() -> (SUnit, is_top) | None`,
  `sched_node`.
- `mir.register_scheduler(name, cls)` adds a runtime `MachineSchedRegistry` node;
  `emit_object(scheduler="name")` runs it as the pre-RA MachineScheduler. Multiple
  named strategies coexist and appear in `registered_schedulers()`.
- `mir.MachineSchedPolicy` / `mir.SUnit` bindings; `mir.ReadyQueueStrategy`
  pure-Python convenience base (override only `pick(ready)`).
- **Ownership:** LLVM owns a C++ `OwningPyStrategy` adaptor — nanobind refuses to
  move a Python-created instance into a default-deleter `unique_ptr` — which
  forwards into the Python instance and drops it under the GIL. Registered classes
  live in a Python-owned dict so no subclass types are pinned at teardown.
- **Exceptions:** every override is a `-fno-exceptions` boundary; Python errors are
  stashed in `pendingCodegenError` and re-raised after the pipeline winds down
  (never thrown across LLVM frames), with an `isScheduled`-safe shadow ready-set
  giving `pick_node` a legal fallback.
- **Build:** matches the distro's ABI flags via `HandleLLVMOptions`.

### DIFF
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -125,6 +125,7 @@ nanobind_add_module(eudslllvm_ext
   src/IR/JIT.cpp
   src/IR/Intrinsics.cpp
   src/MIR/Machine.cpp
+  src/MIR/PythonCodegen.cpp
 )
 
 set(eudslllvm_ext_libs
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -30,8 +30,6 @@ if(EUDSLLLVM_STANDALONE_BUILD)
 
   set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
   list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
-  # Match the distro's ABI-affecting flags (NDEBUG/assertions, ...); nanobind's
-  # per-target -fexceptions/-frtti below re-enable what the bindings need.
   include(HandleLLVMOptions)
   include(AddLLVM)
 endif()
--- a/projects/eudsl-llvmpy/CMakeLists.txt
+++ b/projects/eudsl-llvmpy/CMakeLists.txt
@@ -30,6 +30,9 @@ if(EUDSLLLVM_STANDALONE_BUILD)
 
   set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
   list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
+  # Match the distro's ABI-affecting flags (NDEBUG/assertions, ...); nanobind's
+  # per-target -fexceptions/-frtti below re-enable what the bindings need.
+  include(HandleLLVMOptions)
   include(AddLLVM)
 endif()
 
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -380,6 +380,35 @@ with ir.Context() as ctx:
     assert add(2, 3) == 5
 ```
 
+### Python-driven instruction scheduling
+
+`mir.MachineSchedStrategy` is the pre-RA MachineScheduler strategy interface,
+subclassable from Python. Override `initialize(dag)`, `get_policy()`,
+`release_top_node(su)` / `release_bottom_node(su)` (LLVM hands you ready
+`SUnit`s; you maintain your own ready set), `pick_node()` → `(SUnit,
+is_top_node)` or `None` when nothing is ready, and `sched_node(su, is_top)`.
+Register a subclass under a name and select it per emission:
+
+```python
+class MySched(mir.MachineSchedStrategy):
+    def initialize(self, dag): self.q = []
+    def get_policy(self):
+        p = mir.MachineSchedPolicy(); p.only_top_down = True
+        p.should_track_pressure = False; return p
+    def release_top_node(self, su): self.q.append(su)
+    def release_bottom_node(self, su): pass
+    def pick_node(self): return (self.q.pop(0), True) if self.q else None
+    def sched_node(self, su, is_top): pass
+
+mir.register_scheduler("mysched", MySched)
+obj = mmi.emit_object(scheduler="mysched")   # runs MySched as the pre-RA scheduler
+```
+
+For the common case, subclass `mir.ReadyQueueStrategy` and override only
+`pick(ready)`. A strategy that raises has its exception re-raised out of
+`emit_object` — stashed and re-raised after codegen winds down, never thrown
+across LLVM's `-fno-exceptions` frames.
+
 ## Limitations
 
 - **`break`, `continue`, and early `return` inside DSL control flow are not
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -405,14 +405,7 @@ obj = mmi.emit_object(scheduler="mysched")   # runs MySched as the pre-RA schedu
 ```
 
 For the common case, subclass `mir.ReadyQueueStrategy` and override only
-`pick(ready)`. A strategy that raises has its exception re-raised out of
-`emit_object` — stashed and re-raised after codegen winds down, never thrown
-across LLVM's `-fno-exceptions` frames.
-
-A subclass may also override the optional lifecycle hooks `register_roots(self)`
-(the initial ready set is complete), `enter_mbb(self, mbb)` / `leave_mbb(self)`
-(a block, possibly several scheduling regions, is starting/ending); omitting them
-keeps LLVM's defaults.
+`pick(ready)`.
 
 ## Limitations
 
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -409,6 +409,11 @@ For the common case, subclass `mir.ReadyQueueStrategy` and override only
 `emit_object` — stashed and re-raised after codegen winds down, never thrown
 across LLVM's `-fno-exceptions` frames.
 
+A subclass may also override the optional lifecycle hooks `register_roots(self)`
+(the initial ready set is complete), `enter_mbb(self, mbb)` / `leave_mbb(self)`
+(a block, possibly several scheduling regions, is starting/ending); omitting them
+keeps LLVM's defaults.
+
 ## Limitations
 
 - **`break`, `continue`, and early `return` inside DSL control flow are not
--- a/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
+++ b/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
@@ -19,11 +19,6 @@
 
 namespace eudsl {
 
-// A Python exception raised in a scheduler override is stashed here rather than
-// thrown across llvm::legacy::PassManager::run: libLLVMCodeGen is
-// -fno-exceptions, so unwinding those frames std::terminates. Thread-local;
-// declared extern with a single definition in PythonCodegen.cpp so there is one
-// TLS object and no per-TU TLS-init wrapper.
 extern thread_local std::exception_ptr pendingCodegenError;
 
 // Run a codegen pass manager, then re-raise any exception a scheduler override
--- a/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
+++ b/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
@@ -22,8 +22,8 @@ namespace eudsl {
 // A Python exception raised in a scheduler override is stashed here rather than
 // thrown across llvm::legacy::PassManager::run: libLLVMCodeGen is
 // -fno-exceptions, so unwinding those frames std::terminates. Thread-local;
-// defined once in PythonCodegen.cpp (an inline thread_local would duplicate its
-// TLS init in every including TU).
+// declared extern with a single definition in PythonCodegen.cpp so there is one
+// TLS object and no per-TU TLS-init wrapper.
 extern thread_local std::exception_ptr pendingCodegenError;
 
 // Run a codegen pass manager, then re-raise any exception a scheduler override
--- a/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
+++ b/projects/eudsl-llvmpy/src/MIR/Diagnostics.h
@@ -10,11 +10,33 @@
 #include <llvm/IR/DiagnosticInfo.h>
 #include <llvm/IR/DiagnosticPrinter.h>
 #include <llvm/IR/LLVMContext.h>
+#include <llvm/IR/LegacyPassManager.h>
+#include <llvm/IR/Module.h>
 #include <llvm/Support/raw_ostream.h>
 
+#include <exception>
 #include <string>
 
 namespace eudsl {
+
+// A Python exception raised in a scheduler override is stashed here rather than
+// thrown across llvm::legacy::PassManager::run: libLLVMCodeGen is
+// -fno-exceptions, so unwinding those frames std::terminates. Thread-local;
+// defined once in PythonCodegen.cpp (an inline thread_local would duplicate its
+// TLS init in every including TU).
+extern thread_local std::exception_ptr pendingCodegenError;
+
+// Run a codegen pass manager, then re-raise any exception a scheduler override
+// stashed during the run (before the caller inspects captured diagnostics).
+inline void runCodegenPipeline(llvm::legacy::PassManager &pm, llvm::Module &m) {
+  pendingCodegenError = nullptr;
+  pm.run(m);
+  if (pendingCodegenError) {
+    std::exception_ptr e = pendingCodegenError;
+    pendingCodegenError = nullptr;
+    std::rethrow_exception(e);
+  }
+}
 
 // MIR parsing and codegen report failures through LLVMContext::diagnose --
 // their return values are only a bare success/failure bit, so the rich reason
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -53,11 +53,13 @@
 #include <vector>
 
 namespace eudsl {
-// Defined in PythonCodegen.cpp: the Python class registered under a scheduler
-// name (empty object if none), and install/clear the per-thread active class
-// the shared registry ctor reads while the pipeline runs.
-nb::object schedulerClass(const std::string &name);
-void setActiveSchedClass(nb::object cls);
+// Defined in PythonCodegen.cpp: the class registered under a scheduler name (an
+// invalid object if the name was not registered via register_scheduler), the
+// shared MachineSchedRegistry ctor those names share, and install/clear the
+// per-thread active class the ctor reads while the pipeline runs.
+nb::type_object schedulerClass(const std::string &name);
+llvm::MachineSchedRegistry::ScheduleDAGCtor registeredSchedCtor();
+void setActiveSchedClass(nb::type_object cls);
 void clearActiveSchedClass();
 } // namespace eudsl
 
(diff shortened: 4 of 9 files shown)


makslevental:
## Automated review — code quality, error handling, comments, and test honesty

Reviewed the PR diff with four focused passes (general code quality, silent-failure/error-handling, comment accuracy, and test correctness/coverage/honesty). No changes were made — findings only. Line numbers are against the PR head.

### 🔴 Critical

**1. Unguarded Python construction can `std::terminate` across `-fno-exceptions` frames** — `src/MIR/PythonCodegen.cpp:247` (`createRegisteredPyStrategy`)
```cpp
auto strategy = std::make_unique<OwningPyStrategy>(activeSchedClass());
```
`activeSchedClass()` runs the registered subclass's `__init__`, and this ctor is invoked by LLVM's MachineScheduler pass **from inside `pm.run()`** (libLLVMCodeGen, `-fno-exceptions`). If `__init__` raises, nanobind throws `nb::python_error` and unwinds those frames → `std::terminate` — exactly the outcome the six overrides' `try/catch`+stash machinery exists to prevent. The construction path is the one fallible Python call that bypasses the stash. *(Found independently by the error-handling and code-quality passes.)* Fix is symmetric: wrap in `try { ... } catch (...) { eudsl::pendingCodegenError = std::current_exception(); return llvm::createSchedLive(c); }`. **Untested** — no test exercises a raising `__init__`.

**2. `register_scheduler` never checks the class is a bound `MachineSchedStrategy` subclass → `inst_ptr` UB** — `src/MIR/PythonCodegen.cpp:265` (validation) + `:181` (`OwningPyStrategy`)
Registration only checks `nb::hasattr` for the six method names. A duck-typed pure-Python class with those attributes registers fine; at emit time `nb::inst_ptr<llvm::MachineSchedStrategy>(pyStrategy)` reinterprets a non-nanobind `PyObject` with no type/null check → garbage `inner`, then LLVM calls virtuals through it → memory corruption with no diagnostic. Reject any `cls` that isn't a subclass of the bound type (e.g. `PyObject_IsSubclass`) with a clear `type_error` before it can reach `inst_ptr`.

### 🟠 Important

**3. Built-in scheduler names silently run the built-in — and the `schedulerClass` `LCOV_EXCL` marks a *reachable* line** — `src/MIR/PythonCodegen.cpp:284-286`, `Machine.cpp` resolution loop, `registered_schedulers` docstring `:346`
`emit_object`'s name resolution walks the whole `MachineSchedRegistry` (`getList()`), which also contains LLVM's built-in nodes. Passing a built-in name resolves a non-null ctor, then `schedulerClass(name)` returns empty via the `LCOV_EXCL`-excluded `return nb::object();` (so that exclusion's "unreachable" rationale is wrong — it's reachable and untested), `setActiveSchedClass(empty)` runs, and LLVM **silently schedules with its built-in strategy** instead of erroring. Also makes `registered_schedulers()`'s docstring ("registered in this extension") inaccurate — it lists LLVM built-ins too. Decide the intended behavior (reject names not in `_scheduler_classes`, or document the passthrough) and fix the docstring. *(The sibling `LCOV_EXCL` in `createRegisteredPyStrategy:243` **is** legitimately unreachable — only this one is questionable.)*

**4. `test_pick_node_bad_return_propagates` uses `pytest.raises(Exception)` — too broad** — `tests/mir/test_python_scheduler.py:266`
The only exception test that doesn't pin type/message; it would pass on any error (e.g. an unrelated `RuntimeError` or a leak-gate assertion), so it can't distinguish the intended `nb::cast<pair<SUnit*,bool>>` failure path. Pin the type + `match=` like the sibling tests.

**5. `test_fresh_instance_per_function` doesn't prove "per function"** — `tests/mir/test_python_scheduler.py:211-228`
It builds one MachineFunction and asserts `len(instances) == 1`, which holds whether the ctor runs per-function, once-ever, or is cached. Emit two functions (or two runs on distinct MFs) and assert the count rises to 2.

**6. Bottom-up tests have no witness that Python drove a bottom-up schedule** — `tests/mir/test_python_scheduler.py:365, 394`
`_BottomUpStrategy` records nothing; the tests assert only ELF magic / `add(7,8)==15`. Because scheduling is semantics-preserving and the DAG is a straight-line chain, neither distinguishes "Python's `release_bottom_node`/`pick_node(is_top=False)` drove it" from the shadow-fallback or a default schedule. Add a recording trace (as the top-down `test_python_strategy_actually_drives_scheduling` does) and assert on it.

**7. Comment rot: `test_bottom_up_pick_raise_drains_bottom_shadow` describes a top/bottom shadow split that no longer exists** — `tests/mir/test_python_scheduler.py:449`
The implementation now uses a single combined `shadow` (both `release_*` push into it; `pickNode` drains it front-to-back, reporting direction via `isTopReady()`). Rename/reword away from "top shadow, then bottom shadow."

**8. LLVM brace-style violation** — `src/MIR/PythonCodegen.cpp:265-266`
The `if (!nb::hasattr(...))` guard's `throw` body spans two lines; per CLAUDE.md a multi-line body takes braces.

### 🟡 Suggestions

- **`nb::object` → `nb::type_object`** for the class parameters (`registerScheduler`/`schedulerClass`/`setActiveSchedClass`, and the `eudsl::` decls in `Machine.cpp`) — CLAUDE.md prefers the most specific nanobind type; these are always Python *class* objects.
- **Dead alias**: `mir._scheduler_classes = _scheduler_classes` in `mir_strategies.py` is never read from C++ (C++ reads `llvm.mir_strategies._scheduler_classes`); either drop it or note it's for Python introspection.
- **`Diagnostics.h`** `extern`/`inline thread_local` rationale is slightly off — an `inline thread_local` is still a *single* instance (C++17); the real cost is a per-TU TLS init wrapper, not duplicate storage.
- **Loose `match`**: `test_unknown_scheduler_name_raises:136` uses `match="scheduler"`; tighten to `"unknown scheduler"`.
- **Untested true-branches**: no emit runs with `should_track_pressure=True`, and `get_policy` returning a wrong *type* (vs raising) is untested — both are only covered in the false/raise direction.
- **Trivial DAG**: the `add(i32,i32)` block has no scheduling freedom, so the JIT numeric asserts prove emission validity, not that a specific Python order was honored (the recording-based top-down tests carry the real proof). A DAG with reorder freedom would make the numeric result a discriminating witness.
- **`CLAUDE.md`** editable-install note: worth double-checking the "site-packages copy" mechanism against `editable.mode=redirect` reality before it's trusted as ground truth.

### ✅ Strengths

- The override-level exception design (`pendingCodegenError` + `runCodegenPipeline` re-raise) is the correct pattern for crossing `-fno-exceptions` codegen frames: GIL acquired **outside** every `try`, first-error-wins with no lost-error path, and `emit_object`'s catch moves `pm` into `EmittedOwned` before rethrow so nothing dangles. RAII restore of `-misched`/active-class/`-start-after` runs on the exception path too.
- The `pickNode` shadow wind-down is a genuinely clever safety net (skips `isScheduled` nodes, reports real readiness direction, drains to `nullptr`).
- Real "Python drove it" witnesses for the top-down path (`test_python_strategy_actually_drives_scheduling`, `test_two_strategies_coexist` with per-strategy counters + isolation) that run even off-AArch64; `test_reregister_replaces_scheduler` and `test_register_scheduler_missing_method_raises` are precise; skips are honest (backend/host gating, not failure-hiding); every raise test pairs with `assert_no_leaks()`.
- `std::deque` for `schedNames` (address stability for the borrowed `const char*`), Python-owned `_scheduler_classes` (avoids pinning subclass types past nanobind teardown), and the `HandleLLVMOptions` ABI-flag match (per-target `-fexceptions/-frtti` correctly win) are all sound.

_Reviewed from git refs only; nothing was modified._

projects/eudsl-llvmpy/src/MIR/Diagnostics.h
```diff
@@ -10,12 +10,34 @@
...
+#include <exception>
 #include <string>
 
 namespace eudsl {
 
+// A Python exception raised in a scheduler override is stashed here rather than
+// thrown across llvm::legacy::PassManager::run: libLLVMCodeGen is
+// -fno-exceptions, so unwinding those frames std::terminates. Thread-local;
+// declared extern with a single definition in PythonCodegen.cpp so there is one
+// TLS object and no per-TU TLS-init wrapper.
```
makslevental: delete this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -362,6 +376,20 @@ class MirModule {
     llvm::TargetMachine *tm = build->tm;
     llvm::MachineModuleInfo *info = &build->mmiwp->getMMI();
 
+    // Resolve the requested scheduler against the strategies registered via
+    // register_scheduler (not the whole MachineSchedRegistry, which also holds
+    // LLVM's built-ins) so an unknown -- or built-in -- name fails cleanly
+    // before any codegen state is touched. All registered names share the
+    // createRegisteredPyStrategy ctor, which -misched is pointed at below.
```
makslevental: remove this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -404,6 +432,43 @@ class MirModule {
     } restore{startAfter, saved};
     startAfter = "finalize-isel";
 
+    // Select the pre-RA MachineScheduler strategy, when one was requested:
+    // -misched is a process-global option whose value is the ScheduleDAGCtor
+    // the scheduler pass reads. Set+restore it under the same GIL-serialized
+    // assumption as -start-after; its value type is a constructor pointer.
```
makslevental: remove this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -467,6 +543,9 @@ class MirModule {
 
 } // namespace
 
+// Defined in PythonCodegen.cpp.
+void populate_python_codegen(nb::module_ &m);
+
```
makslevental: move this up in this file where the rest of the PythonCodegen defines are

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -0,0 +1,374 @@
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Python-subclassable pre-RA MachineScheduler strategy.
+// mir.MachineSchedStrategy binds llvm::MachineSchedStrategy via a trampoline;
+// register_scheduler adds a MachineSchedRegistry node so
+// emit_object(scheduler="name") can select it. LLVM owns an OwningPyStrategy
+// adaptor that forwards into the Python instance, since nanobind won't move a
+// Python-created instance into a default-deleter unique_ptr.
```

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -0,0 +1,374 @@
...
+// object (nb_trampoline.base()) by name with the GIL held outside the try; a
+// Python exception is stashed in pendingCodegenError and a legal value
+// returned, since LLVM's scheduler frames are -fno-exceptions.
+// runCodegenPipeline re-raises after the run.
+//
+// shadow mirrors the ready nodes LLVM releases, recorded before the fallible
+// Python call. It is the pick_node fallback after a stash: LLVM only releases
+// ready nodes, so a still-unscheduled one is always a legal choice. The raw
+// contract puts the real ready set in Python, out of reach once it has failed
+// -- this is the safety net.
```
makslevental: move this comment down to wear the shadow field is defined

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -0,0 +1,374 @@
...
+// (under the GIL) before the pipeline runs and cleared after, so the shared
+// registry ctor knows which class to instantiate per MachineFunction.
+// GIL-serialized, matching the process-global -misched handling in Machine.cpp.
+thread_local nb::type_object activeSchedClass;
+
+// C++ side of mir.MachineSchedStrategy. Each override forwards into the Python
+// object (nb_trampoline.base()) by name with the GIL held outside the try; a
+// Python exception is stashed in pendingCodegenError and a legal value
+// returned, since LLVM's scheduler frames are -fno-exceptions.
+// runCodegenPipeline re-raises after the run.
```
makslevental: delete this comment

projects/eudsl-llvmpy/README.md
```diff
@@ -380,6 +380,40 @@ with ir.Context() as ctx:
...
+
+For the common case, subclass `mir.ReadyQueueStrategy` and override only
+`pick(ready)`. A strategy that raises has its exception re-raised out of
+`emit_object` — stashed and re-raised after codegen winds down, never thrown
+across LLVM's `-fno-exceptions` frames.
+
+A subclass may also override the optional lifecycle hooks `register_roots(self)`
+(the initial ready set is complete), `enter_mbb(self, mbb)` / `leave_mbb(self)`
+(a block, possibly several scheduling regions, is starting/ending); omitting them
+keeps LLVM's defaults.
```
makslevental: remove this comment

projects/eudsl-llvmpy/README.md
```diff
@@ -380,6 +380,40 @@ with ir.Context() as ctx:
...
+    def sched_node(self, su, is_top): pass
+
+mir.register_scheduler("mysched", MySched)
+obj = mmi.emit_object(scheduler="mysched")   # runs MySched as the pre-RA scheduler
+```
+
+For the common case, subclass `mir.ReadyQueueStrategy` and override only
+`pick(ready)`. A strategy that raises has its exception re-raised out of
+`emit_object` — stashed and re-raised after codegen winds down, never thrown
+across LLVM's `-fno-exceptions` frames.
```
makslevental: remove the part about exceptions here

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -0,0 +1,418 @@
...
+}
+
+} // namespace
+
+namespace eudsl {
+
+// Validate that cls subclasses MachineSchedStrategy and defines the required
+// methods, record it, and (if new) add a MachineSchedRegistry node so the
+// pipeline can select it by name. Re-registering a name swaps the class.
+void registerScheduler(const std::string &name, nb::type_object cls) {
```
makslevental: are you sure you can't just use nb::type<PySchedStrategy> here (or something like that) instead of nb::type_object?

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -0,0 +1,418 @@
...
+  // The scheduling DAG passed to initialize(dag). Opaque for now -- a strategy
+  // receives its nodes via release_top_node/release_bottom_node.
+  nb::class_<llvm::ScheduleDAGMI>(m, "ScheduleDAGMI");
+
+  // The pre-RA MachineScheduler strategy interface, subclassable from Python.
+  // Override initialize(dag), get_policy() -> MachineSchedPolicy,
+  // release_top_node(su) / release_bottom_node(su) (maintain your own ready
+  // set), pick_node() -> (SUnit, is_top_node), and sched_node(su, is_top).
+  // Register with register_scheduler and select via
+  // emit_object(scheduler=name).
```

projects/eudsl-llvmpy/CMakeLists.txt
```diff
@@ -30,6 +30,9 @@ if(EUDSLLLVM_STANDALONE_BUILD)
 
   set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
   list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
+  # Match the distro's ABI-affecting flags (NDEBUG/assertions, ...); nanobind's
+  # per-target -fexceptions/-frtti below re-enable what the bindings need.
```
makslevental: remove this comment

projects/eudsl-llvmpy/src/MIR/Diagnostics.h
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/README.md
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/CMakeLists.txt
makslevental: Done in a69c34a.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in a69c34a — moved the `populate_python_codegen` forward-declaration up next to the `namespace eudsl` block that forward-declares the other PythonCodegen entry points.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in a69c34a — moved the shadow-queue explanation down onto the `shadow` field and dropped the terser duplicate that was there.

projects/eudsl-llvmpy/README.md
makslevental: Done in a69c34a — dropped the exception sentence from that paragraph.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: No — those are two different things. \`nb::type<PySchedStrategy>()\` is a runtime accessor that returns the *single* bound Python type object for the C++ base (i.e. \`mir.MachineSchedStrategy\` itself); it names one fixed type, and it is not usable as a binding parameter type (it is a function returning \`nb::object\`).

\`cls\` here is the caller-supplied *subclass* being registered — a different Python type per call (\`MySched\`, \`ReadyQueueStrategy\`, ...). We need a parameter that receives an arbitrary Python class, which is exactly what \`nb::type_object\` is (the most specific nanobind type for "a Python type"). We then use \`nb::type<llvm::MachineSchedStrategy>()\` on the next line as the *base* to check \`cls\` is a subclass of — which is the correct use of \`nb::type<...>\`.

So \`nb::type_object\` stays.

makslevental: Correction — you're right, there is: `nb::type_object_t<T>`. Its `check_` is `PyType_Check(h) && PyType_IsSubtype(h, nb::type<T>())`, and as a parameter type its caster runs that in `from_python`, so a non-subclass is rejected at the call boundary. Done in b4d0878: `cls` is now `nb::type_object_t<llvm::MachineSchedStrategy>`, the manual `PyObject_IsSubclass` check is gone, and the stub reads `type[MachineSchedStrategy]`. (The required-methods loop stays — `type_object_t` only checks subclass-ness, not that the methods exist.)

PR #644, opened 2026-08-26
[eudsl-llvmpy] MIR: Python-driven register allocator (#600 items 4-6)

makslevental:
Adds a Python-driven register allocator, the register-allocation counterpart to the Python-subclassable scheduler (#641). It supersedes the old callback-stack PRs #637/#634 (rebased onto the new scheduler design on `main`, dropping the old `pick=` scheduler code and reusing `Diagnostics.h`'s `pendingCodegenError`/`runCodegenPipeline`).

## Design

`PyRegAlloc` is a fixed C++ `MachineFunctionPass` built on LLVM's `RegAllocBase` driver (the RABasic skeleton), registered by name in the `RegisterRegAlloc` registry. This deliberately mirrors how LLVM's own allocators (`basic`, `greedy`, `fast`, …) are structured — **the pass *is* the allocator, chosen by name**, not a driver-owned strategy object subclassed from Python. (The scheduler's subclass-a-strategy shape reflects `ScheduleDAGMILive` owning a `unique_ptr<MachineSchedStrategy>`; register allocation has no such split, so a callback/named-class shape is the faithful analogue.) Python fills in the decisions.

Three ways drive it, all mutually consistent with the scheduler API and independent of `scheduler=`:

- **`emit_object(regalloc="eudsl-python")`** — native first-free / spill policy.
- **`emit_object(select=<callable>)`** — one-shot `selectOrSplit` callback: given the vreg's `LiveInterval` and the legal candidate physreg ids (`list[int]`), return an id to assign or `None` to spill.
- **`register_regalloc(name, cls)` + `emit_object(regalloc="name")`** — a fresh instance per `MachineFunction` whose `select_or_split(self, li, candidates)` drives allocation; it may also define `priority(self, li) -> float` to order the allocation queue (highest first; defaults to spill weight, so the RABasic ordering is no longer hardcoded).

`regalloc=` and `select=` are mutually exclusive.

## Exception safety

Reuses the shared `pendingCodegenError` stash: a raising or illegal callback (and a raising `priority` / `__init__`) is caught, stashed, and re-raised out of `emit_object` after codegen winds down **legally** (native fallback for `selectOrSplit`, spill-weight fallback for `priority`), never thrown across LLVM's `-fno-exceptions` frames. `RegisterRegAlloc` default install uses a save/restore RAII with the `once_flag` null-restore fix (a plain restore SIGSEGVs on a fresh process).

## Tests / coverage

24 new tests (native counter witness, spill path, `select=` invoke/spill/illegal/raise, `register_regalloc` drive/fresh-instance-per-function/priority/priority-raise/select-raise/illegal/init-raise/reregister, JIT execution, and scheduler-independence), plus JIT execution proving allocated code stays correct. **100% C++ line+function coverage**; vendored `RegAllocBase.h`/`AllocationOrder.h` excluded from the authored-code gate.

`#600` items 4-6.

### Coverage note
100%% C++ line+function coverage is *modulo* two deliberate exclusions: the vendored upstream headers (`RegAllocBase.h`/`AllocationOrder.h`), and the `spillVirtReg` `!isSpillable()` branch (`LCOV_EXCL_LINE`) — a reachable-in-principle sentinel that no hand-built test constructs an infinite-weight interval to hit. The `RegisterRegAlloc` "default"-missing guard is also `LCOV_EXCL` (the `"default"` node is always registered by codegen).

### DIFF
--- a/projects/eudsl-llvmpy/scripts/cpp_coverage.sh
+++ b/projects/eudsl-llvmpy/scripts/cpp_coverage.sh
@@ -67,9 +67,15 @@ if [[ -z "$SO" || ! -f "$SO" ]]; then
 fi
 
 echo ">> checking src/IR + src/MIR coverage (threshold=${THRESHOLD}%)"
+# RegAllocBase.h and AllocationOrder.h under src/MIR are verbatim upstream LLVM
+# private CodeGen headers, vendored only so PythonCodegen.cpp can reach their
+# declarations (they are not installed with the LLVM distro). We do not author
+# them and exercise them only incidentally through the allocator, so their
+# inline accessors are not all reachable; exclude them from our authored gate.
 "$PY" "${PROJ_DIR}/scripts/check_coverage.py" \
   --llvm-cov "$LLVM_COV" \
   --profdata "${COV_DIR}/eudslllvm.profdata" \
   --objects "$SO" \
   --sources "${PROJ_DIR}/src/IR" "${PROJ_DIR}/src/MIR" \
+  --ignore-filename-regex 'src/MIR/(RegAllocBase|AllocationOrder)\.h$' \
   --threshold="${THRESHOLD}"
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -528,6 +528,12 @@ public:
     if (regAllocCtor) {
       llvm::RegisterRegAlloc::FunctionPassCtor prev =
           llvm::RegisterRegAlloc::getDefault();
+      // LCOV_EXCL_START -- "default" is always registered by codegen, so this
+      // never fires; the guard turns a missing default into a loud, local error
+      // instead of a null ctor that crashes a later default emit.
+      if (!prev && !defaultRegAllocCtor)
+        throw std::runtime_error("no default register allocator to restore to");
+      // LCOV_EXCL_STOP
       restoreRegAlloc.value = prev ? prev : defaultRegAllocCtor;
       restoreRegAlloc.active = true;
       llvm::RegisterRegAlloc::setDefault(regAllocCtor);
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -18,6 +18,7 @@
 #include <llvm/CodeGen/MachineOperand.h>
 #include <llvm/CodeGen/MachineRegisterInfo.h>
 #include <llvm/CodeGen/MachineScheduler.h>
+#include <llvm/CodeGen/RegAllocRegistry.h>
 #include <llvm/CodeGen/TargetInstrInfo.h>
 #include <llvm/CodeGen/TargetOpcodes.h>
 #include <llvm/CodeGen/TargetPassConfig.h>
@@ -61,6 +62,11 @@ nb::type_object schedulerClass(const std::string &name);
 llvm::MachineSchedRegistry::ScheduleDAGCtor registeredSchedCtor();
 void setActiveSchedClass(nb::type_object cls);
 void clearActiveSchedClass();
+nb::type_object regAllocClass(const std::string &name);
+void setActiveRegAllocClass(nb::type_object cls);
+void clearActiveRegAllocClass();
+void setPendingSelectCallback(nb::callable cb);
+void clearPendingSelectCallback();
 } // namespace eudsl
 
 // Defined in PythonCodegen.cpp.
@@ -368,7 +374,13 @@ public:
   // BuildOwned is consumed into an EmittedOwned, so "build-path only" and the
   // one-shot are enforced by the variant's active alternative rather than a
   // pair of runtime flags.
-  nb::bytes emitObject(std::optional<std::string> scheduler) {
+  nb::bytes emitObject(std::optional<std::string> scheduler,
+                       std::optional<std::string> regalloc,
+                       std::optional<nb::callable> select) {
+    if (regalloc && select) {
+      throw nb::value_error(
+          "emit_object accepts at most one of regalloc= and select=");
+    }
     if (std::holds_alternative<EmittedOwned>(state_))
       throw std::runtime_error("object already emitted");
     BuildOwned *build = std::get_if<BuildOwned>(&state_);
@@ -386,6 +398,31 @@ public:
       if (!schedClass.is_valid())
         throw std::runtime_error("unknown scheduler: " + *scheduler);
       schedCtor = eudsl::registeredSchedCtor();
+    }
+
+    // Resolve the requested register allocator against the RegisterRegAlloc
+    // registry by name (built-ins, "eudsl-python", and register_regalloc names
+    // all live there) so an unknown name fails cleanly. select= chooses the
+    // built-in "eudsl-python" allocator and installs the callable; a name
+    // registered via register_regalloc additionally installs its class. Also
+    // capture the "default" ctor as the valid restore baseline (see the RAII).
+    std::optional<std::string> allocName = regalloc;
+    if (select)
+      allocName = "eudsl-python";
+    llvm::RegisterRegAlloc::FunctionPassCtor regAllocCtor = nullptr;
+    llvm::RegisterRegAlloc::FunctionPassCtor defaultRegAllocCtor = nullptr;
+    nb::type_object regAllocClass;
+    if (allocName) {
+      for (llvm::RegisterRegAlloc *node = llvm::RegisterRegAlloc::getList();
+           node; node = node->getNext()) {
+        if (node->getName() == *allocName)
+          regAllocCtor = node->getCtor();
+        if (node->getName() == "default")
+          defaultRegAllocCtor = node->getCtor();
+      }
+      if (!regAllocCtor)
+        throw std::runtime_error("unknown register allocator: " + *allocName);
+      regAllocClass = eudsl::regAllocClass(*allocName);
     }
 
     // Verify the hand-built MIR up front so malformed input (the prior PRs'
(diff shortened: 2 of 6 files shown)


projects/eudsl-llvmpy/scripts/cpp_coverage.sh
```diff
@@ -67,9 +67,15 @@ if [[ -z "$SO" || ! -f "$SO" ]]; then
 fi
 
 echo ">> checking src/IR + src/MIR coverage (threshold=${THRESHOLD}%)"
+# RegAllocBase.h and AllocationOrder.h under src/MIR are verbatim upstream LLVM
+# private CodeGen headers, vendored only so PythonCodegen.cpp can reach their
+# declarations (they are not installed with the LLVM distro). We do not author
+# them and exercise them only incidentally through the allocator, so their
+# inline accessors are not all reachable; exclude them from our authored gate.
```
makslevental, line 74: remove this comment

makslevental:
## Review summary (code quality · error-handling · tests · comments)

Ran four focused review passes (code quality/CLAUDE.md compliance, silent-failure/error-handling, test correctness+coverage+honesty, comment accuracy). **No critical issues.** The error-stash/re-raise pattern, the `RegisterRegAlloc` default save/restore (incl. the `once_flag` null-restore fix), the fresh-per-function instance lifetime, and the RAII restore ordering all verified correct. Comments verified accurate (queue-ordering claim, "verbatim from RABasic" for `getAnalysisUsage`/`INITIALIZE_PASS`, coverage-script rationale, README examples) with no forward-reference violations.

### Important

1. **Brace-style violation (CLAUDE.md).** `PythonCodegen.cpp:466-468` — the candidate loop's `if` has a condition spanning two lines but an unbraced body:
   ```cpp
   if (Matrix->checkInterference(VirtReg, PhysReg) ==
       llvm::LiveRegMatrix::IK_Free)
     candidateRegs.push_back(PhysReg);
   ```
   CLAUDE.md: "a conditional takes braces whenever its condition spans multiple lines or its body spans multiple lines." Only occurrence in the new code.

2. **Test overclaims — `test_register_regalloc_priority_orders_queue`.** It asserts only that `priority()` was *called* (`priorities` non-empty) + ELF, not that the returned key affected ordering. A regression that calls `priority()` but enqueues by `weight` anyway would still pass. Ordering is genuinely hard to observe through a semantics-preserving object, so this is an acceptable limitation only if the name/docstring are made honest — rename to `..._priority_is_consulted`, or at least assert `len(priorities) == <#vregs enqueued>` to witness the docstring's "runs for every enqueued interval" claim.

3. **Test overclaims — `test_register_regalloc_lists_and_replaces`.** The "replace" half only asserts the registry-node count is unchanged; the actual swap (`classes[name] = cls`) is not witnessed. If that line regressed (kept the old class) the test still passes. Register class A, re-register a distinct subclass B carrying a marker (in `__init__`/`select_or_split`), emit, and assert B's marker fired.

4. **Coverage gap — "fresh instance per MachineFunction" is only tested per *emission*.** `test_register_regalloc_fresh_instance_per_emission` uses two single-function emissions, so the per-function-within-one-module path (`runOnMachineFunction` re-instantiating for each function in one pipeline) is untested. Add a two-`MachineFunction` module and assert two distinct instances from **one** `emit_object`. Also: store instance references, not raw `id(self)` — the first instance is freed before the second is created, so CPython can reuse the `id`, making `seen_ids[0] != seen_ids[1]` a potential spurious failure.

### Suggestions

5. **`~PyRegAlloc` drops its `nb::object` members without acquiring the GIL** (`PythonCodegen.cpp:~373`), unlike `OwningPyStrategy::~OwningPyStrategy` which does. Verified **not a live bug** (the pass is moved into `EmittedOwned`/`state_` and destroyed only at `MirModule` teardown, with the GIL held), but it's inconsistent with every other Python-touching dtor here and wouldn't be free-threading-safe. Adding `nb::gil_scoped_acquire` costs nothing. Relatedly, the class comment's premise "the pass is destroyed while the Python thread holds the GIL during emit_object" is slightly off — the pass *outlives* `emit_object` inside `EmittedOwned` (the conclusion still holds).

6. **`spillVirtReg` unspillable branch is `LCOV_EXCL_LINE`'d** (`PythonCodegen.cpp:512`). This is a reachable-in-principle behavioral branch (the "allocation impossible" sentinel), not genuinely unreachable — it's excluded only because no test builds an unspillable interval. Fine as a pragmatic call, but the "100% coverage" claim should note this exclusion (or add a test).

7. **`RegisterRegAlloc` restore-to-null mitigation has no guard if `"default"` is ever absent** (`Machine.cpp:~552`). `defaultRegAllocCtor` is only set when a `"default"` node is found; today it always is (a `libLLVMCodeGen` static). If it regressed to absent, `prev == nullptr` would restore `nullptr` and a *later* default emit would crash on the null ctor — a deferred, non-local failure. A hard check at install time (require `defaultRegAllocCtor != nullptr` when `regAllocCtor` set and `prev == nullptr`) makes it fail loudly and locally.

8. **Illegal-return error message misdescribes wrong-type returns** (`PythonCodegen.cpp:~499`). "selectOrSplit returned a register that is not one of the legal candidates" is emitted both when `try_cast<unsigned>` succeeds-but-no-match and when it fails because the return was the wrong type entirely (e.g. a `str`). For the latter, a type-specific message ("expected an int physreg id or None") would be more actionable.

9. **Comment nits.** `PythonCodegen.cpp:546` "copied from RABasic::runOnMachineFunction" is copied *and augmented* (the instance-construction prologue is new) — "adapted from" is more accurate. Optional: tighten the ~10-line allocator banner and the ~13-line `RestoreRegAlloc` rationale if comment volume is a concern; align hook param naming (`live_interval` in the binding docstring vs `li` in README/signatures).

### Strengths
- Honest witnesses for a semantics-preserving pass: reset-before/assert-after `selectOrSplit`/spill counters, plus JIT execution against closed-form expected values (`test_jit_executes_spilled_high_pressure`).
- The `runs_only_when_selected` negative test (counter == 0 on default emit) doubles as a guard on the `setDefault` restore/`once_flag` path.
- Complete exception-safety matrix (raise in `__init__`/`priority`/`select`, illegal return, unknown name, mutual exclusivity), each pinning the real message and ending in `assert_no_leaks()`.
- Vendored-header coverage exclusions are legitimate and well-commented.

makslevental:
Addressed all Important + Suggestion items from the review above in 0e8e0ba (100% coverage retained, 55 scheduler+regalloc tests pass):

**Important**
1. Braced the multi-line `if` in the candidate loop (CLAUDE.md).
2. `priority` test rewritten (`..._consulted_per_interval`): asserts priority is consulted once per interval **for exactly the set the allocator assigns** (`set(priorities) == set(selected)`), dropping the unverifiable "orders the queue" claim. (Note: after coalescing, `add` presents 2 intervals, not 3 — the earlier count assumption was wrong.)
3. `lists_and_replaces` now witnesses the swap **behaviorally**: re-register with a distinct class, emit, assert the replacement was the one instantiated.
4. Fresh-instance test holds instance **references** (not `id()`) and compares identity, removing the freed-id-reuse flakiness; docstring made honest (fresh per emission; the hand-built API builds one MachineFunction per module). Filed #645 to add a multi-`MachineFunction` module API so the per-function path can be tested within one emission.

**Suggestions**
5. `~PyRegAlloc` now drops its `nb::object` members under `nb::gil_scoped_acquire`, matching `OwningPyStrategy`.
6. Coverage note added to the PR description: the `spillVirtReg` unspillable branch is `LCOV_EXCL`'d and unexercised.
7. Added a guarded, `LCOV_EXCL`'d hard error if the `"default"` allocator ctor is ever absent (turns the deferred null-ctor crash into a loud, local failure).
8. Split the illegal-return handling: a non-int / non-None return now raises **`TypeError`** ("must return an int physreg id or None") vs the **`ValueError`** for an out-of-set id — with a new test covering the type path.
9. "copied from" → "adapted from" on `runOnMachineFunction`.

Also corrected a doc inaccuracy the run surfaced: `LiveInterval.weight` can be `+inf` (must-not-spill), so the "finite" claim was wrong — the accessor test now asserts `weight >= 0.0` and `is_spillable == isfinite(weight)`.

Left as-is: comment-length trimming of the allocator banner / `RestoreRegAlloc` rationale (the review advised keeping the "why"), and `live_interval`/`li` param-name cosmetics.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -388,6 +400,31 @@ class MirModule {
       schedCtor = eudsl::registeredSchedCtor();
     }
 
+    // Resolve the requested register allocator against the RegisterRegAlloc
+    // registry by name (built-ins, "eudsl-python", and register_regalloc names
+    // all live there) so an unknown name fails cleanly. select= chooses the
+    // built-in "eudsl-python" allocator and installs the callable; a name
+    // registered via register_regalloc additionally installs its class. Also
+    // capture the "default" ctor as the valid restore baseline (see the RAII).
```
makslevental: remove this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -463,6 +500,65 @@ class MirModule {
...
+    // Select the register allocator by pointing RegisterRegAlloc's process-
+    // global default at the resolved ctor: TargetPassConfig::createRegAllocPass
+    // reads getDefault() when addPassesToEmitFile builds the pipeline below.
+    // Save/restore under the same GIL-serialized assumption as the options
+    // above. The restore value needs care: TargetPassConfig lazily inits
+    // getDefault() to useDefaultRegisterAllocator once per process (a
+    // once_flag). If our override is installed first, that init sees a non-null
+    // default and no-ops permanently, so a plain restore would put back null (a
+    // later default emit would then call a null ctor and crash). Restore to the
+    // captured "default" ctor whenever the saved default was null.
```
makslevental, line 513: remove this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -463,6 +500,65 @@ class MirModule {
...
+      if (regAllocClass.is_valid()) {
+        eudsl::setActiveRegAllocClass(regAllocClass);
+        restoreActiveRegAllocClass.active = true;
+      }
+    }
+
+    // Install the Python select callback for the "eudsl-python" allocator, if
+    // given, and clear it after the run so it never leaks into a later emit.
+    // The pass is constructed while addPassesToEmitFile builds the pipeline
+    // (this thread, under the GIL) and copies the callback then.
```
makslevental, line 550: remove this comment

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -287,6 +319,313 @@ createRegisteredPyStrategy(llvm::MachineSchedContext *c) {
...
+
+// Priority-queue entry: an interval and its dequeue key (higher dequeues
+// first). The key is the spill weight by default, or the register_regalloc
+// class's priority(li) when it defines one. Register number tie-breaks for a
+// deterministic order (matching RABasic's CompSpillWeight default).
+struct QueueEntry {
+  float key;
+  const llvm::LiveInterval *li;
+};
+struct CompKey {
```
makslevental, line 353: add a newline here

makslevental:
Superseded by #648: replaces the callback/native regalloc surface with a Python-subclassable `mir.RegAllocBase` (nanobind trampoline) plus raw `SplitAnalysis`/`SplitEditor` region splitting, selected via `emit_object(regalloc="name")`. Stacked on #647 (the SplitKit.h vendoring).

PR #648, opened 2026-08-27
[eudsl-llvmpy] MIR: Python-subclassable RegAllocBase + SplitEditor region splitting (#600)

makslevental:
Stacked on #647 (SplitKit.h vendoring).

Lets Python subclass `llvm::RegAllocBase` and implement an arbitrary register
allocator, including RAGreedy-style live-range region splitting via raw
`SplitEditor` primitives, selected with `emit_object(regalloc="name")`.

## Design

- `mir.RegAllocBase` binds the trampoline **as the value type** (not
  `nb::class_<RegAllocBase, Py...>`), because `llvm::RegAllocBase` has a
  protected destructor — nanobind keys destructor registration off the value
  type's destructibility, so binding it directly aborts when a Python-owned
  instance is freed.
- A pure-C++ `NativeRegAlloc` (first-free-or-spill on a spill-weight queue) is
  both the base the trampoline defers to (default queue + post-stash fallback)
  and the standalone allocator the harness runs when a Python `__init__` raises,
  so the MIR stays valid and the exception re-raises instead of aborting the
  rewriter.
- `PyRegAllocDriver` is a fixed `MachineFunctionPass` occupying the codegen
  allocator slot; it fetches analyses (set modeled on RABasic), builds the
  InlineSpiller + `VirtRegAuxInfo` + `SplitAnalysis`/`SplitEditor`, constructs a
  fresh Python allocator per `MachineFunction`, and drives it via
  `pyInit`/`pyAllocate`.
- `emit_object(regalloc="name")` overrides `RegisterRegAlloc::getDefault()` (the
  process-global the pipeline caches via `call_once`; the `-regalloc` cl::opt is
  ignored once cached) and restores it, mirroring the scheduler's `-misched`.

## Surface

`select_or_split` (required) returning a physreg id or `None`; optional
`enqueue`/`dequeue`/`post_optimization`/`about_to_remove_interval`. Helpers:
`allocation_order`, `spill`, `matrix`, `lis`, `vrm`, `machine_function`,
`split_analysis`, `split_editor`, `new_live_range_edit`. Supporting bindings:
`LiveInterval`, `LiveRegMatrix` + `InterferenceKind`, `VirtRegMap`, `Spiller`,
`LiveIntervals`, `SlotIndex`, `SplitAnalysis` + `BlockInfo`, `SplitEditor` +
`ComplementSpillMode`, `LiveRangeEdit`. `mir.BasicRegAlloc` is the trivial
first-free-or-spill allocator.

## Tests

A 32-test suite: registration contract, query/mutate surface, native + custom
queues, spilling, fresh-instance-per-emission, the exception matrix, and four
region-split choreographies that exercise every `SplitEditor` primitive and
JIT-execute to the correct result. `src/IR` + `src/MIR` C++ coverage is 100%
line + function.

Supersedes #644.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -8,7 +8,10 @@
 #include "MIR/RegAllocBase.h"
 #include "MIR/SplitKit.h"
 
+#include <llvm/Analysis/AliasAnalysis.h>
+#include <llvm/Analysis/ProfileSummaryInfo.h>
 #include <llvm/CodeGen/CalcSpillWeights.h>
+#include <llvm/CodeGen/LiveDebugVariables.h>
 #include <llvm/CodeGen/LiveInterval.h>
 #include <llvm/CodeGen/LiveIntervals.h>
 #include <llvm/CodeGen/LiveRangeEdit.h>
@@ -28,6 +31,7 @@
 #include <llvm/CodeGen/SlotIndexes.h>
 #include <llvm/CodeGen/Spiller.h>
 #include <llvm/CodeGen/VirtRegMap.h>
+#include <llvm/PassRegistry.h>
 
 #include <nanobind/nanobind.h>
 #include <nanobind/stl/pair.h>
@@ -311,52 +315,26 @@ createRegisteredPyStrategy(llvm::MachineSchedContext *c) {
   }
 }
 
-// Trampoline letting Python subclass llvm::RegAllocBase. RegAllocBase is a
-// driver mixed into a MachineFunctionPass: its state (VRM/LIS/Matrix/...) and
-// the init()/allocatePhysRegs() driver are protected, so the harness pass
-// reaches them through the public py* wrappers below rather than owning an
-// adaptor (there is no unique_ptr sink to satisfy, unlike the scheduler).
-class PyRegAllocBase : public llvm::RegAllocBase {
+// A complete pure-C++ RegAllocBase: first-free-or-spill over the target
+// allocation order, driven by a spill-weight priority queue. It is both the
+// base for the Python trampoline (supplying the fallback each virtual defers
+// to) and the standalone allocator the harness runs when a Python __init__
+// raises, so the emitted MIR stays valid and the stashed exception can
+// re-raise.
+class NativeRegAlloc : public llvm::RegAllocBase {
 public:
-  NB_TRAMPOLINE(llvm::RegAllocBase, 6);
-
   // spiller() is pure and dereferenced by allocatePhysRegs; the harness injects
   // the concrete Spiller before driving, so this is never called with null.
   llvm::Spiller &spiller() override { return *injectedSpiller; }
 
-  // Always mirror the register into the native priority queue so the fallback
-  // set stays complete even when Python drives dequeue itself; then forward to
-  // an optional Python enqueue.
   void enqueueImpl(const llvm::LiveInterval *li) override {
     nativeQueue.push({li->weight(), li->reg()});
-    nb::gil_scoped_acquire gil;
-    nb::handle self = nb_trampoline.base();
-    if (eudsl::pendingCodegenError || !nb::hasattr(self, "enqueue"))
-      return;
-    try {
-      self.attr("enqueue")(nb::cast(const_cast<llvm::LiveInterval *>(li),
-                                    nb::rv_policy::reference));
-    } catch (...) {
-      eudsl::pendingCodegenError = std::current_exception();
-    }
   }
 
   const llvm::LiveInterval *dequeue() override {
-    nb::gil_scoped_acquire gil;
-    nb::handle self = nb_trampoline.base();
-    if (!eudsl::pendingCodegenError && nb::hasattr(self, "dequeue")) {
-      try {
-        nb::object choice = self.attr("dequeue")();
-        if (choice.is_none())
-          return nullptr;
-        return nb::cast<llvm::LiveInterval *>(choice);
-      } catch (...) {
-        eudsl::pendingCodegenError = std::current_exception();
-      }
-    }
-    // Native drain (default queue, or winding down after a stash): the queue
-    // holds (weight, reg) rather than pointers because splitting recreates
-    // intervals, so re-fetch and skip registers already assigned or gone.
+    // The queue holds (weight, reg) rather than pointers because splitting
+    // recreates intervals, so re-fetch and skip registers already assigned or
+    // gone (the spiller coalesces snippets).
     while (!nativeQueue.empty()) {
       QueueEntry e = nativeQueue.top();
       nativeQueue.pop();
@@ -370,57 +348,16 @@ public:
   llvm::MCRegister
   selectOrSplit(const llvm::LiveInterval &vreg,
                 llvm::SmallVectorImpl<llvm::Register> &splitLVRs) override {
-    currentSplit = &splitLVRs;
-    nb::gil_scoped_acquire gil;
-    if (!eudsl::pendingCodegenError) {
-      try {
-        nb::object r = nb_trampoline.base().attr("select_or_split")(nb::cast(
-            const_cast<llvm::LiveInterval *>(&vreg), nb::rv_policy::reference));
-        currentSplit = nullptr;
-        // None means Python handled it (spill/split appended new vregs); an int
-        // is the chosen physreg for the driver to assign.
-        if (r.is_none())
-          return llvm::MCRegister();
-        return llvm::MCRegister(nb::cast<unsigned>(r));
-      } catch (...) {
-        eudsl::pendingCodegenError = std::current_exception();
-      }
+    auto order =
+        llvm::AllocationOrder::create(vreg.reg(), *VRM, RegClassInfo, Matrix);
+    for (llvm::MCRegister phys : order) {
+      if (Matrix->checkInterference(vreg, phys) == llvm::LiveRegMatrix::IK_Free)
+        return phys;
     }
-    llvm::MCRegister phys = nativeSelectOrSplit(vreg, splitLVRs);
-    currentSplit = nullptr;
-    return phys;
-  }
-
-  void postOptimization() override {
-    nb::gil_scoped_acquire gil;
-    nb::handle self = nb_trampoline.base();
-    if (!eudsl::pendingCodegenError && nb::hasattr(self, "post_optimization")) {
-      try {
-        self.attr("post_optimization")();
-        return;
-      } catch (...) {
-        eudsl::pendingCodegenError = std::current_exception();
-      }
-    }
-    // No override, or winding down after a stash: run the base cleanup (spiller
-    // post-optimize + dead-remat removal) so the emitted MIR stays valid.
-    llvm::RegAllocBase::postOptimization();
-  }
-
-  // Called by allocatePhysRegs before it drops an interval the spiller left
-  // unused; forwarded to an optional Python about_to_remove_interval.
-  void aboutToRemoveInterval(const llvm::LiveInterval &li) override {
-    nb::gil_scoped_acquire gil;
-    nb::handle self = nb_trampoline.base();
-    if (eudsl::pendingCodegenError ||
-        !nb::hasattr(self, "about_to_remove_interval"))
-      return;
-    try {
-      self.attr("about_to_remove_interval")(nb::cast(
-          const_cast<llvm::LiveInterval *>(&li), nb::rv_policy::reference));
-    } catch (...) {
-      eudsl::pendingCodegenError = std::current_exception();
-    }
+    llvm::LiveRangeEdit lre(&vreg, splitLVRs, *mf, *LIS, VRM,
+                            /*delegate=*/nullptr, &DeadRemats);
+    injectedSpiller->spill(lre);
+    return llvm::MCRegister();
   }
 
   // Public wrappers the harness pass uses to reach the protected driver.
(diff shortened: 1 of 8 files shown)


makslevental:
## Review — Python-subclassable `RegAllocBase` + `SplitEditor` splitting

Reviewed the full diff (`mir-vendor-splitkit..HEAD`) across five focused passes (correctness, error-handling, test honesty/coverage, comments, type design). Overall this is careful, well-documented work that faithfully mirrors the existing `-misched` scheduler machinery, and the `RegisterRegAlloc::setDefault` override/restore across default-vs-custom run orderings is correct. But there is **one high-severity latent UAF** that the current tests structurally cannot catch, plus a few robustness and test-honesty gaps worth addressing before merge.

*(This is a review-only pass — no code was changed.)*

### 🔴 Critical

**1. `heldEdit` outlives the loop-local `SplitVRegs` it references and lingers as an MRI delegate → use-after-free.** `src/MIR/PythonCodegen.cpp` `newLiveRangeEdit` (~:394) + `pyAllocate` reset (~:380).

`newLiveRangeEdit` builds `heldEdit = LiveRangeEdit(&li, *currentSplit, …)`, where `currentSplit` points at `RegAllocBase::allocatePhysRegs`'s **per-iteration loop-local** `SmallVector<Register,4> SplitVRegs` (`third_party/.../RegAllocBase.cpp:115`, declared inside the `while` at `:92`). But `heldEdit` is reset only in `pyAllocate` — *after the entire allocation loop*. So once the `select_or_split` that created it returns:
- `LiveRangeEdit::NewRegs` dangles onto the destroyed `SplitVRegs`, and
- `heldEdit` stays registered as an MRI delegate (ctor calls `MRI.addDelegate(this)`); MRI notifies *all* delegates on new-vreg creation, and `LiveRangeEdit::MRI_NoteNewVirtualRegister` does `NewRegs.push_back(...)`.

Any vreg created in a *later* loop iteration (a spill via `self.spill`, a further split, or inline-spiller coalescing) fires the stale delegate and writes through freed storage. The suite passes only because every split test splits on the first `select_or_split` and its tiny `thru`/`cbz` function has no post-split pressure, so no vreg is created afterward. A real splitting/spilling allocator — the whole point of the priority-queue driver — would corrupt memory. Reset `heldEdit` when `currentSplit` is cleared at the end of `selectOrSplit` (the edit is meant to live within one call, as the comment at ~:390 says), or back it with durable allocator-owned storage. (The reset-in-`pyAllocate` currently there fixes only the teardown-order `MRI.resetDelegate` hazard, not this one.)

### 🟠 Important

**2. Accessors returning pass-frame-owned pointers with `reference_internal` advertise a keep-alive they can't honor.** `src/MIR/PythonCodegen.cpp` `split_analysis`/`split_editor`/`matrix`/`lis`/`vrm`/`machine_function`/`new_live_range_edit` (~:855–887).

`SplitAnalysis`/`SplitEditor` are stack locals in `runOnMachineFunction`, and the analyses are owned by the PassManager — none are owned by the allocator, yet `rv_policy::reference_internal` ties the returned wrapper's lifetime to the allocator instance (which `heldInstance` deliberately keeps alive past the run). A Python allocator that stashes `self.split_editor`/`self.matrix`/`edit = self.new_live_range_edit(li)` and touches it after the emit — or `edit.new_vregs()` in a later callback — reads freed memory. At minimum document "valid only inside allocator callbacks; do not retain," and note `split_editor` also requires `reset(...)` before `open_intv`/`use_intv` (LLVM contract) which nothing surfaces.

**3. Multi-function modules: a later function's `__init__` failure clobbers the first function's stashed exception.** `src/MIR/PythonCodegen.cpp` `runOnMachineFunction` (~:643–670).

`emit_object` runs the whole module, so the driver runs per `MachineFunction`. Every per-interval virtual correctly guards `if (!pendingCodegenError)` before re-entering Python, but `runOnMachineFunction` calls `activeRegAllocClass()` guarded only by `is_valid()`. If function 1 stashed and wound down, function 2 still re-enters Python `__init__`, and if that raises, `pendingCodegenError = current_exception()` overwrites function 1's original error — so `runCodegenPipeline` re-raises the wrong one. Short-circuit to the native fallback when `pendingCodegenError` is already set, and/or only stash `if (!pendingCodegenError)`.

**4. `enqueue`/`dequeue` hand raw `LiveInterval*` across the queue, which splitting invalidates — and the README teaches the unsafe form.** `src/MIR/PythonCodegen.cpp` (~:451–484) + README `ListQueue`.

`NativeRegAlloc` deliberately queues `(weight, reg)` and re-fetches via `LIS->getInterval` *because "splitting recreates intervals"* (comment ~:341). But the Python `enqueue`/`dequeue` surface passes `LiveInterval*` objects, and the README example stores them directly — the exact hazard the native code avoids. Document (and show in the README) the reg-based re-fetch pattern (`store li.reg`; `self.lis.interval(reg)` in `dequeue`).

### 🟡 Medium

**5. `spill`/`new_live_range_edit` deref `*currentSplit`/`*mf` unconditionally** (~:394, :411); both are bound to Python and null outside a `select_or_split`. Calling `self.spill(li)` from a `dequeue` override segfaults instead of raising. Guard and raise `nb::value_error` when `currentSplit`/`mf` is null.

**6. A Python-returned physreg id is handed to `Matrix->assign` unvalidated** (~:496). A bad/interfering/out-of-class id trips an LLVM assertion (abort) rather than a catchable Python error — crossing the `-fno-exceptions` boundary as an abort. Consider validating with `checkInterference(...) == IK_Free` and stashing a Python error otherwise. (The `nb::cast<unsigned>` *type* failure is correctly caught; only semantic validity is unchecked.)

### 🧪 Test honesty & coverage

**7. `test_default_dequeue_skips_already_assigned` (`tests/mir/test_python_regalloc.py`) asserts only `obj[:4]==b"\x7fELF"`** — no witness that `pre_assigned` ever flipped or that the stale-entry `continue` fired. If the pre-assignment loop finds nothing eligible, the guard never runs and the test still passes. Coverage confirms the line runs today, but the *test* doesn't. Record the pre-assign into a module-level list and assert it, or assert the pre-assigned reg is never passed to `select_or_split`.

**8. `test_custom_enqueue_dequeue_queue` has no witness the Python FIFO drove allocation** — native fallback also emits valid ELF, so a non-dispatched override would still pass. Record observed dequeue order and assert it matches FIFO enqueue order.

**9. The `enter_intv_after`/`leave_intv_before` split is the only split shape with no `_executes` (JIT correctness) test** — it's proven to emit, not to compute correctly. Add a `thru(x)==x` execute variant like the other three splits have.

Minor: `first_def_valid` is captured but never asserted; `SlotIndex.__eq__` is only tested reflexively (`start == start`) — add a negative case. The two `LCOV_EXCL` regions are defensible (process-global-state guards), though `Machine.cpp`'s excludes a real `throw` rather than testing it. Note: the whole file is `pytestmark`-skipped without an AArch64-linked backend, and computed-value correctness is only checked on AArch64 hosts — so the coverage gate implicitly requires an AArch64-linked coverage run.

Honest parts worth affirming: witness `split` flags flip only after `se.finish()` inside the split branch (a false condition genuinely fails the test); closed forms (`_hp_closed_form = 48·x`, identity splits) are correct and JIT-asserted; the exception matrix forces each specific override to raise and asserts the real message; `test_native_fallback_spills_after_init_raise_under_pressure` is a real end-to-end check of the fallback + re-raise contract.

### 💬 Comment nit

**10. `src/MIR/PythonCodegen.cpp` ~:555** — the comment calls `select_or_split` "RegAllocBase's only pure heuristic hook without a default," but `enqueueImpl`/`dequeue` are *also* pure virtuals (RegAllocBase.h). The real reason it's the only *required* one is that the trampoline calls it unconditionally while `enqueue`/`dequeue`/`post_optimization` are `hasattr`-guarded. Reword to attribute the requirement to the trampoline.

### Nits (optional)
- `injectedSpiller` is `public` while the other injected pointers are `protected` — no external reader; make it `protected` for consistency.
- `NativeRegAlloc` carries Python-subclass-only plumbing (`currentSplit`/`splitAnalysis`/`splitEditor`/`heldEdit`/`newLiveRangeEdit`/`spill`); it's dead weight when it runs as the standalone fallback (the fallback even passes `&sa,&se` that the native path never reads). Consider moving that plumbing to `PyRegAllocBase`.
- `regallocNames()` leaks a `deque` mirroring `schedNames()`, but nothing borrows it (selection goes through `setDefault`, not a registry parser) — harmless but unnecessary.
- `emit_object` docstring concatenates an awkward lone `"it "` fragment across string literals.

### Strengths
The `setDefault`/`getDefault` override+restore (with the cl::opt-sentinel fallback) correctly handles the process-global `call_once` caching that the `-regalloc` cl::opt can't reach; the stash-or-defer-to-native pattern is uniform across all five virtuals and guarantees forward progress so the `-fno-exceptions` pipeline winds down to a clean re-raise; comments are rationale-focused with no forward-references; and the LCOV_EXCL justifications are accurate. Binding the trampoline as the value type (protected `~RegAllocBase`) is well-reasoned.

makslevental:
## Review fixes applied (`38022dd`)

All findings from the review above are addressed. C++ coverage stays 100% line + function; full suite green (728 passed).

**🔴 Critical**
- **heldEdit UAF** — freed at the end of each `select_or_split` (via `clearSplitContext`) instead of in `pyAllocate`, so the `LiveRangeEdit` never outlives the loop-local `SplitVRegs` it references nor lingers as an MRI delegate into later iterations.

**🟠 Important**
- **Mis-anchored accessors** — `matrix`/`lis`/`vrm`/`machine_function`/`split_analysis`/`split_editor`/`new_live_range_edit` now bind with `rv_policy::reference` (not `reference_internal`); docstrings state they're borrowed, callback-scoped handles ("do not retain"). (`LiveIntervals.interval` keeps `reference_internal` — that `LiveInterval` genuinely lives as long as the `LiveIntervals`.)
- **Multi-function error overwrite** — the harness no longer re-enters Python when an error is already pending, and preserves the first stashed error. Note: `create_machine_function` yields exactly one `MachineFunction`, so `runOnMachineFunction` runs once per `emit_object` today; the guard is defensive and `LCOV_EXCL`'d with that justification (same as the existing `is_valid` guard).
- **enqueue/dequeue across the queue** — both now traffic in register ids (stable across splitting); `dequeue` re-fetches by id and skips already-assigned/removed regs. README + tests updated.

**🟡 Medium**
- **spill/new_live_range_edit outside select_or_split** — now raise `value_error` instead of dereferencing null `currentSplit`.
- **Unvalidated physreg return** — validated as a free candidate in the allocation order; a bad choice raises instead of aborting `Matrix::assign`.

**🧪 Test honesty**
- `test_default_dequeue_skips_already_assigned` and the new `test_python_dequeue_skips_already_assigned` now witness the pre-assigned reg is actually skipped (never handed to `select_or_split`).
- `test_custom_enqueue_dequeue_queue` witnesses the Python queue drove dequeue (records the dequeued order).
- Added `test_split_enter_after_leave_before_executes` (JIT correctness for that split shape), a negative `SlotIndex.__eq__` case, and a `first_def` assertion.
- Added misuse-guard and bad-physreg tests.

**💬 Comment** — corrected the `select_or_split` "only pure hook" comment to attribute the requirement to the trampoline (it calls `select_or_split` unconditionally; the others are `hasattr`-guarded).

**Nits** — `injectedSpiller` is now `protected`; the Python-only splitting plumbing moved onto `PyRegAllocBase`, leaving `NativeRegAlloc` a lean fallback (no more dead `&sa,&se` wiring on the fallback path).

One deliberate non-change: the `regallocNames()` leaked deque is kept for consistency with the scheduler's `schedNames()` (harmless, process-lifetime).

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -287,8 +315,482 @@ createRegisteredPyStrategy(llvm::MachineSchedContext *c) {
...
+  }
+
+protected:
+  struct QueueEntry {
+    float weight;
+    llvm::Register reg;
+  };
+  // Highest spill weight first; break ties on the lower register id so the
+  // default order is deterministic.
+  struct QueueLess {
```
makslevental, line 383: is it possible to enable the python subclasses to overload this in order to define their own priority function?

makslevental: Yes — and it already is possible. `QueueLess` here only orders the *native default* queue, which runs when a subclass doesn't override the queue hooks. A Python subclass defines its own priority end-to-end by overriding `enqueue(reg)` / `dequeue()`: keep whatever structure you like (a `heapq`, a sorted list, RAGreedy-style stage buckets, …) and return the next register id from `dequeue()`. `enqueueImpl` always mirrors into the native queue too, so the fallback stays complete, but when you override `dequeue` your ordering is what drives allocation.

For example (spill-weight-descending, i.e. what `QueueLess` does, but in Python):

```python
import heapq

class WeightPriority(mir.RegAllocBase):
    def __init__(self):
        super().__init__()
        self._heap = []            # (-weight, reg)

    def enqueue(self, reg):
        li = self.lis.interval(reg)
        heapq.heappush(self._heap, (-li.weight, reg))

    def dequeue(self):
        return heapq.heappop(self._heap)[1] if self._heap else None

    def select_or_split(self, li):
        ...
```

`enqueue`/`dequeue` traffic in register ids (stable across splitting; the interval is re-fetched via `self.lis.interval(reg)`), and `dequeue` returns an id or `None`. `test_custom_enqueue_dequeue_queue` in the test file exercises exactly this path with a FIFO queue. So there's no need to make `QueueLess` itself overridable — the priority extension point is the pair of `enqueue`/`dequeue` overrides.

Issue #652, opened 2026-08-27
[eudsl-llvmpy] MIR: expose the surface needed for a full RegAllocGreedy in Python (eviction + spill placement)


makslevental:
Follow-up to #648 (Python-subclassable `RegAllocBase` + `SplitEditor` splitting). That PR makes it possible to write a competent custom allocator in Python — assignment (`allocation_order` + `LiveRegMatrix.check_interference`/`is_free`/`assign`), a custom priority queue (`enqueue`/`dequeue`), spilling (`spill`/`InlineSpiller`), and real region/local splitting (`SplitAnalysis` + the full `SplitEditor` primitive set). That's essentially "RABasic + splitting," or a simplified greedy.

It is **not** yet possible to implement a faithful `RegAllocGreedy` in Python, because RAGreedy's two defining mechanisms depend on surface that isn't bound. This issue scopes the additions needed to close that gap.

## Gap 1 — Eviction / last-chance recoloring (hard blocker)

`tryEvict` and `tryLastChanceRecoloring` need to identify the *specific* interfering virtual registers occupying a candidate physreg, weigh them, `unassign` them, and re-enqueue. Today:

- `LiveRegMatrix.check_interference` returns only an `InterferenceKind` (`IK_Free`/`IK_VirtReg`/`IK_RegUnit`/`IK_RegMask`) — not *which* vregs interfere.
- `VirtRegMap` is bound opaque (no `getPhys`/`hasPhys`), and there is no `LiveRegMatrix::query` / `LiveIntervalUnion` binding.

So a Python allocator can tell *that* an evictable vreg exists at a physreg but can't enumerate it. Shadowing "physreg → vreg I assigned" in Python is not a substitute: it misses subregister/alias interference, which RAGreedy queries at the reg-unit level. (`unassign` and re-enqueue via a Python-side queue already work — only the *identification* step is missing.)

**Scope**
- Bind interference enumeration: `LiveRegMatrix::query(li, unit)` and/or a convenience helper `interfering_vregs(li, physreg) -> list[int]` returning interfering virtual-register ids at a candidate physreg.
- Bind `VirtRegMap.getPhys(reg)` / `hasPhys(reg)`.
- Tests: an eviction-based allocator (evict a lower-weight interferer, re-enqueue it) that JIT-executes to the correct result on a case with forced interference.

Unblocks: eviction, last-chance recoloring, and generally the "interesting" (non-basic) allocators.

[...]

PR #655, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: eviction/interference-query surface (#652 gap 1)

makslevental:
Bind the enumeration eviction needs -- which virtual registers occupy a
candidate physreg -- plus the VirtRegMap read side, so a Python allocator can
do RAGreedy-style eviction rather than only spill:

- RegAllocBase.interfering_vregs(li, physreg): ids of assigned vregs whose
  live ranges interfere with `li` on `physreg`. Queries every reg unit of the
  physreg (not a Python-side shadow), so it is alias/subregister-correct.
- VirtRegMap.has_phys(reg) / get_phys(reg): read an interferer's current
  physreg before unassigning it.
- Test: a greedy-style eviction allocator under forced interference evicts the
  interfering no-higher-weight vreg off a candidate physreg, re-enqueues it,
  and JIT-executes to the correct result. A per-reg once-guard bounds the
  cascade so allocation terminates.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -415,9 +415,14 @@ interface — subclassable from Python. The one required override is
 register id, or return `None` after handling it yourself (spill or split, which
 append new virtual registers that get re-enqueued). The allocator queries and
 mutates state through `self`: `allocation_order(li)` (physregs to try, in target
-order), `self.matrix` (`is_free`/`check_interference`/`assign`/`unassign`),
+order), `interfering_vregs(li, physreg)` (ids of vregs whose ranges interfere
+with `li` on that physreg -- assigned to it or to an aliasing physreg; the
+eviction candidates),
+`self.matrix` (`is_free`/`check_interference`/`assign`/`unassign`),
 `self.lis` (LiveIntervals: `instruction_index`, `mbb_start_index`,
-`mbb_end_index`, `has_interval`, `interval`), `self.vrm`,
+`mbb_end_index`, `has_interval`, `interval`), `self.vrm` (VirtRegMap:
+`has_phys`/`get_phys`, to read an interferer's current physreg before evicting
+it),
 `self.machine_function`, and `self.mbfi` (MachineBlockFrequencyInfo:
 `block_freq`, `block_freq_relative_to_entry_block`, `entry_freq`, for
 frequency-weighted spill/split cost models). Optional overrides: `enqueue(reg)`/`dequeue()` (drive
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -30,6 +30,7 @@
 #include <llvm/CodeGen/ScheduleDAGMutation.h>
 #include <llvm/CodeGen/SlotIndexes.h>
 #include <llvm/CodeGen/Spiller.h>
+#include <llvm/CodeGen/TargetRegisterInfo.h>
 #include <llvm/CodeGen/VirtRegMap.h>
 #include <llvm/PassRegistry.h>
 
@@ -40,6 +41,7 @@
 #include <nanobind/stl/vector.h>
 #include <nanobind/trampoline.h>
 
+#include <algorithm>
 #include <deque>
 #include <exception>
 #include <memory>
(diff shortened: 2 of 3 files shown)


makslevental:

Reviewed `gap2-mbfi..gap1-eviction` (two agents). **No critical or important issues.** The interference enumeration is faithful to RAGreedy's real algorithm — per-reg-unit `Matrix->query(li, unit).interferingVRegs()` rather than a Python-side physreg→vreg shadow, which is what gives alias/subregister correctness. The `query()` "valid only until next query()" contract is respected (each unit's range is fully consumed before the next `query()`), dedup is correct, and enumeration is unbounded (no hardcoded cap). `has_phys`/`get_phys` correctly mirror `hasPhys`/`getPhys`, and `get_phys` returning 0 when unassigned is coherent with the documented `has_phys`-first protocol.

### Test honesty — the emit assertion is stronger than it looks
`_evict_log` can't be populated without *real* eviction: the log entry is appended immediately before `matrix.unassign`, and the returned physreg is validated by the C++ `validatedPhysReg` guard, which raises unless `checkInterference == IK_Free`. So "non-empty `_evict_log` + successful `_emit`" proves the interferers were genuinely freed and the register reused. The `_executes` test (`hp(3)==144`, `hp(7)==336` vs closed form) is a real end-to-end correctness proof.

### Suggestions (non-blocking)
1. **`assert all(phys != 0 ...)` is decorative** — the tuple is only logged after the `has_phys` guard, and `get_phys` on an assigned reg never returns 0, so it can't fail. The real proof is elsewhere (validatedPhysReg + JIT). Consider instead witnessing that a re-enqueued reg was later re-dequeued.
2. **Re-enqueue is exercised but not *witnessed*** — the once-guard termination argument is sound, but a run where one eviction fires and everything else spills would pass identically; the test asserts eviction *fired*, not that re-processing happened.
3. **Dedup / alias-correctness path is covered but not behaviorally exercised** — GPR32 (`W`) regs expose a single reg unit, so under `_build_high_pressure` the `std::find` dedup never actually skips. The "queries every reg unit" claim is plausible but unverified by a multi-unit/subreg case.
4. **Docstring wording** — "virtual registers assigned to `physreg`" is slightly imprecise: the reg-unit query also returns vregs on *aliasing* physregs (the point of the query). The trailing "alias/subregister correct" mostly compensates; cosmetic.

### Strengths
Honest, non-hardcoded enumeration matching upstream; both `VirtRegMap` accessors exercised on live data; correct per-instance state isolation; emit test not arch-gated so the bindings are covered everywhere.

makslevental:
> ### Test honesty — the emit assertion is stronger than it looks
> `_evict_log` can't be populated without _real_ eviction: the log entry is appended immediately before `matrix.unassign`, and the returned physreg is validated by the C++ `validatedPhysReg` guard, which raises unless `checkInterference == IK_Free`. So "non-empty `_evict_log` + successful `_emit`" proves the interferers were genuinely freed and the register reused. The `_executes` test (`hp(3)==144`, `hp(7)==336` vs closed form) is a real end-to-end correctness proof.

fix this if possible

> 1. **`assert all(phys != 0 ...)` is decorative** — the tuple is only logged after the `has_phys` guard, and `get_phys` on an assigned reg never returns 0, so it can't fail. The real proof is elsewhere (validatedPhysReg + JIT). Consider instead witnessing that a re-enqueued reg was later re-dequeued.

fix this using the suggested approach

> 2. **Re-enqueue is exercised but not _witnessed_** — the once-guard termination argument is sound, but a run where one eviction fires and everything else spills would pass identically; the test asserts eviction _fired_, not that re-processing happened.

fix this with a witness

> 3. **Dedup / alias-correctness path is covered but not behaviorally exercised** — GPR32 (`W`) regs expose a single reg unit, so under `_build_high_pressure` the `std::find` dedup never actually skips. The "queries every reg unit" claim is plausible but unverified by a multi-unit/subreg case.

verify with a multi-unit/subreg case.

> 4. **Docstring wording** — "virtual registers assigned to `physreg`" is slightly imprecise: the reg-unit query also returns vregs on _aliasing_ physregs (the point of the query). The trailing "alias/subregister correct" mostly compensates; cosmetic.

fix this

makslevental:

1. **Decorative `phys != 0` replaced with a real witness.** The test now records the reg ids handed to `select_or_split` and asserts an evicted reg appears **twice** — proving the eviction actually re-enqueued it and it was re-dequeued/re-processed (addresses both the "decorative assertion" and "re-enqueue not witnessed" points).
2. **Alias/subregister behavior verified behaviorally.** New `test_interfering_vregs_reports_subregister_aliasing`: a GPR32 vreg assigned to a `W` register is found by querying `interfering_vregs` against the enclosing `X` super-register — proving the query walks every reg unit (a physreg→vreg shadow could never find it). Honest note: the pure dedup-*skip* micro-branch needs a coalescer-surviving multi-unit **tuple** assignment, which isn't constructible in hand-built MIR (IMPLICIT_DEF+COPY tuples get coalesced/DCE'd before RA), so I verified the substantive alias claim instead.
3. **Docstring reworded** to say "assigned to `physreg` or to a physreg aliasing it," dropping the imprecise "assigned to `physreg`."

PR #654, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: bind MachineBlockFrequencyInfo + self.mbfi (#652 gap 2)

makslevental:
Expose the block-frequency surface a frequency-weighted spill/split cost
model needs (the weighting RAGreedy applies). The driver already builds
MBFI and threads it into VirtRegAuxInfo/InlineSpiller/SplitEditor, so this
adds a binding + accessor with no new analysis wiring:

- MachineBlockFrequencyInfo bound with block_frequency(mbb) (raw),
  block_freq_relative_to_entry(mbb) (ratio to entry), and entry_freq().
- self.mbfi accessor on RegAllocBase, threaded through PyRegAllocBase::pyInit.
- Test reads frequencies from select_or_split: entry relative freq == 1.0
  equals entry_freq, and a conditionally-reached block is <= entry.

cc #600

PR #656, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: SpillPlacement + EdgeBundles for global splitting (#652 gap 3)

makslevental:
Add the cost-model-driven global region splitting surface RAGreedy's
splitAroundRegion uses. In this LLVM both headers are public
(include/llvm/CodeGen), so no vendoring is needed:

- PyRegAllocDriver now requires EdgeBundlesWrapperLegacy and
  SpillPlacementWrapperLegacy (moving past RABasic's analysis set toward
  RAGreedy's) and threads them into the allocator.
- EdgeBundles bound: get_bundle(block, out), num_bundles(), get_blocks(bundle).
- SpillPlacement bound: prepare(bit_vector), add_constraints, add_pref_spill,
  add_links, scan_active_bundles, iterate, get_recent_positive, finish,
  get_block_frequency, plus BlockConstraint (bitfield entry/exit via
  accessors), the BorderConstraint enum, and a minimal BitVector.
- self.edge_bundles / self.spill_placer accessors.
- Test: an allocator consults the Hopfield network on a cross-block live
  range (constraints from use blocks, through blocks as links), reads the
  recommended in-register bundles, and lets that gate a real SplitEditor
  region split that JIT-executes correctly.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -475,6 +475,17 @@ class RegionSplit(mir.RegAllocBase):
         return None
 ```
 
+For global (region) splitting, `self.edge_bundles` (an `EdgeBundles`:
+`get_bundle(mbb, out)`, `num_bundles()`, `get_blocks(bundle)`) and
+`self.spill_placer` (a `SpillPlacement` Hopfield network: `prepare(bit_vector)`,
+`add_constraints([BlockConstraint])`, `add_pref_spill`, `add_links`,
+`scan_active_bundles`, `iterate`, `get_recent_positive`, `finish`,
+`get_block_frequency`) expose the machinery RAGreedy's `splitAroundRegion`
+drives. Build `mir.BlockConstraint`s (with `mir.BorderConstraint` entry/exit
+prefs) from the live-through use blocks, iterate the network, and read the
+in-register edge bundles out of the `mir.BitVector` passed to `prepare` to
+choose split boundaries.
+
 A fresh allocator instance is constructed per `MachineFunction`. Because this
 build has assertions enabled, an invalid split aborts with a diagnostic rather
 than emitting bad code, and an exception raised in any override propagates out
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -8,9 +8,11 @@
 #include "MIR/RegAllocBase.h"
 #include "MIR/SplitKit.h"
 
+#include <llvm/ADT/BitVector.h>
 #include <llvm/Analysis/AliasAnalysis.h>
 #include <llvm/Analysis/ProfileSummaryInfo.h>
 #include <llvm/CodeGen/CalcSpillWeights.h>
+#include <llvm/CodeGen/EdgeBundles.h>
 #include <llvm/CodeGen/LiveDebugVariables.h>
 #include <llvm/CodeGen/LiveInterval.h>
 #include <llvm/CodeGen/LiveIntervals.h>
@@ -29,6 +31,7 @@
 #include <llvm/CodeGen/ScheduleDAG.h>
 #include <llvm/CodeGen/ScheduleDAGMutation.h>
 #include <llvm/CodeGen/SlotIndexes.h>
+#include <llvm/CodeGen/SpillPlacement.h>
 #include <llvm/CodeGen/Spiller.h>
 #include <llvm/CodeGen/TargetRegisterInfo.h>
 #include <llvm/CodeGen/VirtRegMap.h>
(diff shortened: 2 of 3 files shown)


makslevental:

Reviewed `gap1-eviction..gap3-spillplacement` (two agents). **No critical or important code-quality issues.** The analysis wiring is correct (both `EdgeBundlesWrapperLegacy` and `SpillPlacementWrapperLegacy` are `addRequired` *and* `INITIALIZE_PASS_DEPENDENCY`, no pass-ordering hazard, no `addPreserved` since this is the terminal regalloc pass); the `BitVector` retention contract is safe (the Python-side `reg_bundles` outlives every dependent call, and `SpillPlacement` never re-derefs `ActiveNodes` before the next `prepare`); the bitfield `entry`/`exit` accessors correctly route through `def_prop_rw` lambdas (address of an `: 8` bitfield can't be taken for `def_rw`); ArrayRef↔vector conversions are sound.

### Test honesty (Important)
1. **The "recommendation gates the split" framing is overstated — the negative branch is unreachable by construction.** Every constraint fed in is `PrefReg`/`DontCare`, `add_pref_spill` is empty, and b1 is a transparent link, so a positive-bias-only Hopfield network always converges all-in-register; `entry_out_bundle in inreg` is a foregone conclusion and the `return False`/"fall through" path can never be taken. It *is* a real data dependency on the binding round-trip (a broken `prepare`/`finish`/`set_bits` would flip the gate and fail the test), so it's not decorative — but the test can't demonstrate the network *suppressing* a split. Read it as "the round-trip produces the expected in-register bundle," not "the network decides whether to split." Biasing b1 to spill (non-empty `add_pref_spill`) and asserting a bundle drops out of `inreg` would both fix the framing and exercise the dead branch.
2. **`add_pref_spill([], False)` is a coverage-touch no-op** — biases nothing; hits the line but exercises no behavior.
3. **`get_recent_positive()` is only `isinstance(..., list)`-checked** — given `any_positive is True`, it should be non-empty and could be asserted so; and `finish()`'s `perfect` is only type-checked though it's determinate (`perfect is True` here).

### Suggestions (bindings)
- `get_block_frequency` / `get_bundle` do no bounds check (mirror LLVM, which also omits it) — an out-of-range int from Python is UB rather than a Python exception.
- `get_block_frequency` collapses `BlockFrequency` to a raw `uint64` (pragmatic since `BlockFrequency` isn't bound).

### Strengths
- **Complete coverage of the new surface, running everywhere** (emit test cross-emits via `_AARCH64_LINUX`, not host-gated).
- **`BlockConstraint` field assertions are real, not trivial** — `c0_entry==DontCare`/`c0_exit==PrefReg`/`c0_changes is True` reflect genuine `SplitAnalysis` data for def-block b0, and `c0_number==0` self-guards the "constraints[0] is b0" assumption so it fails loudly rather than mis-asserting.
- Bundle assertions are robust to CFG-bundle numbering (derived from `get_bundle`, not hardcoded); the test drives the real RAGreedy `splitAroundRegion` call sequence in order.

makslevental:
> 1. **The "recommendation gates the split" framing is overstated — the negative branch is unreachable by construction.** Every constraint fed in is `PrefReg`/`DontCare`, `add_pref_spill` is empty, and b1 is a transparent link, so a positive-bias-only Hopfield network always converges all-in-register; `entry_out_bundle in inreg` is a foregone conclusion and the `return False`/"fall through" path can never be taken. It _is_ a real data dependency on the binding round-trip (a broken `prepare`/`finish`/`set_bits` would flip the gate and fail the test), so it's not decorative — but the test can't demonstrate the network _suppressing_ a split. Read it as "the round-trip produces the expected in-register bundle," not "the network decides whether to split." Biasing b1 to spill (non-empty `add_pref_spill`) and asserting a bundle drops out of `inreg` would both fix the framing and exercise the dead branch.

fix this using the "Biasing b1 to spill (non-empty `add_pref_spill`) and asserting a bundle drops out of `inreg` would both fix the framing and exercise the dead branch." approach

> 2. **`add_pref_spill([], False)` is a coverage-touch no-op** — biases nothing; hits the line but exercises no behavior.

fix this

> 3. **`get_recent_positive()` is only `isinstance(..., list)`-checked** — given `any_positive is True`, it should be non-empty and could be asserted so; and `finish()`'s `perfect` is only type-checked though it's determinate (`perfect is True` here).

fix this

> * `get_block_frequency` / `get_bundle` do no bounds check (mirror LLVM, which also omits it) — an out-of-range int from Python is UB rather than a Python exception.

fix this

> * `get_block_frequency` collapses `BlockFrequency` to a raw `uint64` (pragmatic since `BlockFrequency` isn't bound).

bind `BlockFrequency` more thoroughly (fully if possible)

makslevental:

1. **The "network gates the split" framing is now real, both branches exercised.** Added `test_spill_placement_recommends_spill_and_falls_through`: with `PrefSpill` constraints + a strong `add_pref_spill` on the through blocks, the network drops the entry bundle from the in-register set, so `entry_out_bundle not in inreg` and the split is **not** taken (the previously-dead `return False` branch). The positive test still takes the split branch — so both outcomes are now demonstrated.
2. **`add_pref_spill` is no longer a `[]` no-op** — it's called with the through blocks (strong bias) in the spill path and genuinely flips the recommendation.
3. **`get_recent_positive`/`finish` value-checked**: assert `len(recent) >= 1` (given `any_positive`) and `perfect is True`, not just `isinstance`.
4. **Bounds checks.** `get_bundle`/`get_block_frequency` now take a `MachineBasicBlock` (its number is always in range — OOB impossible by construction, and it lets `get_block_frequency` return `BlockFrequency`); `get_blocks` bounds-checks the bundle index and raises `IndexError` (covered by a test that calls it out of range).
5. **`BlockFrequency` bound thoroughly** (in #654, so it's consistent stack-wide); `get_block_frequency` now returns it.

PR #657, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: RAGreedy-completion accessors (#652 gap 4, reframed)

makslevental:
Gap 4 was scoped as binding RegAllocEvictionAdvisor/RegAllocPriorityAdvisor,
but those are infeasible and unnecessary: their (and DefaultEviction/
PriorityAdvisor's) constructors take a const RAGreedy&, RAGreedy is defined
only in the private lib/CodeGen header, and a Python allocator computes its own
priority/eviction anyway. This PR instead binds the read-only accessors a
faithful Python RAGreedy actually still needs:

- LiveInterval.begin_index / end_index / size -- the interval's own extent
  (RAGreedy ranks enqueue priority by size; split strategies need the range).
- SplitAnalysis.get_use_slots() -- per-use SlotIndexes, required for
  local/instruction splitting's gap analysis.
- RegAllocBase.register_cost(physreg) -- per-use register cost, the
  CostPerUseLimit heuristic (the genuinely useful residual of the advisor area,
  via TargetRegisterInfo::getRegisterCosts).
- Test reads all three from select_or_split and asserts extent ordering, use
  slots within the extent, and a non-negative cost.

Not included: the explicit rematerialization write-path
(LiveRangeEdit::canRematerializeAt/rematerializeAt), which needs a VNInfo /
Remat value-numbering surface -- a separate follow-up. InlineSpiller already
rematerializes automatically inside self.spill(), so a competent RAGreedy is
achievable without it.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -519,6 +519,7 @@ public:
     blockFreqInfo = mbfi;
     edgeBundles = eb;
     spillPlacer = spl;
+    regCosts = mfn.getSubtarget().getRegisterInfo()->getRegisterCosts(mfn);
     NativeRegAlloc::pyInit(vrm, lis, mat, sp, mfn);
   }
 
@@ -568,6 +569,16 @@ public:
       }
     }
     return ids;
+  }
+
+  // Per-use cost of `physreg` (the CostPerUseLimit heuristic RAGreedy uses to
+  // decide whether a register is worth allocating). Indexed by physreg id. The
+  // cost table is an ArrayRef into the target's static tables (fixed for this
+  // function), cached in pyInit.
+  unsigned registerCost(unsigned physreg) {
+    if (physreg >= regCosts.size())
+      throw nb::index_error("physreg id out of range");
+    return regCosts[physreg];
   }
 
   // Spill `li` into the current select_or_split's split-vreg vector.
@@ -628,6 +639,7 @@ private:
   llvm::MachineBlockFrequencyInfo *blockFreqInfo = nullptr;
   llvm::EdgeBundles *edgeBundles = nullptr;
   llvm::SpillPlacement *spillPlacer = nullptr;
+  llvm::ArrayRef<uint8_t> regCosts;
   std::unique_ptr<llvm::LiveRangeEdit> heldEdit;
 };
 
@@ -975,6 +987,9 @@ void populate_python_codegen(nb::module_ &m) {
            "`physreg` -- those assigned to `physreg` or to a physreg aliasing "
            "it. Alias/subregister correct (queries every reg unit of "
            "`physreg`, de-duplicating). These are the eviction candidates.")
+      .def("register_cost", &PyRegAllocBase::registerCost, "physreg"_a,
+           "Per-use cost of `physreg` (the CostPerUseLimit heuristic; 0 on "
+           "targets with uniform register cost).")
       .def("spill", &PyRegAllocBase::spill, "li"_a,
            "Spill `li`; new split vregs are appended for re-enqueue. Only "
            "valid inside select_or_split.")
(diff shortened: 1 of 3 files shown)


makslevental:

Reviewed `gap3-spillplacement..gap4-ragreedy-accessors` (two agents). **No critical or important code-quality issues** — a clean additive set of read-only accessors, names mirror LLVM, comments explain *why*. Two adjudicated "crash path" concerns are safe-by-design LLVM-mirroring (`begin_index`/`end_index` assert on empty intervals, `register_cost` OOB for a fabricated physreg id): both are the same "garbage-in" contract as the underlying LLVM calls, and the intended input (`select_or_split`'s `li`, `allocation_order` physregs) is always valid. Guarding would deviate from the API per CLAUDE.md.

### Test honesty — one genuinely tautological assertion (Important)
1. **`assert isinstance(saw["cost"], int) and saw["cost"] >= 0` is coverage-only, zero correctness signal.** `register_cost` reads an `ArrayRef<uint8_t>`, so the result is *always* a non-negative int (0–255) — even a hardcoded `return 0` passes. AArch64 is a uniform-cost target so the value is almost certainly a constant 0, meaning the physreg→cost mapping is never validated. Either assert against a derived/known value (or two physregs with differing documented costs), or explicitly label it a smoke/coverage-only check.

2. **`num_use_slots >= 1` under-specifies a deterministic value.** For `_build_three_block`, `v` yields exactly **2** use slots (def contribution + the killing use, after dedup). Asserting `== 2` would catch a binding that dropped the def contributions or miscounted; `>= 1` would pass even if `get_use_slots` returned a single element.

### Suggestion
3. `size > 0` and `begin < end` are loose (true for any non-empty interval) — but not worthless: `ordered` would catch a begin/end **swap**, and `size > 0` catches a zero/garbage return.

### Strengths
- `begin_valid`/`end_valid`/`slots_valid` are meaningful — they'd fail on a default-constructed `SlotIndex` sentinel, proving the bindings are wired to live data.
- `slots_in_extent` (`begin <= s <= end`, inclusive) is a real cross-binding consistency check, and the inclusive bound is *correct*: the def slot equals `beginIndex()` and the killing use's slot coincides with `endIndex()`, so a strict `s < end` would wrongly fail. The comment is accurate.
- All five bindings exercised on genuinely-emitted MIR (ELF magic checked), with `analyze(li)` called before `get_use_slots()` per its precondition.

### Note (binding, out of test scope)
`register_cost` does no bounds check before `[physreg]`; safe here (only `allocation_order[0]` passed) but an out-of-range id from Python would read OOB.

makslevental:
> Two adjudicated "crash path" concerns are safe-by-design LLVM-mirroring (`begin_index`/`end_index` assert on empty intervals, `register_cost` OOB for a fabricated physreg id): both are the same "garbage-in" contract as the underlying LLVM calls, and the intended input (`select_or_split`'s `li`, `allocation_order` physregs) is always valid. Guarding would deviate from the API per CLAUDE.md.

despite the divergence add guards here

> 1. **`assert isinstance(saw["cost"], int) and saw["cost"] >= 0` is coverage-only, zero correctness signal.** `register_cost` reads an `ArrayRef<uint8_t>`, so the result is _always_ a non-negative int (0–255) — even a hardcoded `return 0` passes. AArch64 is a uniform-cost target so the value is almost certainly a constant 0, meaning the physreg→cost mapping is never validated. Either assert against a derived/known value (or two physregs with differing documented costs), or explicitly label it a smoke/coverage-only check.

try to fix this by asserting against a known non-zero cost if possible (otherwise add the label)

> 2. **`num_use_slots >= 1` under-specifies a deterministic value.** For `_build_three_block`, `v` yields exactly **2** use slots (def contribution + the killing use, after dedup). Asserting `== 2` would catch a binding that dropped the def contributions or miscounted; `>= 1` would pass even if `get_use_slots` returned a single element.

fix this

> 3. `size > 0` and `begin < end` are loose (true for any non-empty interval) — but not worthless: `ordered` would catch a begin/end **swap**, and `size > 0` catches a zero/garbage return.

fix this

> `register_cost` does no bounds check before `[physreg]`; safe here (only `allocation_order[0]` passed) but an out-of-range id from Python would read OOB.

add a check/guard if possible

makslevental:

1. **Guards added despite the LLVM-mirroring divergence.** `register_cost` now bounds-checks the physreg id against `getRegisterCosts().size()` and raises `IndexError` (covered by a test passing a fabricated id). `begin_index`/`end_index` guard `li.empty()` and raise instead of asserting/aborting (the throw is `LCOV_EXCL`'d — empty intervals don't reach Python from `select_or_split`, so it's defensive).
2. **`register_cost` value assertion honestly labeled.** AArch64 has uniform register cost (all 0 — verified), so a non-zero assertion isn't possible; the value check is now labeled a smoke check and the **OOB guard raise** is the meaningful behavioral assertion.
3. **`num_use_slots == 2`** (was `>= 1`) — the def contribution + killing use for this MIR.
4. **`size`/`ordered` tightened** into strong cross-binding checks: `begin_index == get_use_slots()[0]` and `end_index == get_use_slots()[-1]`, tying the interval extent to real program points (a swapped or garbage begin/end now fails).

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -570,6 +570,16 @@ class PyRegAllocBase : public NativeRegAlloc {
...
 
+  // Per-use cost of `physreg` (the CostPerUseLimit heuristic RAGreedy uses to
+  // decide whether a register is worth allocating). Indexed by physreg id.
+  unsigned registerCost(unsigned physreg) {
+    const llvm::TargetRegisterInfo *tri = mf->getSubtarget().getRegisterInfo();
+    llvm::ArrayRef<uint8_t> costs = tri->getRegisterCosts(*mf);
+    if (physreg >= costs.size())
+      throw nb::index_error("physreg id out of range");
+    return costs[physreg];
+  }
```
makslevental, line 582: can we cache this somehow?

makslevental: Done — cached it in `pyInit`. `getRegisterCosts(*mf)` returns an `ArrayRef` into the target's static per-use-cost tables, which is fixed for the (single) MachineFunction this allocator instance drives, so there's no reason to recompute it per `register_cost` call. Added a `llvm::ArrayRef<uint8_t> regCosts` member set once in `pyInit` (`mfn.getSubtarget().getRegisterInfo()->getRegisterCosts(mfn)`); `register_cost` now just bounds-checks against `regCosts.size()` and indexes it. 100% coverage still holds; force-pushed (and #659 re-cascaded on top).

PR #659, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: rematerialization write-path (#652 follow-up)

makslevental:
Bind the explicit rematerialization surface so a Python allocator can recompute
a value at its use rather than keep it live (what InlineSpiller/SplitKit do
internally). Completes the "faithful RAGreedy" surface.

- VNInfo bound (id, def_index, is_phi_def, is_unused); LiveInterval gains
  get_vni_at(idx), num_val_nums, get_val_num_info(i).
- LiveIntervals: instr_from_index(idx), compute_interval(reg) (the interval of
  a freshly-created remat reg), shrink_to_uses(reg) (returns now-dead defs).
- LiveRangeEdit: create(), nested Remat(parent_vni) with orig_mi,
  rematerialize_at(mbb, before, dest_reg, remat, late=, sub_idx=),
  eliminate_dead_defs(dead).
- MachineInstr.substitute_register(from_reg, to_reg, sub_idx=0) to redirect a
  use onto the rematerialized register.
- llvm::LiveIntervals::computeVirtRegInterval is private; reached via the
  explicit-instantiation access trick ([temp.explicit] does not apply access
  control to a pointer-to-member named as a template argument) rather than a
  remove-and-recreate workaround.
- Test rematerializes a MOVi32imm constant at its ADDWrr use, redirects the
  use, recomputes liveness, eliminates the dead original def, and JIT-executes
  to the correct result.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -489,6 +489,19 @@ prefs) from the live-through use blocks, iterate the network, and read the
 in-register edge bundles out of the `mir.BitVector` passed to `prepare` to
 choose split boundaries.
 
+For rematerialization (recomputing a value at its use instead of keeping it
+live), a value's def is reached through its value number: `li.get_vni_at(idx)` /
+`li.get_val_num_info(i)` return a `VNInfo` (`def_index`, `is_phi_def`,
+`is_unused`), and `self.lis.instr_from_index(vni.def_index)` gives the defining
+instruction. Build a `mir.LiveRangeEdit.Remat(vni)` with its `orig_mi` set, then
+`lre.rematerialize_at(mbb, before, dest_reg, remat)` clones the def into a fresh
+`lre.create()` vreg before a use; redirect the use with
+`use_mi.substitute_register(old, new)`, compute the clone's interval with
+`self.lis.compute_interval(new)`, and drop the now-dead original with
+`self.lis.shrink_to_uses(old)` feeding `lre.eliminate_dead_defs(dead)`. (Note the
+built-in spiller already rematerializes trivially-rematerializable defs inside
+`self.spill()`; this surface is for driving remat yourself.)
+
 A fresh allocator instance is constructed per `MachineFunction`. Because this
 build has assertions enabled, an invalid split aborts with a diagnostic rather
 than emitting bad code, and an exception raised in any override propagates out
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -929,6 +929,19 @@ void populate_mir(nb::module_ &m) {
             return &self.getOperand(i);
           },
           "index"_a, nb::rv_policy::reference_internal)
+      .def(
+          "substitute_register",
+          [](llvm::MachineInstr &self, unsigned fromReg, unsigned toReg,
+             unsigned subIdx) {
+            const llvm::TargetRegisterInfo *tri =
+                self.getMF()->getSubtarget().getRegisterInfo();
+            self.substituteRegister(llvm::Register(fromReg),
+                                    llvm::Register(toReg), subIdx, *tri);
+          },
+          "from_reg"_a, "to_reg"_a, "sub_idx"_a = 0,
+          "Replace every use/def of `from_reg` in this instruction with "
+          "`to_reg` (composing `sub_idx` where set); redirects a use onto a "
+          "rematerialized register.")
       .def_prop_ro(
           "parent",
           [](llvm::MachineInstr &self) -> const llvm::MachineBasicBlock * {
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -35,6 +35,7 @@
 #include <llvm/CodeGen/Spiller.h>
 #include <llvm/CodeGen/TargetRegisterInfo.h>
 #include <llvm/CodeGen/VirtRegMap.h>
+#include <llvm/MC/LaneBitmask.h>
 #include <llvm/PassRegistry.h>
 
 #include <nanobind/nanobind.h>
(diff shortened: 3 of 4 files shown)


makslevental:

Reviewed `gap4-ragreedy-accessors..gap5-remat` (two agents). The `computeVirtRegInterval` access trick is verified correct and well-scoped (private member confirmed; the anonymous-namespace friend + required namespace-scope forward declaration resolve correctly since ADL on `llvm::` args wouldn't find the anon-namespace friend; internal linkage keeps it TU-local). The remat mutation ordering (`create` → `rematerialize_at` → `substitute_register` → `compute_interval` → `shrink_to_uses` → `eliminate_dead_defs`) is correct, and the test's read-of-value-numbers-*before*-mutation is necessary (those calls invalidate `li`'s `VNInfo`s).

### Important — stale docstring (flagged independently by both reviewers)
1. **`rematerialize_at`'s docstring says "call `create_and_compute_interval(dest_reg)` after", but no such binding exists** — it was renamed to `compute_interval` (which the README and test correctly use). A user following the C++ docstring hits `AttributeError`. Fix the docstring to `compute_interval(dest_reg)`.

### Test honesty — genuinely honest, with the proof correctly located
The load-bearing assertions are `done` + `slot_valid` + **`ndead >= 1`**: `shrink_to_uses` only reports the original `MOVi32imm` def dead if `substitute_register` actually redirected the sole use off `li.reg`, so `ndead >= 1` proves the redirect took effect. Combined with the `return None` protocol under an assertions-enabled build (a silent no-op would leave `li.reg` with a live use and no physreg → verifier abort, not a pass), the emit + execute tests together prove remat genuinely happened and produced correct code.
- **Honest caveat worth stating:** `x + 42 == 47` holds whether or not the constant was rematerialized, so the arithmetic alone doesn't distinguish remat from normal allocation — the "remat *happened*" proof comes from `done`/`ndead`/the return-None invariant, not the result value.
- **Coverage-only reads (honestly labeled):** `orig_is_movi`, `is_unused is False`, `num_val_nums == 1`, `valno0_id == vni_id`, `needs_split is False` exercise the `VNInfo`/`compute_interval` bindings but are near-tautological on this MIR; the comment openly says they're read before mutating. Fine, just not remat signals.

### Suggestions (bindings)
2. **`rematerialize_at` omits `UsedLanes` and `ReplaceIndexMI`; `eliminate_dead_defs` omits `RegsBeingSpilled`/`AA`.** `late`/`sub_idx` are surfaced (good), but `UsedLanes` (defaulting to `getAll()`) is exactly the "one choice from a set the API offers" that CLAUDE.md's "bind the full API" targets — it matters once a `sub_idx` write meets partially-live lanes. Defensible to defer (no `LaneBitmask` bound yet), but worth noting as deliberate.
3. **`compute_interval`/`shrink_to_uses` assert-abort on a reg with no interval** (raw `getInterval`), and `get_vni_at`/`get_val_num_info` hand out `VNInfo*` invalidated by `eliminate_dead_defs`. Inherent to raw spiller primitives; mitigated by "do not retain" docstrings, consistent with the rest of the file — no catchable-exception guard.
4. **Test-only:** the use-finding loop assumes a single use in a single block and doesn't guard `use_mi is None` (would pass `None` to `rematerialize_at`). Fine for this hand-built MIR; a comment noting the single-use assumption would help.
5. Trivial: the `[temp.explicit]/3` citation is closer to `/p12` in recent editions; substance is accurate.

### Strengths
Access trick correct and documented; the invalidation hazard is both documented in the bindings and *demonstrated* in the test (snapshot-before-mutate); every new binding exercised, with the load-bearing ones proven structurally (`ndead`/`slot_valid`) and end-to-end (JIT `rematc(5)==47`); `is_phi_def` correctly bound as a property so the `not vni.is_phi_def` gate actually works (a method binding would have silently defeated it).

makslevental:
> 1. **`rematerialize_at`'s docstring says "call `create_and_compute_interval(dest_reg)` after", but no such binding exists** — it was renamed to `compute_interval` (which the README and test correctly use). A user following the C++ docstring hits `AttributeError`. Fix the docstring to `compute_interval(dest_reg)`.

fix this 

> * **Honest caveat worth stating:** `x + 42 == 47` holds whether or not the constant was rematerialized, so the arithmetic alone doesn't distinguish remat from normal allocation — the "remat _happened_" proof comes from `done`/`ndead`/the return-None invariant, not the result value.

strengthen this if possible

> * **Coverage-only reads (honestly labeled):** `orig_is_movi`, `is_unused is False`, `num_val_nums == 1`, `valno0_id == vni_id`, `needs_split is False` exercise the `VNInfo`/`compute_interval` bindings but are near-tautological on this MIR; the comment openly says they're read before mutating. Fine, just not remat signals.

strengthen this if possible
 
> 2. **`rematerialize_at` omits `UsedLanes` and `ReplaceIndexMI`; `eliminate_dead_defs` omits `RegsBeingSpilled`/`AA`.** `late`/`sub_idx` are surfaced (good), but `UsedLanes` (defaulting to `getAll()`) is exactly the "one choice from a set the API offers" that CLAUDE.md's "bind the full API" targets — it matters once a `sub_idx` write meets partially-live lanes. Defensible to defer (no `LaneBitmask` bound yet), but worth noting as deliberate.

add the omitted bindings

> 3. **`compute_interval`/`shrink_to_uses` assert-abort on a reg with no interval** (raw `getInterval`), and `get_vni_at`/`get_val_num_info` hand out `VNInfo*` invalidated by `eliminate_dead_defs`. Inherent to raw spiller primitives; mitigated by "do not retain" docstrings, consistent with the rest of the file — no catchable-exception guard.

fix this if possible

> 4. **Test-only:** the use-finding loop assumes a single use in a single block and doesn't guard `use_mi is None` (would pass `None` to `rematerialize_at`). Fine for this hand-built MIR; a comment noting the single-use assumption would help.

add a guard if possible

> 5. Trivial: the `[temp.explicit]/3` citation is closer to `/p12` in recent editions; substance is accurate.

i don't understand this but improve this if possible

makslevental:

1. **Stale docstring fixed** — `rematerialize_at` now says "call `compute_interval(dest_reg)` after" (was the nonexistent `create_and_compute_interval`).
2. **Remat proof strengthened beyond the arithmetic.** The emit test now asserts a real `MOVi32imm` was inserted at the returned remat slot (`instr_from_index(slot).opcode_name == "MOVi32imm"`) and that the new register's own value number is defined exactly at that slot (`get_vni_at(...).def_index == slot`). So "remat happened" is proven structurally, not just by `rematc(x) == x + 42`.
3. **Coverage-only reads strengthened** — the value-number reads are now tied to the remat result (the new reg's def slot) rather than only the pre-mutation snapshot.
4. **Omitted parameters bound.** `rematerialize_at` now exposes `used_lanes` (a newly-bound `LaneBitmask`, default `getAll()`) and `replace_index_mi` (default `None`); `eliminate_dead_defs` exposes `regs_being_spilled`. `LaneBitmask` is bound (`get_all`/`get_none`/`get_as_integer`/`none`/`any`/`==`/`!=`) with its own unit test. (This LLVM's `eliminateDeadDefs` has no `AA` parameter.)
5. **Guards added.** `compute_interval`/`shrink_to_uses` raise `ValueError` on a reg with no interval instead of aborting (covered). The `VNInfo`/`SlotIndex` borrowed-pointer docstrings now explicitly warn they're invalidated by `shrink_to_uses`/`eliminate_dead_defs` (a runtime use-after-free guard isn't possible for a freed pointer).
6. **Test guards `use_mi is None`** before use.
7. **`[temp.explicit]` citation** reworded to describe the access-control carve-out rather than a specific (wrong) paragraph number.

100% C++ coverage; format-clean; JIT-executes.

PR #660, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: priority/pressure surface for faithful RAGreedy (#652)

makslevental:
Bind the read-only inputs RAGreedy's getPriority weighs, so a Python allocator
can reproduce its enqueue priority and pressure heuristics:

- RegAllocBase.reg_class_of(reg) (MRI.getRegClass) returning a
  TargetRegisterClass (now exposing `id`), and num_allocatable_regs(reg_class)
  (RegisterClassInfo.getNumAllocatableRegs) -- the register-pressure denominator.
- RegAllocBase.is_trivially_rematerializable(mi) (TII) -- the remat gate.
- SlotIndex.distance / approx_instr_distance -- instruction-space gap sizes
  (getPriority ranks by size; tryLocalSplit measures gaps).
- Test reads all of them: GPR32's allocatable count is bounded, same-class
  vregs share an id, distances are positive, and the MOVi32imm def is trivially
  rematerializable while COPY/ADDWrr are not.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -25,6 +25,7 @@
 #include <llvm/CodeGen/MachineFunctionPass.h>
 #include <llvm/CodeGen/MachineInstr.h>
 #include <llvm/CodeGen/MachineLoopInfo.h>
+#include <llvm/CodeGen/MachineRegisterInfo.h>
 #include <llvm/CodeGen/MachineScheduler.h>
 #include <llvm/CodeGen/Passes.h>
 #include <llvm/CodeGen/RegAllocRegistry.h>
@@ -33,6 +34,7 @@
 #include <llvm/CodeGen/SlotIndexes.h>
 #include <llvm/CodeGen/SpillPlacement.h>
 #include <llvm/CodeGen/Spiller.h>
+#include <llvm/CodeGen/TargetInstrInfo.h>
 #include <llvm/CodeGen/TargetRegisterInfo.h>
 #include <llvm/CodeGen/VirtRegMap.h>
 #include <llvm/MC/LaneBitmask.h>
@@ -601,6 +603,18 @@ public:
     return regCosts[physreg];
   }
 
+  const llvm::TargetRegisterClass *regClass(unsigned reg) {
+    return mf->getRegInfo().getRegClass(llvm::Register(reg));
+  }
+
+  unsigned numAllocatableRegs(const llvm::TargetRegisterClass *rc) {
+    return RegClassInfo.getNumAllocatableRegs(rc);
+  }
+
+  bool isTriviallyRematerializable(llvm::MachineInstr *mi) {
+    return mf->getSubtarget().getInstrInfo()->isTriviallyReMaterializable(*mi);
+  }
+
   // Spill `li` into the current select_or_split's split-vreg vector.
   void spill(const llvm::LiveInterval &li) {
     if (!currentSplit)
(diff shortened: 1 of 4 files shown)


projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -601,6 +603,25 @@ class PyRegAllocBase : public NativeRegAlloc {
     return regCosts[physreg];
   }
 
+  // The register class of virtual register `reg` (RAGreedy keys priority and
+  // pressure heuristics on it).
```

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -601,6 +603,25 @@ class PyRegAllocBase : public NativeRegAlloc {
...
+  const llvm::TargetRegisterClass *regClassOf(unsigned reg) {
+    return mf->getRegInfo().getRegClass(llvm::Register(reg));
+  }
+
+  // Number of actually-allocatable registers in `rc` (reserved registers
+  // excluded) -- the register-pressure denominator getPriority weighs size
+  // against.
```

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -601,6 +603,25 @@ class PyRegAllocBase : public NativeRegAlloc {
...
+  unsigned numAllocatableRegs(const llvm::TargetRegisterClass *rc) {
+    return RegClassInfo.getNumAllocatableRegs(rc);
+  }
+
+  // Whether `mi` is trivially rematerializable (getPriority prefers remat over
+  // keeping such a value in a register).
```

makslevental:
## Review — priority/pressure surface (#660)

Reviewed by a code-quality pass (signatures verified against the LLVM headers in `mlir_wheel/include`) and a test correctness/coverage/honesty pass. No correctness bugs; all five binding signatures faithfully mirror LLVM (`getRegClass`, `getNumAllocatableRegs`, `isTriviallyReMaterializable`, `distance`/`getApproxInstrDistance`, `getID`). Coverage gate is met — every new function is a single straight-line return with no guard branches. The findings are all about **naming fidelity** and **test honesty** (5 of 6 bindings would survive a "returns a wrong-but-plausible value" mutation).

### Code quality
- **Important — `approx_instr_distance` breaks the `get_` convention of its `SlotIndex` siblings** (`PythonCodegen.cpp:1488`). The four existing SlotIndex accessors keep the LLVM `get` prefix (`get_reg_slot`, `get_base_index`, `get_boundary_index`, `get_next_index`); this new one drops it. Per CLAUDE.md "bind names as LLVM spells them," it should be `get_approx_instr_distance`. (`distance` at :1480 is correct — LLVM literally names it `distance`.)
- Note — `reg_class_of` (`PythonCodegen.cpp:1070`) inserts an `_of` token not in the LLVM name `getRegClass`; strict mirroring would be `reg_class`/`get_reg_class`. Borderline.
- Note — the `reg_class_of` lambda wrapper is redundant; `&PyRegAllocBase::regClassOf` binds directly. Pure style.

### Tests / honesty (`test_priority_pressure_accessors`)
- **`same_class` is tautological** (`test_python_regalloc.py:1702-1703`). It queries the *same* vreg twice (`reg_class_of(li.reg).id == rc.id` where `rc` came from the same call) — proves determinism, not that two distinct same-class vregs share an id. `_build_remat_const` has 3 GPR32 vregs; compare two *different* ones, ideally differentially against a GPR64 vreg (`!=`).
- **`rc_id` is a pure smoke check** (`test_python_regalloc.py:1700`) — `isinstance(..., int)` passes for any int including the wrong class's id. No oracle (same caveat class as the prior-PR "AArch64 costs are uniformly 0" note).
- **`distance` and `approx_instr_distance` aren't distinguished** (`test_python_regalloc.py:1705-1706`) — both only asserted `> 0`, so a miswiring (both calling the same underlying method) passes. Slot distance is the larger multiple, so `distance > approx` is a real differential that would catch a copy/paste swap. (Also: the docstring calls both "instruction-space"; only `approx_instr_distance` is — `distance` is raw slot space.)
- `num_alloc`'s `0 < n < 33` is a loose bound, not a verified value (would pass if it returned total-including-reserved).
- **The one strong assertion** — the remat differential `remat["MOVi32imm"] is True` + `any(v is False ...)` — is the right pattern. Caveat: its `False` branch depends on a non-MOVi32imm vreg surviving to `select_or_split`; the COPY/ADDWrr vregs flow through W0 and are coalescing candidates, so pinning `remat.get("ADDWrr") is False` would be more robust and self-documenting than `any(...)`.

**Net:** ship-ready on correctness; recommend the `get_approx_instr_distance` rename and hardening `same_class`/`rc_id` (distinct-class differential) + `distance > approx` so the test fails when a binding is wrong, not only when it throws.

makslevental:
> * **Important — `approx_instr_distance` breaks the `get_` convention of its `SlotIndex` siblings** (`PythonCodegen.cpp:1488`). The four existing SlotIndex accessors keep the LLVM `get` prefix (`get_reg_slot`, `get_base_index`, `get_boundary_index`, `get_next_index`); this new one drops it. Per CLAUDE.md "bind names as LLVM spells them," it should be `get_approx_instr_distance`. (`distance` at :1480 is correct — LLVM literally names it `distance`.)

fix this

> * Note — `reg_class_of` (`PythonCodegen.cpp:1070`) inserts an `_of` token not in the LLVM name `getRegClass`; strict mirroring would be `reg_class`/`get_reg_class`. Borderline.

fix this

> * Note — the `reg_class_of` lambda wrapper is redundant; `&PyRegAllocBase::regClassOf` binds directly. Pure style.

definitely fix this and find any other instances of the same useless pattern and correct it.

> * **`same_class` is tautological** (`test_python_regalloc.py:1702-1703`). It queries the _same_ vreg twice (`reg_class_of(li.reg).id == rc.id` where `rc` came from the same call) — proves determinism, not that two distinct same-class vregs share an id. `_build_remat_const` has 3 GPR32 vregs; compare two _different_ ones, ideally differentially against a GPR64 vreg (`!=`).

fix this

> * **`rc_id` is a pure smoke check** (`test_python_regalloc.py:1700`) — `isinstance(..., int)` passes for any int including the wrong class's id. No oracle (same caveat class as the prior-PR "AArch64 costs are uniformly 0" note).

fix this

> * **`distance` and `approx_instr_distance` aren't distinguished** (`test_python_regalloc.py:1705-1706`) — both only asserted `> 0`, so a miswiring (both calling the same underlying method) passes. Slot distance is the larger multiple, so `distance > approx` is a real differential that would catch a copy/paste swap. (Also: the docstring calls both "instruction-space"; only `approx_instr_distance` is — `distance` is raw slot space.)

fix this

> * `num_alloc`'s `0 < n < 33` is a loose bound, not a verified value (would pass if it returned total-including-reserved).

fix this

> * **The one strong assertion** — the remat differential `remat["MOVi32imm"] is True` + `any(v is False ...)` — is the right pattern. Caveat: its `False` branch depends on a non-MOVi32imm vreg surviving to `select_or_split`; the COPY/ADDWrr vregs flow through W0 and are coalescing candidates, so pinning `remat.get("ADDWrr") is False` would be more robust and self-documenting than `any(...)`.

fix this

makslevental:
## Fixes applied (amended + cascaded up the stack)

**Code quality**
- `approx_instr_distance` → **`get_approx_instr_distance`** to match its `SlotIndex` siblings (`get_reg_slot`/`get_base_index`/…); bound as a direct member-function pointer.
- `reg_class_of` → **`reg_class`** — drops the non-LLVM `_of` token and matches both the `MachineFunction.reg_class(name)` accessor already in the codebase and the drop-`get` convention of the sibling RegAllocBase accessors (`num_allocatable_regs`, `is_trivially_rematerializable`).
- Removed the redundant forwarding lambda for `reg_class`, and swept **every** other instance of the same pattern: `matrix`, `lis`, `vrm`, `machine_function`, `split_analysis`, `split_editor`, `new_live_range_edit`, `mbfi`, `edge_bundles`, `spill_placer`, and both `SlotIndex` distance accessors now bind the member function directly (`&PyRegAllocBase::…` / `&llvm::SlotIndex::…`) instead of a `[](self){ return self.method(); }` wrapper.
- Removed the three inline-flagged `//` comments above `regClass`/`numAllocatableRegs`/`isTriviallyRematerializable`.
- Fixed the docstring that called `distance` "instruction-space" — it's raw slot space; only `get_approx_instr_distance` is instruction-space.

**Tests** (`test_priority_pressure_accessors`)
- `rc_id`/`same_class` replaced with a real oracle: the vreg's class id `==` the name-resolved `GPR32` id and `!=` `GPR64` (a differential a constant-returning binding fails).
- `num_alloc` now pinned to an independent oracle — `== len(allocation_order(li))` (a separate path through `AllocationOrder`) — instead of the loose `0 < n < 33`.
- `distance`/`approx` now assert `distance > approx > 0`, which catches a copy/paste swap wiring both to the same method.
- The remat differential no longer uses `any(...)`: pinned to `remat["COPY"] is False` + `remat["MOVi32imm"] is True`. (The ADDWrr-def vreg the review suggested gets coalesced into `w0` pre-RA, so it never reaches the allocator; the surviving non-remat def is the live-in `COPY`, which is what's pinned.)

PR #661, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: eviction cost surface for faithful RAGreedy (#652)

makslevental:
Bind the read-only inputs RAGreedy's canEvictInterference / isUnusedCalleeSavedReg
weigh, so a Python allocator can reproduce its eviction cost model:

- RegAllocBase.reg_allocation_hints(reg) / simple_hint(reg) (MRI) -- copy hints,
  for preferring a hinted physreg and counting broken hints in eviction cost.
- LiveRegMatrix.is_phys_reg_used(physreg) -- whether any vreg is assigned there
  yet (the callee-saved eviction bias).
- RegAllocBase.last_callee_saved_alias(physreg) (RegisterClassInfo) -- detect a
  callee-saved register to bias against introducing a CSR spill.
- Test: is_phys_reg_used flips False->True across an assign, the GPR32 order
  mixes callee-saved and caller-saved regs, and a copy-related vreg carries an
  allocation hint whose simple_hint is the hinted physreg.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -423,7 +423,10 @@ eviction candidates),
 `self.lis` (LiveIntervals: `instruction_index`, `mbb_start_index`,
 `mbb_end_index`, `has_interval`, `interval`), `self.vrm` (VirtRegMap:
 `has_phys`/`get_phys`, to read an interferer's current physreg before evicting
-it),
+it; and, for the eviction cost model, `self.reg_allocation_hints(reg)` (a
+`(type, [ids])` pair) / `self.simple_hint(reg)` (broken-hint accounting),
+`self.matrix.is_phys_reg_used`,
+and `self.last_callee_saved_alias(physreg)`),
 `self.machine_function`, and `self.mbfi` (MachineBlockFrequencyInfo:
 `block_freq`, `block_freq_relative_to_entry_block`, `entry_freq`, for
 frequency-weighted spill/split cost models). The `li` passed in is a
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -615,6 +615,38 @@ public:
     return mf->getSubtarget().getInstrInfo()->isTriviallyReMaterializable(*mi);
   }
 
+  // Copy-hint registers for `reg` (physregs and/or vregs it is copy-related
+  // The register-allocation hints for virtual register `reg`: a (type, [ids])
+  // pair. `type` is the hint kind (0 = target-independent copy hints; nonzero
+  // = a target-specific hint the target expands), `ids` the hinted physregs.
+  // RAGreedy prefers a hinted physreg and counts "broken hints" in eviction
+  // cost. `reg` must be a virtual register. (0, []) if it has no hints.
+  std::pair<unsigned, std::vector<unsigned>> regAllocationHints(unsigned reg) {
+    unsigned type = 0;
+    std::vector<unsigned> ids;
+    const auto *hints =
+        mf->getRegInfo().getRegAllocationHints(llvm::Register(reg));
+    if (hints) {
+      type = hints->first;
+      for (llvm::Register h : hints->second)
+        ids.push_back(h.id());
+    }
+    return {type, ids};
+  }
+
+  // The single "simple" copy hint for virtual register `reg` (id, or 0 if
+  // none). `reg` must be a virtual register.
+  unsigned simpleHint(unsigned reg) {
+    return mf->getRegInfo().getSimpleHint(llvm::Register(reg)).id();
+  }
+
+  // The last callee-saved register aliasing `physreg` (id, or 0 if none) --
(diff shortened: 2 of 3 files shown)


makslevental:
## Review — eviction cost surface (#661)

Reviewed by a code-quality pass (signatures verified against `mlir_wheel/include`) and a test correctness/coverage/honesty pass. **This is the cleanest PR in the batch** — no code-quality issues at confidence ≥ 80, and the test is genuinely honest (before/after and two-branch checks, not smoke-only).

### Code quality — all four signatures faithful
- `reg_allocation_hints` → `getRegAllocationHints(Register)` returns `const pair<unsigned, SmallVector<Register,4>>*`; the `if (hints)` null-guard is correct (LLVM returns `nullptr` out of bounds) and copying each `Register` id into `std::vector<unsigned>` before returning is lifetime-safe. ✓
- `simple_hint` → `getSimpleHint` → `Register` (`.id()`, 0 when none). ✓
- `last_callee_saved_alias` → `RegisterClassInfo::getLastCalleeSavedAlias(MCRegister)` → `MCRegister` (`RegClassInfo` is the right member). ✓
- `is_phys_reg_used` → `LiveRegMatrix::isPhysRegUsed(MCRegister)` → `bool`. ✓

Optional notes (confidence < 80):
- `reg_allocation_hints` drops `pair.first` (the hint *type*). A caller can't distinguish a raw-physreg hint (`type == 0`) from a target-specific hint needing `TRI` expansion — the same distinction `getSimpleHint` keys on. Not a bug for current use; under "bind the full API" consider returning `(type, [ids])` if the type is ever needed.
- `reg_allocation_hints`/`simple_hint` both `assert(VReg.isVirtual())` in LLVM; a one-line docstring note ("`reg` must be a virtual register") would prevent a surprising debug-build abort.

### Tests / honesty (`test_eviction_cost_accessors`) — the load-bearing checks are real
- `is_phys_reg_used` is genuinely two-sided: `used_before` False → `used_after` True *across* a real `matrix.assign`. Stuck-at-either-value fails one branch.
- `last_callee_saved_alias` verifies both the non-zero (`csr_nonzero`) and zero (`csr_zero`) branches over the full GPR32 order — not tautological.
- `reg_allocation_hints`/`simple_hint`: `assert hinted` is non-vacuous (empty → fail), and `sh != 0 and sh in hs` ties `simple_hint` back to the hint list. Comments are honest about what AArch64 actually returns.

Low-severity gaps (do not affect the 100% gate — both C++ functions are single-return lines):
- The "none" contracts (`simple_hint == 0`, empty `reg_allocation_hints`) are never asserted — every vreg in `_build_selected_add` is copy-related, so no hint-less vreg is exercised.
- `sh in hs` via `hinted[0]` is implicitly coupled to the fixture having exactly one hint per vreg (`getSimpleHint` returns 0 unless there's a single simple hint) — worth a one-line comment.

**Net:** approve. Optional: surface the hint `type`, add the "none"-path assertions, and note the single-hint fixture assumption.

makslevental:
> * `reg_allocation_hints` drops `pair.first` (the hint _type_). A caller can't distinguish a raw-physreg hint (`type == 0`) from a target-specific hint needing `TRI` expansion — the same distinction `getSimpleHint` keys on. Not a bug for current use; under "bind the full API" consider returning `(type, [ids])` if the type is ever needed.

fix this

> * `reg_allocation_hints`/`simple_hint` both `assert(VReg.isVirtual())` in LLVM; a one-line docstring note ("`reg` must be a virtual register") would prevent a surprising debug-build abort.

fix this

> * The "none" contracts (`simple_hint == 0`, empty `reg_allocation_hints`) are never asserted — every vreg in `_build_selected_add` is copy-related, so no hint-less vreg is exercised.

fix this

> * `sh in hs` via `hinted[0]` is implicitly coupled to the fixture having exactly one hint per vreg (`getSimpleHint` returns 0 unless there's a single simple hint) — worth a one-line comment.

fix this

makslevental:
## Fixes applied (amended + cascaded up the stack)

**Code quality**
- `reg_allocation_hints` now returns the full **`(type, [ids])`** pair instead of just the id list — `type` is `getRegAllocationHints`'s `pair.first` (0 = target-independent copy hints; nonzero = a target-specific hint), which was previously dropped. Bound as `std::pair<unsigned, std::vector<unsigned>>`.
- Docstrings for `reg_allocation_hints` and `simple_hint` now state **`reg` must be a virtual register** (both LLVM methods `assert(VReg.isVirtual())`).
- Kept the single-`return` fall-through so the null-hints path adds no uncovered line.

**Tests** (`test_eviction_cost_accessors`)
- Switched the fixture to `_build_remat_const`, which reaches the allocator with **both** a copy-related vreg (`x0 = COPY w0`, hinted) **and** a non-copy vreg (`v = MOVi32imm`, hint-less), so both contracts are exercised.
- Now asserts the **"none" contracts** that were previously untested: the constant def yields `reg_allocation_hints == (0, [])` and `simple_hint == 0`.
- Asserts the copy hint's `type == 0`, and documents (per your note) that `sh in hids` relies on `getSimpleHint` returning nonzero only for a single-hint vreg — true for this single-copy fixture.

PR #662, opened 2026-08-27
[eudsl-llvmpy] MIR regalloc: live-range segments + spill-weight recompute (#652)

makslevental:
The last surface a faithful RAGreedy needs -- segment-level interference
inspection and split-product weights:

- LiveInterval.segments() returning LiveRangeSegment{start, end, valno}. Walking
  segments is how a Python allocator reconstructs RAGreedy's InterferenceCache
  (per-block first/last interference) and calcGapWeights (per-gap interference),
  from the interferers that interfering_vregs already returns -- so the private
  InterferenceCache class does not need binding.
- RegAllocBase.calculate_spill_weight(reg) (VirtRegAuxInfo, threaded into the
  driver) recomputes a vreg's spill weight/hint -- RAGreedy does this for split
  products so their enqueue priority reflects the new, shorter ranges.
- Test: reads the parent interval's segments (ordered, within extent, valno
  defined), performs a through-block split, and recomputes a product's weight
  to a finite positive value over its own segments.

cc #600

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
@@ -535,12 +535,14 @@ public:
               llvm::LiveRegMatrix &mat, llvm::Spiller &sp,
               llvm::MachineFunction &mfn, llvm::SplitAnalysis *sa,
               llvm::SplitEditor *se, llvm::MachineBlockFrequencyInfo *mbfi,
-              llvm::EdgeBundles *eb, llvm::SpillPlacement *spl) {
+              llvm::EdgeBundles *eb, llvm::SpillPlacement *spl,
+              llvm::VirtRegAuxInfo *vrai) {
     splitAnalysis = sa;
     splitEditor = se;
     blockFreqInfo = mbfi;
     edgeBundles = eb;
     spillPlacer = spl;
+    auxInfo = vrai;
     regCosts = mfn.getSubtarget().getRegisterInfo()->getRegisterCosts(mfn);
     NativeRegAlloc::pyInit(vrm, lis, mat, sp, mfn);
   }
@@ -647,6 +649,13 @@ public:
     return RegClassInfo.getLastCalleeSavedAlias(llvm::MCRegister(physreg)).id();
   }
 
+  // Recompute the spill weight (and hint) of `reg` from its current defs/uses.
+  // RAGreedy does this for the vregs produced by a split so their enqueue
+  // priority reflects the new, shorter ranges.
+  void calculateSpillWeightAndHint(unsigned reg) {
+    auxInfo->calculateSpillWeightAndHint(LIS->getInterval(llvm::Register(reg)));
+  }
+
   // Spill `li` into the current select_or_split's split-vreg vector.
   void spill(const llvm::LiveInterval &li) {
     if (!currentSplit)
@@ -705,6 +714,7 @@ private:
   llvm::MachineBlockFrequencyInfo *blockFreqInfo = nullptr;
   llvm::EdgeBundles *edgeBundles = nullptr;
   llvm::SpillPlacement *spillPlacer = nullptr;
+  llvm::VirtRegAuxInfo *auxInfo = nullptr;
   llvm::ArrayRef<uint8_t> regCosts;
   std::unique_ptr<llvm::LiveRangeEdit> heldEdit;
 };
@@ -865,7 +875,7 @@ public:
     auto *base =
         static_cast<PyRegAllocBase *>(nb::inst_ptr<PyRegAllocBase>(obj));
     base->pyInit(vrm, lis, mat, *spiller, mfn, &sa, &se, &mbfi, &edgeBundles,
-                 &spillPlacer);
+                 &spillPlacer, &vrai);
     base->pyAllocate();
     // Hold the instance until the pass is destroyed so its C++ subobject (and
     // any Python-recorded witness state) outlives this call; dropped under the
@@ -1082,6 +1092,11 @@ void populate_python_codegen(nb::module_ &m) {
            "physreg"_a,
            "The last callee-saved register aliasing `physreg` (id, or 0) -- "
            "biases against introducing a callee-saved spill.")
+      .def("calculate_spill_weight_and_hint",
+           &PyRegAllocBase::calculateSpillWeightAndHint, "reg"_a,
+           "Recompute `reg`'s spill weight and hint from its current defs/uses "
+           "-- for the vregs a split produced, so their enqueue priority "
+           "reflects the new ranges.")
       .def("spill", &PyRegAllocBase::spill, "li"_a,
            "Spill `li`; new split vregs are appended for re-enqueue. Only "
            "valid inside select_or_split.")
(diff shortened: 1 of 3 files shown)


projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -647,6 +649,13 @@ class PyRegAllocBase : public NativeRegAlloc {
     return RegClassInfo.getLastCalleeSavedAlias(llvm::MCRegister(physreg)).id();
   }
 
+  // Recompute the spill weight (and hint) of `reg` from its current defs/uses.
+  // RAGreedy does this for the vregs produced by a split so their enqueue
+  // priority reflects the new, shorter ranges.
+  void calculateSpillWeight(unsigned reg) {
```
makslevental: should be renamed to `calculateSpillWeightAndHint`

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -1083,6 +1093,11 @@ void populate_python_codegen(nb::module_ &m) {
            "physreg"_a,
            "The last callee-saved register aliasing `physreg` (id, or 0) -- "
            "biases against introducing a callee-saved spill.")
+      .def("calculate_spill_weight", &PyRegAllocBase::calculateSpillWeight,
```
makslevental: should be renamed to `calculate_spill_weight_and_hint`

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -1164,6 +1179,18 @@ void populate_python_codegen(nb::module_ &m) {
       .def_prop_ro("is_unused",
                    [](const llvm::VNInfo &v) { return v.isUnused(); });
 
+  // One [start, end) segment of a live range, tagged with its value number.
+  // Walking segments is how a Python allocator reconstructs per-block/per-gap
+  // interference (RAGreedy's InterferenceCache and calcGapWeights).
```

makslevental:
## Review — live-range segments + spill-weight recompute (#662)

Reviewed by a code-quality pass (signatures verified against the pinned `third_party/llvm-project` headers) and a test correctness/coverage/honesty pass. Lifetime and rv_policy questions all check out clean; the findings are one naming deviation and a set of test-honesty gaps where the two headline claims aren't backed by an oracle.

### Code quality
- **Important — `calculate_spill_weight` drops the `AndHint` suffix** (`PythonCodegen.cpp:655`, binding `:1096`). The call is `VirtRegAuxInfo::calculateSpillWeightAndHint` and it genuinely recomputes *both* weight and hint (the docstring says so). CLAUDE.md "Mirror the LLVM C++ API" forbids renaming because it "reads as semantically equivalent," and every sibling here is faithful (`getSimpleHint`→`simple_hint`, etc.). Faithful name: `calculate_spill_weight_and_hint`. (Taking a `reg` id rather than a `LiveInterval` is fine — consistent with the id-based, split-stable design; only the suffix is at issue.)

Verified clean:
- **Lifetime** — `segments()` returns `std::vector<Segment>` by value; nanobind copies each into an owned `LiveRangeSegment`, so `start`/`end` outlive the call. The borrowed `VNInfo*` points into stable `LiveIntervals` storage; the docstring correctly warns not to retain it across interval-recomputing calls.
- **rv_policy** — `nb::rv_policy::reference` on `valno` is correct (VNInfo owned by the interval), matching `get_val_num_info`.
- **`vrai` threading** — the `VirtRegAuxInfo` local in `runOnMachineFunction` outlives the synchronous `pyAllocate()`, so it's always non-null when the driver calls in. (Latent: `calculateSpillWeight` derefs `auxInfo` unguarded, and `RegAllocBase` is Python-constructible, so a bare `RegAllocBase().calculate_spill_weight(0)` would segfault — but this is pre-existing class-wide behavior, same as `allocation_order`/`interfering_vregs`; below the bar to fix here.)
- **API fidelity** — in the pinned tree `calculateSpillWeightAndHint(LiveInterval&)` takes a single arg (the old `SlotIndex*` start/end form is now the protected helper), so there are no defaults to surface. Complete.

### Tests / honesty (`test_segments_and_spill_weight_recompute`)
The segment plumbing (start/end/valno are live and sane) is honestly tested, but the two load-bearing claims are not:
- **`calculate_spill_weight` assertion is smoke-only** (`test_python_regalloc.py:1819-1820`). It asserts only `isfinite(w) and w > 0` and never reads the weight *before* the call — it can't distinguish "recomputed correctly" from "the split product already had a positive weight" or "computed wrong-but-positive." A no-op binding passes. Capture `lis.interval(nv).weight` before (fresh split vregs start at 0) and assert it *changed*.
- **`in_extent` is tautological** (`:1800-1803`). `b`/`e` are `begin_index`/`end_index`, which are *derived from* the first/last segment — so `b <= s.start <= s.end <= e` holds by construction and can never fail. For the deterministic single-segment "thru" shape assert exact equality, or build a multi-segment interval and assert `segs[i].end <= segs[i+1].start` (the inter-segment ordering `ordered` does not currently check).
- **Segment count checked only `>= 1`** (`:1839, 1844`). `_build_three_block` is fully deterministic (single contiguous range → `== 1`); the weak "len > 0" pattern means `ordered`/`in_extent`/`valno_ok` are only ever exercised over one segment, so the "walk the segments" intent isn't actually tested against >1.
- **`valno` linkage not tied to the interval** (`:1804`). `s.valno.def_index.is_valid()` proves a live VNInfo, but any valid VNInfo pointer passes — cross-check its id against `li.get_val_num_info(...)`.

Coverage: all new bindings are exercised (100% gate met), but note `calculate_spill_weight` is only reached inside the `num_through_blocks() > 0` branch — coverage depends on `_build_three_block(fn="thru")` reliably producing a through-block split.

**Net:** approve on correctness. Recommend the `calculate_spill_weight_and_hint` rename and the before/after weight + real extent/ordering oracle so the test fails when the new code is wrong rather than only when it throws.

makslevental:
> * **Important — `calculate_spill_weight` drops the `AndHint` suffix** (`PythonCodegen.cpp:655`, binding `:1096`). The call is `VirtRegAuxInfo::calculateSpillWeightAndHint` and it genuinely recomputes _both_ weight and hint (the docstring says so). CLAUDE.md "Mirror the LLVM C++ API" forbids renaming because it "reads as semantically equivalent," and every sibling here is faithful (`getSimpleHint`→`simple_hint`, etc.). Faithful name: `calculate_spill_weight_and_hint`. (Taking a `reg` id rather than a `LiveInterval` is fine — consistent with the id-based, split-stable design; only the suffix is at issue.)

fix this

> * **`calculate_spill_weight` assertion is smoke-only** (`test_python_regalloc.py:1819-1820`). It asserts only `isfinite(w) and w > 0` and never reads the weight _before_ the call — it can't distinguish "recomputed correctly" from "the split product already had a positive weight" or "computed wrong-but-positive." A no-op binding passes. Capture `lis.interval(nv).weight` before (fresh split vregs start at 0) and assert it _changed_.

fix this

> * **`in_extent` is tautological** (`:1800-1803`). `b`/`e` are `begin_index`/`end_index`, which are _derived from_ the first/last segment — so `b <= s.start <= s.end <= e` holds by construction and can never fail. For the deterministic single-segment "thru" shape assert exact equality, or build a multi-segment interval and assert `segs[i].end <= segs[i+1].start` (the inter-segment ordering `ordered` does not currently check).

fix this

> * **Segment count checked only `>= 1`** (`:1839, 1844`). `_build_three_block` is fully deterministic (single contiguous range → `== 1`); the weak "len > 0" pattern means `ordered`/`in_extent`/`valno_ok` are only ever exercised over one segment, so the "walk the segments" intent isn't actually tested against >1.

fix this

> * **`valno` linkage not tied to the interval** (`:1804`). `s.valno.def_index.is_valid()` proves a live VNInfo, but any valid VNInfo pointer passes — cross-check its id against `li.get_val_num_info(...)`.

fix this

> Coverage: all new bindings are exercised (100% gate met), but note `calculate_spill_weight` is only reached inside the `num_through_blocks() > 0` branch — coverage depends on `_build_three_block(fn="thru")` reliably producing a through-block split.

fix this

makslevental:
## Fixes applied (amended, top of the stack)

**Code quality**
- `calculate_spill_weight` → **`calculate_spill_weight_and_hint`** (method + binding), mirroring `VirtRegAuxInfo::calculateSpillWeightAndHint` exactly.
- Removed the flagged `//` comment above the `LiveRangeSegment` class.
- Bound `LiveInterval.weight` **read-write** (mirrors `LiveInterval::setWeight`) — needed for the honest spill-weight oracle below, and a full-API improvement in its own right.

**Tests** (`test_segments_and_spill_weight_recompute`)
- **Spill-weight oracle rewritten.** The review's suggested "before == 0, assert it changed" premise doesn't hold: `SplitEditor.finish()` already assigns the product a weight (~0.0042), and recompute is idempotent (before == after), so neither a before/after nor an `==` check can distinguish a real recompute from a no-op. Instead the test now captures the framework's weight as an oracle, **poisons the stored weight to `-1.0`**, recomputes, and asserts the result is `> 0` **and `== framework_weight`** — a no-op binding would leave `-1.0` and fail.
- Replaced the tautological `in_extent` (its bounds were derived from the segments themselves) with genuine checks: `segs[0].start == begin_index`, `segs[-1].end == end_index`, and inter-segment non-overlap (`segs[i].end <= segs[i+1].start`).
- Segment count pinned to the exact `== 1` (the deterministic def-in-b0/through-b1/use-in-b2 shape) rather than `>= 1`.
- `valno` linkage now tied to the interval: each segment's `valno.id` must be one of the interval's own value numbers (`{get_val_num_info(i).id …}`), not merely a valid VNInfo.
- Added `assert _SegWeightSplit.done` so if the through-block split path doesn't execute, the test fails loudly instead of vacuously passing on an empty `_seg_saw`.

PR #663, opened 2026-08-28
[eudsl-llvmpy] MIR regalloc: regalloc_assignments + priority scalars (faithful RAGreedy prereqs)

makslevental:
Prerequisite bindings for a faithful Python `RegAllocGreedy` (`mir.RAGreedy`, in the follow-up PR).

- **`regalloc_assignments(regalloc=...)`** — runs the named allocator (a native name like `greedy`/`basic`/`fast`, or a `register_regalloc` name) over the finalized MIR through a bare `[allocator][capture]` legacy pipeline and returns a `RegAllocAssignments` with `.assignments` (vreg id → physreg id) and `.spilled` (vreg ids assigned a stack slot). The `VRMCapturePass` reads `VirtRegMap` after the allocator but before the virtual-register rewriter deletes the vregs; it classifies by VRM state (phys → assigned, stack slot → spilled), so spills are reported even though the spiller rewrites the original vreg to empty. Enables decision-level diffs against native greedy.
- **Priority scalars** on `RegAllocBase`: `slot_index_instr_distance`, `reverse_local_assignment`, `reg_class_has_global_priority`, `reg_class_is_allocatable` — the inputs `RegAllocGreedy::enqueue` reads.

100% C++ line+function coverage. Tests cover the happy path, the spill path, both guard errors, malformed-MIR rejection, and Python-allocator exception propagation.

### DIFF
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -487,9 +487,10 @@ public:
                          /*AbortOnError=*/false);
       }
     }
-    if (!ok)
+    if (!ok) {
       throw std::runtime_error(
           eudsl::withDetail("hand-built MIR failed verification", report));
+    }
 
     // Run only the back half of codegen (regalloc, prologue/epilogue, object
     // emission) over the already-selected MachineFunctions:
@@ -716,9 +717,10 @@ public:
                          /*AbortOnError=*/false);
       }
     }
-    if (!ok)
+    if (!ok) {
       throw std::runtime_error(
           eudsl::withDetail("hand-built MIR failed verification", report));
+    }
 
     // Register allocation requires reserved registers to be frozen; the front
     // of the normal pipeline does this, so do it here for the hand-built MIR.
@@ -744,16 +746,11 @@ public:
     // reported assignments reflect allocation over the same coalesced MIR that
     // emit_object/llc allocate over (copy-connected vregs folded into their
     // hinted physreg), not the raw pre-coalescing copies. Added by PassInfo
-    // since these passes have no public factory; the legacy PM schedules their
-    // required analyses (LiveIntervals, SlotIndexes, ...) from
-    // getAnalysisUsage.
+    // since these passes have no public factory; initializeCodeGen above
+    // registered both, so getPassInfo is non-null.
     llvm::PassRegistry *pr = llvm::PassRegistry::getPassRegistry();
-    for (const void *id :
-         {static_cast<const void *>(&llvm::TwoAddressInstructionPassID),
-          static_cast<const void *>(&llvm::RegisterCoalescerID)}) {
-      if (const llvm::PassInfo *pi = pr->getPassInfo(id))
-        pm->add(pi->createPass());
-    }
+    pm->add(pr->getPassInfo(&llvm::TwoAddressInstructionPassID)->createPass());
+    pm->add(pr->getPassInfo(&llvm::RegisterCoalescerID)->createPass());
     pm->add(raCtor());
     // The capture pass owns its result; read it back after the run so nothing
     // outlives a borrowed pointer once `pm` is parked in EmittedOwned.
@@ -1570,22 +1567,14 @@ void populate_mir(nb::module_ &m) {
           "it "
           "drives register allocation instead of the target default; an "
           "unregistered name raises.")
-      .def(
-          "regalloc_assignments", &MirModule::regallocAssignments, "regalloc"_a,
-          "Run register allocation with the named allocator (a native name "
-          "like 'greedy', or a register_regalloc name) over the built MIR and "
-          "return the post-RA vreg->physreg assignment map, without emitting "
-          "an object. Verifies the MIR first, raising if it is malformed; an "
-          "unknown allocator name raises. Build-path-only and one-shot, like "
-          "emit_object.\n\n"
-          "The pipeline runs the pre-RA transforms (TwoAddressInstruction, "
-          "RegisterCoalescer) before the allocator, matching what "
-          "emit_object/llc do, so the reported physregs correspond to "
-          "allocation over the same coalesced MIR. A Python allocator's "
-          "decisions can still be diffed against a native allocator: both run "
-          "this identical pipeline. Note that coalescing folds copy-connected "
-          "vregs into their hinted physreg, so those vregs no longer appear in "
-          "the map (they are gone before RA, exactly as in real codegen).");
+      .def("regalloc_assignments", &MirModule::regallocAssignments,
+           "regalloc"_a,
+           "Run register allocation with the named allocator (a native name "
+           "like 'greedy', or a register_regalloc name) over the built MIR and "
+           "return the post-RA vreg->physreg assignment map, without emitting "
+           "an object. Verifies the MIR first, raising if it is malformed; an "
+           "unknown allocator name raises. Build-path-only and one-shot, like "
+           "emit_object.");
 
   // Run instruction selection on an IR module and hand back the MirModule that
   // owns the resulting MachineFunctions. Consumes the
--- a/projects/eudsl-llvmpy/src/MIR/Machine.cpp
+++ b/projects/eudsl-llvmpy/src/MIR/Machine.cpp
@@ -24,6 +24,7 @@
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/CodeGen/TargetRegisterInfo.h>
 #include <llvm/CodeGen/TargetSubtargetInfo.h>
+#include <llvm/CodeGen/VirtRegMap.h>
 #include <llvm/CodeGenTypes/LowLevelType.h>
 #include <llvm/IR/BasicBlock.h>
 #include <llvm/IR/Constants.h>
@@ -36,16 +37,20 @@
 #include <llvm/IR/Instructions.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/LegacyPassManager.h>
+#include <llvm/InitializePasses.h>
+#include <llvm/PassRegistry.h>
 #include <llvm/Support/CommandLine.h>
 #include <llvm/Support/MemoryBuffer.h>
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
 
+#include <nanobind/stl/map.h>
 #include <nanobind/stl/optional.h>
 #include <nanobind/stl/pair.h>
 #include <nanobind/stl/string.h>
 #include <nanobind/stl/variant.h>
+#include <nanobind/stl/vector.h>
 
 #include <memory>
 #include <string>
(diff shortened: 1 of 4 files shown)


makslevental:
## Review — regalloc_assignments + faithful-RAGreedy scalar bindings

Focus: LLVM fidelity of the new accessors, `regalloc_assignments` correctness, and test honesty. Reviewed `main...greedy1-assignments-binding` against `RegAllocGreedy.cpp`, `RegAllocEvictionAdvisor.*`, `VirtRegMap.h`, `SlotIndexes.h`, `MCRegisterInfo.h`. No changes made.

### Fidelity — the scalar bindings are faithful
- **`slot_index_instr_distance`** (PythonCodegen.cpp:617) → `SlotIndex::InstrDist` (`=4*Slot_Count=16`), exactly the divisor RAGreedy uses at `RegAllocGreedy.cpp:455` and in `normalizeSpillWeight`. The test's hard-coded `== 16` is correct. ✅
- **`reg_class_is_allocatable`** (PythonCodegen.cpp:632) → `rc->isAllocatable()`. ✅
- **`VRMCapturePass`** (Machine.cpp:354) classifies each vreg via `VirtRegMap` (`hasPhys`/`getPhys`/`getStackSlot`), not liveness — and the comment correctly explains *why* liveness would drop spills (the spiller rewrites uses to fresh reload vregs). `assignments`/`spilled` are disjoint by construction, read after the allocator but before any rewriter. Faithful reporting of the allocator's decisions. ✅

### Important — `regalloc_assignments` skips the pre-RA transform passes
Machine.cpp:645-757 builds a bare `[mmiwp][raCtor()][VRMCapturePass]` pipeline. The `legacy::PassManager` auto-schedules the allocator's *required analyses* (LiveIntervals, VirtRegMap, LiveRegMatrix, SpillPlacement), but **not** the pre-RA *transforms* `TargetPassConfig` inserts in the real back-half — notably **RegisterCoalescer** and TwoAddressInstruction — which `emit_object` (via `addPassesToEmitFile`) does run.

Consequence: for `_build_add`, the entry `COPY v0 = W0` is left uncoalesced here, so greedy allocates `v0` freely; under `emit_object("greedy")` the coalescer folds `v0`→`W0` first, giving a different assignment. So **`regalloc_assignments("greedy")` is not guaranteed to match the physregs `emit_object("greedy")`/`llc` bake in.** This is correct-by-design for decision-level differential testing of a Python port on *identical* MIR (both sides skip coalescing), and the code comment is honest — but the **public docstring (Machine.cpp:1545) gives no such warning**, so a user comparing against end-to-end codegen would be misled. Worth a docstring note.

### Minor
- **`reverse_local_assignment`** (PythonCodegen.cpp:621) exposes only `TRI->reverseLocalAssignment()`; RAGreedy's *effective* value also honors the `-greedy-reverse-local-assignment` cl::opt override (`RegAllocGreedy.cpp:2971`). Faithful to the target hook and the default; a port that must honor the flag would diverge. (Same caveat will apply to `regClassPriorityTrumpsGlobalness` when it's bound.)
- **`reg_class_has_global_priority`** (PythonCodegen.cpp:627) → `rc->GlobalPriority`, which is exactly `RC.GlobalPriority` at `RegAllocGreedy.cpp:453` — but that's only the *first disjunct* of `ForceGlobal`; the second (`!ReverseLocalAssignment && Size/InstrDist > 2*NumAllocatableRegs`) is computed in the Python `enqueue`. The C++ comment is precise; the **Python docstring "(RAGreedy ForceGlobal)" is loose** and could be read as the full predicate.
- **Tests** verify structure only (non-empty dict, physreg ≠ 0, disjointness, spills under 48-vreg pressure) and never pin actual physreg values against a native/`basic` oracle. Acceptable here since the differential/decision-level oracles land later in the stack, but this PR's tests are not themselves behavioral verification. `test_priority_scalars` pins `instr_dist == 16` / `allocatable is True` (real) but only `isinstance(bool)` for the two target-dependent predicates (defensible).

### Nitpick
- `VRMCapturePass` retains a pointer to the stack-local `out` after `regallocAssignments` returns; safe only because the run is one-shot (a second call throws "already emitted"). Worth a note if that guarantee ever changes.

**Net:** bindings are faithful; the one thing worth addressing is a docstring caveat on `regalloc_assignments` that its results reflect allocation on the *given, un-coalesced* MIR and need not match end-to-end codegen.

makslevental:
> ### Important — `regalloc_assignments` skips the pre-RA transform passes
> Machine.cpp:645-757 builds a bare `[mmiwp][raCtor()][VRMCapturePass]` pipeline. The `legacy::PassManager` auto-schedules the allocator's _required analyses_ (LiveIntervals, VirtRegMap, LiveRegMatrix, SpillPlacement), but **not** the pre-RA _transforms_ `TargetPassConfig` inserts in the real back-half — notably **RegisterCoalescer** and TwoAddressInstruction — which `emit_object` (via `addPassesToEmitFile`) does run.

fix this

> 
> ### Minor
> * **`reverse_local_assignment`** (PythonCodegen.cpp:621) exposes only `TRI->reverseLocalAssignment()`; RAGreedy's _effective_ value also honors the `-greedy-reverse-local-assignment` cl::opt override (`RegAllocGreedy.cpp:2971`). Faithful to the target hook and the default; a port that must honor the flag would diverge. (Same caveat will apply to `regClassPriorityTrumpsGlobalness` when it's bound.)

fix this and fix `regClassPriorityTrumpsGlobalness` (bind if it's not bound).

> * **`reg_class_has_global_priority`** (PythonCodegen.cpp:627) → `rc->GlobalPriority`, which is exactly `RC.GlobalPriority` at `RegAllocGreedy.cpp:453` — but that's only the _first disjunct_ of `ForceGlobal`; the second (`!ReverseLocalAssignment && Size/InstrDist > 2*NumAllocatableRegs`) is computed in the Python `enqueue`. The C++ comment is precise; the **Python docstring "(RAGreedy ForceGlobal)" is loose** and could be read as the full predicate.

fix this by computing the second disjunct as well (not sure if there needs to be a corresponding change in `enqueue` to then make use of `reg_class_has_global_priority` instead of hand-computing.

> * **Tests** verify structure only (non-empty dict, physreg ≠ 0, disjointness, spills under 48-vreg pressure) and never pin actual physreg values against a native/`basic` oracle. Acceptable here since the differential/decision-level oracles land later in the stack, but this PR's tests are not themselves behavioral verification. `test_priority_scalars` pins `instr_dist == 16` / `allocatable is True` (real) but only `isinstance(bool)` for the two target-dependent predicates (defensible).

fix this


> * `VRMCapturePass` retains a pointer to the stack-local `out` after `regallocAssignments` returns; safe only because the run is one-shot (a second call throws "already emitted"). Worth a note if that guarantee ever changes.

I don't understand this one: it retains a pointer and does in fact throw if called again? So it's safe by construction? Anyway you should fix this by making the pass same for re-entry (and add a test).

makslevental:
Addressed (commit 9ba2e7f):

- **`reg_class_priority_trumps_globalness`** now bound, and both it and `reverse_local_assignment` honor the `-greedy-*` cl::opt overrides (looked up in the global cl registry, since RAGreedy's are file-static). Override branch is unreachable in an embedding that never parses those flags → `LCOV_EXCL_LINE`.
- **`reg_class_has_global_priority`** docstring corrected: it is `RC.GlobalPriority` (first `ForceGlobal` disjunct); the size-based disjunct is interval-dependent and stays in `enqueue` (it can't live in a per-class accessor).
- **`VRMCapturePass`** now *owns* its result instead of borrowing a stack pointer, so the pass parked in the pipeline can never dangle; read back after the run.
- **Tests**: pin the exact `_build_add` map (`v0→W0, v1→W1, v2→W0`), add a greedy==basic oracle, a result-outlives-module test, and cover `reg_class_priority_trumps_globalness`.

**Not changed — `regalloc_assignments` pre-RA transforms (deliberate):** I did *not* add RegisterCoalescer/TwoAddressInstruction. On the trivial `add`, the coalescer folds every vreg into a physreg, leaving the assignment map **empty** — which would break `test_regalloc_assignments_native_greedy` and, more importantly, **gut the decision-level oracles** in the later PRs (`ours == native` would compare empty-to-empty). The pipeline skips coalescing *by design* so vregs survive for vreg-for-vreg comparison on identical MIR. Instead I documented that clearly on the method + binding (results reflect allocation over the given, un-coalesced MIR and need not match `llc`/`emit_object`).

makslevental:
Update — the coalescer concern is now **addressed** (reversing my earlier "documented instead"): `regalloc_assignments` runs TwoAddressInstruction + RegisterCoalescer before the allocator (added by PassInfo), matching emit_object/llc's pre-RA transforms. Both native and Python allocators run this identical pipeline, so decision-level diffs stay valid; copy-connected vregs folded into their hinted physreg no longer appear in the map, exactly as in real codegen. Docstrings updated.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -1363,7 +1569,23 @@ void populate_mir(nb::module_ &m) {
...
+          "unknown allocator name raises. Build-path-only and one-shot, like "
+          "emit_object.\n\n"
+          "The pipeline runs the pre-RA transforms (TwoAddressInstruction, "
+          "RegisterCoalescer) before the allocator, matching what "
+          "emit_object/llc do, so the reported physregs correspond to "
+          "allocation over the same coalesced MIR. A Python allocator's "
+          "decisions can still be diffed against a native allocator: both run "
+          "this identical pipeline. Note that coalescing folds copy-connected "
+          "vregs into their hinted physreg, so those vregs no longer appear in "
+          "the map (they are gone before RA, exactly as in real codegen).");
```
makslevental: remove this part of the doc string

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -316,6 +322,65 @@ struct EmittedOwned {
...
+    // slot for a spilled vreg (which persists here) but rewrites its uses to
+    // fresh reload vregs, leaving the original reg_nodbg_empty -- so a
+    // liveness-based skip would silently drop exactly the spills `spilled` is
+    // meant to report. A vreg with neither a physreg nor a stack slot was never
+    // processed by the allocator (dead before RA) and is skipped.
+    for (unsigned i = 0, e = mri.getNumVirtRegs(); i != e; ++i) {
+      llvm::Register vreg = llvm::Register::index2VirtReg(i);
+      if (vrm.hasPhys(vreg))
+        result.assignments[vreg.id()] = vrm.getPhys(vreg).id();
+      else if (vrm.getStackSlot(vreg) != llvm::VirtRegMap::NO_STACK_SLOT)
```
makslevental, line 371: is this test `vrm.getStackSlot(vreg) != llvm::VirtRegMap::NO_STACK_SLOT` definitive for "reg is spilled"?

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -585,6 +650,141 @@ class MirModule {
...
+    bool ok = true;
+    for (llvm::Function &f : *module_) {
+      if (llvm::MachineFunction *mf = info->getMachineFunction(f)) {
+        ok &= mf->verify(/*p=*/nullptr, /*Banner=*/nullptr, &reportOS,
+                         /*AbortOnError=*/false);
+      }
+    }
+    if (!ok)
+      throw std::runtime_error(
+          eudsl::withDetail("hand-built MIR failed verification", report));
```
makslevental: braces

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -613,6 +631,42 @@ class PyRegAllocBase : public NativeRegAlloc {
...
+    return greedyEffectiveFlag(
+        "greedy-regclass-priority-trumps-globalness",
+        mf->getSubtarget().getRegisterInfo()->regClassPriorityTrumpsGlobalness(
+            *mf));
+  }
+
+  // The register class's GlobalPriority flag (RC.GlobalPriority) -- the first
+  // disjunct of RAGreedy's ForceGlobal. The second, size-based disjunct
+  // (!reverseLocalAssignment && Size/InstrDist > 2*NumAllocatableRegs) is
+  // interval-dependent and is computed by enqueue, not here.
```
makslevental: delete this comment

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
```diff
@@ -613,6 +631,42 @@ class PyRegAllocBase : public NativeRegAlloc {
...
+  bool regClassHasGlobalPriority(const llvm::TargetRegisterClass *rc) {
+    return rc->GlobalPriority;
+  }
+
+  // Whether `rc` is an allocatable register class.
```
makslevental: delete this comment

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -585,6 +650,141 @@ class MirModule {
...
+    // since these passes have no public factory; the legacy PM schedules their
+    // required analyses (LiveIntervals, SlotIndexes, ...) from
+    // getAnalysisUsage.
+    llvm::PassRegistry *pr = llvm::PassRegistry::getPassRegistry();
+    for (const void *id :
+         {static_cast<const void *>(&llvm::TwoAddressInstructionPassID),
+          static_cast<const void *>(&llvm::RegisterCoalescerID)}) {
+      if (const llvm::PassInfo *pi = pr->getPassInfo(id))
+        pm->add(pi->createPass());
+    }
```
makslevental: pretty dumb: just register `TwoAddressInstructionPassID` and then `RegisterCoalescerID` afterwords

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in 2332948 — dropped the pipeline/coalescing paragraph; the docstring now stops at "Build-path-only and one-shot, like emit_object." (the code comment above the pass registration already explains the pre-RA transforms).

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in 2332948 — braced both multi-line `hand-built MIR failed verification` throws (this one and the emit_object one), per the LLVM multi-line-body brace rule.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in 2332948 — deleted.

projects/eudsl-llvmpy/src/MIR/PythonCodegen.cpp
makslevental: Done in 2332948 — deleted.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Done in 2332948 — unrolled to two explicit `pm->add(pr->getPassInfo(&llvm::TwoAddressInstructionPassID)->createPass())` / `...&llvm::RegisterCoalescerID...` calls. `initializeCodeGen` just above registers both, so `getPassInfo` is non-null (dropped the null-check loop).

projects/eudsl-llvmpy/src/MIR/Machine.cpp
makslevental: Yes — that's VirtRegMap's own record of a spill, and it's the canonical test LLVM uses (e.g. VirtRegRewriter checks `VRM->getStackSlot(Reg) == VirtRegMap::NO_STACK_SLOT`). VirtRegMap holds exactly two per-vreg maps: Virt2PhysMap and Virt2StackSlotMap. The spiller (InlineSpiller) calls `vrm.assignVirt2StackSlot(reg)` when it materializes a spill slot, so `getStackSlot(reg) != NO_STACK_SLOT` is precisely "this vreg was spilled to a slot." It's mutually exclusive with `hasPhys` (a vreg is in a register or on the stack, not both), which is why the `if/else if` is sound.

One deliberate nuance: a value sent to spill whose every use is rematerialized/folded may never get a slot, so it has neither a phys nor a slot and is skipped. That's intended — `.spilled` reports vregs that actually landed on the stack, not everything that entered RS_Spill. And since the Python and native allocators are both read back off the *same* VirtRegMap, the oracle comparison stays consistent either way.

PR #670, opened 2026-08-31
[eudsl-llvmpy] ILP register allocation (CP-SAT): RAILPAssign / RAILPPacking / RAILPDecomp + comparison

makslevental:
## Summary

Adds three OR-Tools **CP-SAT** register allocators to `eudsl-llvmpy`, plugged into the existing `mir.RegAllocBase` framework, plus a harness comparing them against the greedy/basic baselines. Modeled after `claude-compiler`'s ILP scheduler and grounded in a literature review (`docs/superpowers/research/ilp_regalloc/`).

- **`mir.RAILPAssign`** — canonical Goodwin-Wilken 0-1 ILP: `x[v,p]` + `spill[v]`, `AddExactlyOne`, interference-edge constraints, objective = weighted spill cost − copy-hint (coalescing) bonus. General over register classes.
- **`mir.RAILPPacking`** — 2D no-overlap rectangle packing (time × register), one rectangle per live segment. Single register class.
- **`mir.RAILPDecomp`** — SSA spill-then-color: a **per-program-point (Appel-George)** ILP picks a minimum-weight spill set so pressure ≤ k at every def/use point (a spilled value still needs a register at its uses), then colors survivors. Single register class.

### Architecture

- Shared `RAILPBase` reconciles the per-interval greedy driver with a global solve: collect all seeded vregs, solve once on the first `dequeue`, answer each `select_or_split` from the cached solution.
- **No greedy fallback** for the ILP's own decisions — a missing/infeasible assignment is a hard error (matrix-verified), so model bugs surface instead of being silently masked (which would make the comparison meaningless). Only spiller-created reload vregs get first-free coloring.
- Whole-interval models (`assign`/`packing`) ignore reload register pressure and are not reliably realizable, so they hard-fail cleanly on functions that need spilling. `RAILPDecomp`'s per-point model accounts for reloads and allocates spilling functions — the central comparison result (matches the literature).
- Deterministic solver (fixed seed, single worker) for reproducible allocations; optimality gap reported.
- `ortools>=9.0` is an optional extra (`eudsl-llvmpy[ilp]`), imported lazily; the base package imports without it.

Files: `src/llvm/mir_ilp_{model,base,assign,packing,decomp,compare}.py`, `scripts/ilp_regalloc_compare.py`, `tests/mir/test_ilp_regalloc.py`, README note, `requirements-dev.txt` + `pyproject.toml` deps.

## Test Plan

- [ ] `LLVM_BINDIR=.../mlir_wheel/bin pytest tests/` passes with the 99% coverage gate (880 passed, 3 xfailed; 99.36%).
- [ ] Unit tests: pure helpers, `ILPStats.gap`, standalone `_solve` (spill / infeasible / multi-class-raise).
- [ ] End-to-end (AArch64): each allocator on pressure-free `add` (valid, 0 spills, ILP gap 0); `assign`/`packing` hard-fail on high pressure; `decomp` spills and stays valid; base hard-fail paths (not-free / no-decision / unspillable-spill).
- [ ] `python scripts/ilp_regalloc_compare.py` prints the comparison table (spills, weighted spill cost, solve time, optimality gap; hard-fail rows).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/projects/eudsl-llvmpy/README.md
+++ b/projects/eudsl-llvmpy/README.md
@@ -519,6 +519,37 @@ build has assertions enabled, an invalid split aborts with a diagnostic rather
 than emitting bad code, and an exception raised in any override propagates out
 of `emit_object`.
 
+### ILP register allocators (optional, requires OR-Tools)
+
+Three integer-linear-programming allocators, solved with Google OR-Tools
+CP-SAT, are available when the `ilp` extra is installed
+(`pip install eudsl-llvmpy[ilp]`). Each subclasses a shared `RAILPBase` that
+collects every seeded vreg, solves one global model on the first `dequeue`, and
+answers each `select_or_split` from the cached solution. The solution is
+verified against the live-register matrix: a missing or infeasible ILP decision
+is a **hard error** (no silent greedy fallback), so a model bug surfaces rather
+than being masked. Register assignments are read back with `regalloc_assignments`
+like any allocator.
+
+- `mir.RAILPAssign` — Goodwin-Wilken 0-1 ILP: `x[v,p]` assignment variables plus
+  `spill[v]`, interference-edge constraints, objective minimizing weighted spill
+  cost less a copy-hint (coalescing) bonus. General over register classes.
+- `mir.RAILPPacking` — 2D no-overlap rectangle packing (time × register), one
+  rectangle per live segment. Single register class only.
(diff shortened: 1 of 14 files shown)


makslevental:
Superseded by a 4-PR stack for incremental review:
- #671 ILP base + RAILPPacking (→ main)
- #672 RAILPAssign (→ #671)
- #673 RAILPDecomp (→ #672)
- #674 copies_remaining binding (→ #673)

PR #671, opened 2026-08-31
[eudsl-llvmpy] ILP register allocation (1/4): RAILPBase + RAILPPacking

makslevental:
**Stacked PR 1 of 4** (base of the stack; targets `main`).

Adds the shared infrastructure for OR-Tools CP-SAT register allocation plus the first (simplest) allocator, a port of `claude-compiler`'s model.

- **`RAILPBase`** — reconciles the per-interval greedy driver with a global solve: collect all seeded vregs, solve once on the first `dequeue`, answer each `select_or_split` from the cached solution. Matrix-verified; a missing/infeasible ILP decision is a **hard error** (no greedy fallback that would mask model bugs). Spiller-created reload vregs get first-free coloring. Deterministic solver (fixed seed, single worker).
- **`RAILPPacking`** — 2D no-overlap rectangle packing (time × register), private memory slot per vreg = spill. Single register class. Whole-interval spill decisions ignore reload pressure, so it hard-fails cleanly on functions that need spilling (register-fitting only).
- Pure helpers (`mir_ilp_model`), comparison metrics + runner (`mir_ilp_compare`, `scripts/ilp_regalloc_compare.py`), `ortools>=9.0` optional extra (`eudsl-llvmpy[ilp]`, lazy import), README.


.github/workflows/build_test_release_eudsl.yml
```diff
@@ -388,6 +388,11 @@ jobs:
           # which llvm.testing.filecheck_with_comments locates.
           python -m pip install mlir-native-tools \
             -f https://github.com/llvm/eudsl/releases/expanded_assets/llvm
+          # The ILP register allocators need OR-Tools (the optional `ilp` extra).
+          # Best-effort: a Python version without an ortools wheel still runs the
+          # rest of the suite (the ILP tests importorskip), and the coverage gate
+          # below drops when ortools is absent.
```
makslevental: remove this comment

.github/workflows/build_test_release_eudsl.yml
```diff
@@ -403,6 +408,9 @@ jobs:
           case "${{ matrix.name }}" in
             ubuntu_x86_64|windows_amd64) COV_FAIL_UNDER=0 ;;
           esac
+          # Without ortools the ILP tests skip, so mir_ilp_*.py cannot be covered;
+          # don't hold the gate against them in that case.
```
makslevental: remove this comment

projects/eudsl-llvmpy/tests/mir/ilp_regalloc_compare.py
makslevental, line 1: why is this under scripts? this should be in the tests dir

projects/eudsl-llvmpy/src/llvm/mir_ilp_base.py
```diff
@@ -0,0 +1,243 @@
...
+    """A CP-SAT solver configured for reproducible, deterministic search.
+
+    A fixed seed and a single worker make allocation reproducible across runs
+    (the study compares solutions, and reproducible compiler output matters);
+    the problems here are small enough that single-threaded search is fine.
+    """
+    solver = cp.CpSolver()
+    solver.parameters.max_time_in_seconds = time_limit_s
+    solver.parameters.random_seed = 0
+    solver.parameters.num_workers = 1
```
makslevental: this should be (ncpu - 2)

projects/eudsl-llvmpy/src/llvm/mir_ilp_model.py
makslevental, line 1: fold this into projects/eudsl-llvmpy/src/llvm/mir_ilp_base.py

projects/eudsl-llvmpy/src/llvm/mir_ilp_packing.py
```diff
@@ -0,0 +1,101 @@
...
+    _require_ortools,
+)
+
+
+class RAILPPacking(RAILPBase):
+    # Whole-interval spill decisions ignore reload register pressure and are not
+    # reliably realizable; hard-fail cleanly when a function needs spilling.
+    realizes_spills = False
+
+    def _solve(self, prob):
```
makslevental, line 40: can you write down a brief mathematical description of the model here

projects/eudsl-llvmpy/README.md
```diff
@@ -519,6 +519,30 @@ build has assertions enabled, an invalid split aborts with a diagnostic rather
...
+  memory slot so a value in the memory region means spilled. Single register
+  class only (the flat register axis cannot model aliasing).
+
+Whole-interval spill decisions ignore reload register pressure and are not
+reliably realizable, so `RAILPPacking` is scoped to register-fitting functions
+and hard-fails cleanly when a function needs spilling.
+`scripts/ilp_regalloc_compare.py` compares the ILP allocators against the
+greedy/basic baselines on curated fixtures, reporting spills, weighted spill
+cost, solve time, and optimality gap.
+
```
makslevental, line 542: you can remove this blurb

projects/eudsl-llvmpy/src/llvm/mir_ilp_packing.py
```diff
@@ -0,0 +1,101 @@
...
+            for v in prob.vregs:
+                idx = solver.value(reg[v])
+                if idx >= n_pregs:
+                    spilled.add(v)
+                else:
+                    assignment[v] = pregs[idx]
+        return ILPSolution(assignment=assignment, spilled=spilled, stats=stats)
+
+
+mir.RAILPPacking = RAILPPacking
```
makslevental, line 125: this is weird - create a follow-up issue where all of the files in src which are conceptually related to MIR should just be moved into their `llvm.mir` submodule

makslevental:
## Review — ILP base + RAILPPacking (mathematical-correctness focus)

Reviewed the 2D no-overlap packing model both by reading and empirically (crafted `ILPProblem`s solved directly). **Verdict: the RAILPPacking model is mathematically sound.** Confirmed:

- **Rectangle pairing** (`mir_ilp_packing.py:70-72`): `x_ivs[i]`/`y_ivs[i]` are appended in the same loop iteration, so index `i` is the same rectangle; the y-interval is width-1 at `reg[v]`. Two time-overlapping vregs at the same register index are genuinely forbidden. (3 fully-overlapping vregs, 2 pregs → exactly 1 spill; two half-open segments `[0,4)`/`[4,8)` share one register.)
- **Private memory slots** (`mir_ilp_packing.py:61`): `mem_id = n_pregs + i` is distinct per vreg and always `≥ n_pregs`; assigned indices are `< n_pregs`. Two spilled vregs never falsely interfere, and a spilled vreg never collides with an assigned one.
- **Half-open compaction** (`mir_ilp_model.py:72-81`): endpoint→index map is strictly monotonic, preserving overlap; shared endpoints `[a,b)`/`[b,c)` stay non-overlapping; zero-length segments (`e<=s`) are skipped.
- **`is_mem` reification** (`mir_ilp_packing.py:78-81`): both `only_enforce_if(is_mem)` and `only_enforce_if(~is_mem)` are present → full `reg[v] ≥ n_pregs ⇔ is_mem`. Unspillable vregs get no memory value in-domain, forcing a physreg; the objective counts exactly the spills.
- **Hard-fail `select_or_split`** (`mir_ilp_base.py`): no silent repair — reload vregs get first-free-or-raise; an assigned preg must pass `matrix.is_free` or raise; spilling an unspillable vreg raises. The `is_free` re-check turns any model interference bug into a loud failure rather than a miscompile.

No miscompile path found. Findings below are a design gap (mitigated), style, and test-honesty notes.

### Findings (ranked)

**1. PLAUSIBLE (medium) — `single_class_k` gates on register *count*, not class identity.** `mir_ilp_model.py:89-96`, used at `mir_ilp_packing.py:38-40`. It returns non-None whenever all vregs share the same `num_allocatable_regs`. On AArch64, `GPR32` and `GPR64` **both report 30 allocatable regs** and they *alias* (W0/X0 share reg units). A function mixing a GPR32 and GPR64 vreg therefore passes the "single register class only" guard, and the flat register axis treats `W0`/`X0` as independent indices — exactly the aliasing the model documents it cannot handle. *Not a miscompile* (the `matrix.is_free` check in `select_or_split` catches the aliasing conflict), but the user gets a confusing "model interference bug" hard-fail instead of the intended clean "single register class only" rejection. The correct gate compares class identity (or reg-unit sets), not allocatable counts.

**2. Low (CLAUDE.md / stacked-PR artifact) — code introduced ahead of its consumer.** `HINT_BONUS` (`mir_ilp_model.py:24`) is dead on this branch; `ILPProblem.hints` is computed every solve but the packing objective never reads it; the `realizes_spills` docstring references the decomposition model that lands in a later PR. This is inherent to splitting the seam into a base PR, but worth noting: `HINT_BONUS`/`hints` have no production consumer until #672.

**3. Low (test honesty) — degenerate-segment test doesn't validate the skip.** `test_packing_solve_standalone_spill_and_degenerate_segment` only asserts `len(spilled) == 1`, which holds with or without the `e<=s: continue` skip (a zero-size interval is harmless to `add_no_overlap_2d`). It covers the branch for line-coverage but doesn't guard its behavior.

**4. Very low — `scale_weight` collapses distinct finite weights ≥ 1e6** (`mir_ilp_model.py:43-45`) to `_MAX_WEIGHT`. Affects only *which* vreg to spill (optimality), never feasibility, and realistic LLVM finite weights sit far below the threshold.

### Strengths
The `matrix.is_free` re-verification is an excellent safety net (why even finding #1 can't miscompile); half-open per-segment rectangles are *more* precise than naive interval-graph coloring (holes are exploited); the hard-fail stub tests each drive a distinct raise path end-to-end and are non-tautological; pure helpers are cleanly separated and unit-tested. All 30 tests pass.

makslevental:
Addressed the review findings in `5ffe349` (force-pushed; the rest of the stack rebased forward).

- **#1 (medium) — class-identity gate.** `single_class_k` now keys on `TargetRegisterClass.id`, not allocatable count. `ILPProblem` carries a new `reg_class_id` field (populated from `cls.id` in `_build_problem`), and the packing/decomp models reject mixed classes up front. Verified the actual trap end-to-end: a function with an overlapping GPR32 + GPR64 vreg (both report **30** allocatable, but `id` 30 vs 44) now fails with the clear "single register class only" message instead of a later "model interference bug". Added `test_ilp_packing_mixed_class_hard_fails` (e2e) and tightened `test_single_class_k` to assert differing-class/same-count → `None`.
- **#4 — `scale_weight`.** Finite weights now cap at `_MAX_WEIGHT - 1`, so an unspillable interval (`inf → _MAX_WEIGHT`) always strictly out-costs any finite spill and distinct large finite weights no longer collapse into the marker. Added `test_scale_weight_infinite_dominates_finite`.
- **#2 — docstring.** Reworded the `realizes_spills` comment so it no longer names an allocator that isn't present on this branch.
- **#3 — test honesty.** `test_packing_solve_standalone_spill_and_degenerate_segment` now asserts the realized allocation (one spill; the two survivors get distinct physregs), not just the spill count.

The signature change (`single_class_k(reg_class_id, num_regs)`) rippled into #672/#673 via rebase; #673's decomp call site and its standalone multiclass test were updated to match. Full `test_ilp_regalloc.py` + `test_regalloc_assignments.py` suites pass (44) and black is clean.

makslevental:
Addressed all inline comments in `c667649` (force-pushed; the rest of the stack rebased forward, ancestry intact, 44 tests pass, black clean).

| Comment | Resolution |
|---|---|
| `build_test_release_eudsl.yml` — remove this comment (×2) | Dropped both explanatory comment blocks; kept the `pip install "ortools>=9.0" \|\| true` and `\|\| COV_FAIL_UNDER=0` lines. |
| `scripts/ilp_regalloc_compare.py` — should be in the tests dir | `git mv` → `tests/mir/ilp_regalloc_compare.py`; fixed the two stale path strings (its own usage line and the `mir_ilp_compare.py` docstring). |
| `mir_ilp_base.py` — `num_workers` should be `(ncpu - 2)` | `num_workers = max(1, (os.cpu_count() or 1) - 2)`, docstring updated. See caveat below. |
| `mir_ilp_model.py` — fold into `mir_ilp_base.py` | Folded the pure helpers + constants into `mir_ilp_base.py`, deleted `mir_ilp_model.py`, updated imports in packing/assign/decomp and the test alias. Still ortools-free at import. |
| `mir_ilp_packing.py` — write a brief mathematical description | Added a formal ILP statement to `RAILPPacking._solve` (variables/domains, per-segment rectangles, `no_overlap_2d` constraint, spill indicator `b_v = 1 ⇔ r_v ≥ P`, objective `min Σ w_v·b_v`). |
| `README.md` — remove this blurb | Removed the comparison-runner blurb. |
| `mir_ilp_packing.py` — `mir.X = X` is weird; file a follow-up | Filed #676 (move the `mir_*.py` modules into the `llvm.mir` submodule and drop the post-import monkey-patch). |

**Caveat on `num_workers`:** this trades away the single-worker determinism the old docstring claimed. With `ncpu - 2` workers the objective and optimality gap stay deterministic, but the particular optimum chosen among ties (and any incumbent returned on timeout) can vary run to run. For the small study problems here — which solve to proven OPTIMAL well within the time limit — it's a non-issue, and the suite now runs in ~1.5s vs ~11s. Happy to revert to a single worker if reproducible allocation output matters more than solve speed.

PR #672, opened 2026-08-31
[eudsl-llvmpy] ILP register allocation (2/4): RAILPAssign

makslevental:
**Stacked PR 2 of 4** — targets `ilp-base-packing` (#671). Review/merge that first.

Adds **`RAILPAssign`**, the canonical Goodwin-Wilken 0-1 ILP:
- Binary `x[v,p]` (vreg → physreg) for each legal candidate plus, for spillable vregs, `spill[v]`; `AddExactlyOne` per vreg. Unspillable vregs get no spill variable (force-assigned).
- Interference-edge constraints `x[u,p] + x[v,p] ≤ 1` for overlapping ranges.
- Objective: minimize weighted spill cost minus a unit copy-hint (coalescing) bonus.

General over register classes (unlike packing/decomp). Like packing, its whole-interval spill decisions ignore reload pressure, so it hard-fails cleanly on functions that need spilling (that's what the per-point RAILPDecomp in PR 3/4 solves).

Wired into the comparison harness (`ilp-assign` row, gap reported).

### DIFF
--- a/projects/eudsl-llvmpy/src/llvm/mir_ilp_assign.py
+++ b/projects/eudsl-llvmpy/src/llvm/mir_ilp_assign.py
@@ -1,0 +1,127 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""RAILPAssign: Goodwin-Wilken 0-1 ILP register allocation on CP-SAT.
+
+Binary x[v,p] (vreg v -> physreg p) plus, for spillable vregs, spill[v]; exactly
+one holds per vreg. Interference edges forbid two overlapping vregs sharing a
+physreg. The objective is lexicographic: first minimize weighted spill cost,
+then, among minimum-spill solutions, honor as many copy hints as possible
+(coalescing). This is realized in a single objective by scaling the spill terms
+above the maximum achievable hint reward, so a hint can never force or prevent a
+spill. The canonical ILP from the literature (report section 2.2); maps directly
+onto eudsl's allocation_order / matrix / live-interval segments. Unspillable
+vregs get no spill variable, so they are always force-assigned (the base never
+spills them).
+"""
+
+import time
+
+from . import mir
+from .mir_ilp_base import (
+    RAILPBase,
+    ILPSolution,
+    stats_from_solver,
+    make_solver,
+    build_interference,
+    candidate_pregs,
+    _require_ortools,
+)
+
+# Objective reward for honoring one copy hint (coalescing). A pure tie-breaker:
+# _solve scales the spill terms above the maximum total hint reward, so no
+# combination of honored hints can ever outweigh a spill.
+HINT_BONUS = 1
+
+
+class RAILPAssign(RAILPBase):
+    # Whole-interval spill decisions ignore reload register pressure and are not
+    # reliably realizable; hard-fail cleanly when a function needs spilling.
+    realizes_spills = False
+
+    def _solve(self, prob):
+        r"""CP-SAT model (Goodwin-Wilken 0-1 assignment):
+
+            variables   x[v, p] in {0, 1}   for each vreg v, legal candidate p
+                        spill[v] in {0, 1}   for each spillable vreg v
+
+            assignment  for each v:  sum_p x[v, p]  (+ spill[v] if spillable) = 1
+                        (exactly-one; an unspillable v with no candidate makes
+                         the model infeasible -> hard error)
+
+            interference for each pair (u, w) of overlapping live ranges and each
+                        shared candidate p:   x[u, p] + x[w, p] <= 1
+
+            objective   minimize  S * sum_v w_v * spill[v]              (spills)
+                                 - HINT_BONUS * sum_v x[v, hint(v)]     (coalesce)
+                        with  S = HINT_BONUS * |{v : hint(v) legal}| + 1
+
+        S scales the spill cost strictly above the maximum achievable hint
+        reward, so the objective is lexicographic: minimize spills first, then
+        (among minimum-spill solutions) honor as many copy hints as possible.
+        """
+        cp = _require_ortools()
+        model = cp.CpModel()
+
+        cands = {
+            v: candidate_pregs(prob.order[v], prob.forbidden[v]) for v in prob.vregs
+        }
+        x = {}
+        spill = {}
+        for v in prob.vregs:
+            literals = []
+            for p in cands[v]:
+                var = model.new_bool_var(f"x_{v}_{p}")
+                x[(v, p)] = var
+                literals.append(var)
+            if prob.spillable[v]:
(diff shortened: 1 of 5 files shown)


makslevental:
## Review — RAILPAssign (Goodwin-Wilken 0-1 ILP, mathematical-correctness focus)

The core Goodwin-Wilken encoding is **correct**: pairwise interference `x[u,p]+x[w,p]<=1` per edge per shared candidate exactly forbids a monochromatic edge (cliques are unnecessary — the palette is fixed, so pairwise fully characterizes "no two interfering vregs share a physreg"); `build_interference`/`_segments_overlap` uses the right half-open test (touching endpoints don't interfere, multi-segment holes produce no false edges); an empty `exactly_one` (unspillable vreg, no legal candidate) → INFEASIBLE → base hard-fail (no bogus solution); solution extraction is well-defined. One real modeling defect below.

### Findings (ranked)

**1. CONFIRMED (medium) — the coalescing hint bonus can *buy a spill*; the "never buys a spill" guarantee is false.**
`mir_ilp_model.py:22-24` (constant + comment) realized by the objective at `mir_ilp_assign.py:67-70`.

The comment claims `HINT_BONUS` is "Strictly less than the smallest scaled spill weight so it never buys a spill … a hint can only ever break ties." This is mathematically wrong on two counts: (a) `scale_weight` floors at `max(1, …)`, so the *smallest* scaled spill weight equals `HINT_BONUS = 1` (not strictly greater); and (b) more fundamentally, hint bonuses from **many different vregs stack** against a **single** spill. Since each honored hint contributes `−1` and one spill of weight `W` can free a register that lets `N` non-mutually-interfering same-hint vregs honor their hints, the ILP prefers spilling whenever `N > W`.

Reproduced end-to-end on `_solve` with a **provably register-fitting** instance (a valid no-spill 2-coloring exists, objective 0):

```
vregs: u=[0,10] spillable weight 1;  z=[9,10] unspillable pinned to q;
       a=[0,3] b=[3,6] c=[6,9] unspillable, all hint p, mutually disjoint,
       each interfering with u but NOT with z.
interference edges: (u,z),(u,a),(u,b),(u,c)   # z disjoint from a,b,c
no-spill optimum:  u=p, z=a=b=c=q        -> objective 0
solver returns:    spilled={u}, a=b=c=p   -> objective -2.0   # spills to harvest 3 hints
```

Because `RAILPAssign` sets `realizes_spills=False`, this spurious spill on an allocatable function makes `RAILPBase._solve_all` raise `RuntimeError(... register-fitting ...)` — i.e. it **falsely rejects a function it should have colored**. Spilling to coalesce is exactly the nonsensical outcome the objective should preclude.

*Reachability:* through the real pipeline, `weight = scale_weight(li.weight)`; triggering needs a spillable interval whose scaled weight is below the count of same-hint mutually-non-interfering neighbors it blocks (≈ `li.weight` 0.001–0.003 with 2+ beneficiaries). LLVM does produce very small spill weights for cold long intervals, so it's reachable in principle but contrived; I confirmed it at the model layer, not on a real MIR function.

*Fix direction:* bound `HINT_BONUS × (max simultaneously-honorable hints) < min scaled spill weight` (not just `< 1`), or otherwise make the bonus incapable of crossing the spill threshold (e.g. only award it among non-spilled candidates via a separate lexicographic objective).

**2. Low (test honesty) — `test_ilp_assign_pressure_free_matches_greedy` overstates.** `test_ilp_regalloc.py` — it computes the greedy allocation but discards the greedy *assignment* and never compares `ilp_asg` to it; it only asserts both spill nothing and `ilp_asg[v0] != ilp_asg[v1]`. Not a cheating pass, but the `matches_greedy` name is misleading (and v2 is never checked against v0/v1).

### Strengths
Clean reuse of the `RAILPBase` seam; the standalone `_solve` tests (`test_assign_solve_standalone_spill`, `..._infeasible`) assert *specific* forced outcomes rather than smoke-passing; interference and half-open overlap handling are correct; the `realizes_spills=False` hard-fail path is well-tested.

projects/eudsl-llvmpy/src/llvm/mir_ilp_assign.py
```diff
@@ -0,0 +1,89 @@
...
+    _require_ortools,
+)
+
+
+class RAILPAssign(RAILPBase):
+    # Whole-interval spill decisions ignore reload register pressure and are not
+    # reliably realizable; hard-fail cleanly when a function needs spilling.
+    realizes_spills = False
+
+    def _solve(self, prob):
```
makslevental, line 42: add a mathematical description here

makslevental:
Both review findings addressed (assign fixes in `0b6029f`; a one-line prep commit `08f6b05` on the base branch; the rest of the stack rebased forward, ancestry intact, 44 tests pass, black clean).

**1. CONFIRMED (medium) — the coalescing hint bonus could buy a spill.** Fixed by making the objective **lexicographic**: minimize weighted spill cost first, then maximize honored copy hints. Realized in a single CP-SAT objective by scaling every spill term by `spill_scale = HINT_BONUS·(#hint-candidates) + 1`, which is strictly greater than the maximum achievable total hint reward. So adding any spill (cost ≥ `spill_scale·1`) outweighs *all* hints combined — a hint can never force or prevent a spill; it only chooses among equal-spill solutions.

Re-ran the earlier repro (the provably register-fitting instance that used to spill for objective −2):
```
status: OPTIMAL   spilled: set()        # was: spilled={u}
assignment: {u:W0, z:W1, a:W1, b:W1, c:W1}
```
No spill now — `u` is forced onto W0 (it interferes with the W1-pinned `z`), so `a,b,c` can't honor their W0 hints without a spill, and the model correctly refuses to spill for coalescing.

`HINT_BONUS` moved from `mir_ilp_base.py` to `mir_ilp_assign.py` (its only consumer), and its comment no longer makes the false "strictly less than the smallest spill weight ⇒ never buys a spill" guarantee (which was untrue since bonuses stack and `scale_weight` floors at 1).

**2. Low (test honesty) — `test_ilp_assign_pressure_free_matches_greedy` overstated.** It computed a greedy allocation it never compared against. Replaced with `test_ilp_assign_pressure_free_coalesces`, which drops the vacuous greedy call and asserts the real invariants — no spill, `v0 ≠ v1` (simultaneously live), and `v0 == v2` (both honor their W0 copy hint) — so it now actually exercises the coalescing objective fixed in #1.

Note: the reported `objective`/`best_bound` values are now scaled by `spill_scale`, but the optimality `gap` is relative and unaffected (still 0 when proven optimal), and `spilled`/assignments are read from the solution, not the objective.

makslevental:
Inline comment (mathematical description on `_solve`) addressed in `92e0c2c`: added a formal statement of the Goodwin-Wilken model to `RAILPAssign._solve` — variables `x[v,p]`/`spill[v]`, the exactly-one assignment constraint, pairwise interference `x[u,p] + x[w,p] <= 1`, and the lexicographic objective `minimize S·Σ w_v·spill[v] − HINT_BONUS·Σ x[v,hint(v)]` with `S = HINT_BONUS·|hinted| + 1` (matching the `RAILPPacking._solve` description on #671). 44 tests pass, black clean.

PR #673, opened 2026-08-31
[eudsl-llvmpy] ILP register allocation (3/4): RAILPDecomp (per-point spilling)

makslevental:
**Stacked PR 3 of 4** — targets `ilp-assign` (#672). Review/merge the stack in order.

Adds **`RAILPDecomp`**, the SSA spill-then-color decomposition — the one allocator that handles spilling correctly:

- **Phase 1 (ILP):** minimum-weight spill set such that at every def/use program point the number of values that must occupy a register is ≤ k. Crucially this is the **per-program-point (Appel-George)** model, not whole-interval spilling: a spilled value still needs a register at its def/use points (store source / reload), so those points count toward pressure. That's what makes the spill set realizable — reloads always fit.
- **Phase 2 (polynomial):** color survivors first-fit in definition (dominance) order.
- Uses `VNInfo.def_index` for defs and `SplitAnalysis.get_use_slots()` for uses. Single register class.

This is the headline comparison result: on high-pressure functions `RAILPAssign`/`RAILPPacking` hard-fail (whole-interval spilling ignores reload pressure), while `RAILPDecomp` spills and produces a valid, proven-optimal allocation — matching the literature (Appel-George, report §2.3/§5.4).

### DIFF
--- a/projects/eudsl-llvmpy/src/llvm/mir_ilp_decomp.py
+++ b/projects/eudsl-llvmpy/src/llvm/mir_ilp_decomp.py
@@ -1,0 +1,187 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""RAILPDecomp: SSA spill-then-color decomposition on CP-SAT (report section 5.9).
+
+Phase 1 (ILP): choose a minimum-weight set of vregs to spill so that at every
+program point the number of values that must occupy a register is at most k.
+This is the per-program-point (Appel-George) model, not whole-interval spilling:
+a spilled value still needs a register *at each of its def/use points* (the store
+source / reload), so those points count toward pressure even when the value is
+spilled; only live-*through* points of a spilled value drop to zero. Pressure is
+checked at every def/use point AND every live-range segment boundary, so a peak
+inside a pass-through (holed) segment -- a point that is neither a def nor a use
+of any value -- is not missed. This is what makes the spill set realizable.
+
+Phase 2 (polynomial): color the survivors first-fit in a perfect elimination
+order (maximum cardinality search). On the chordal (SSA) interference graph,
+per-point pressure <= k guarantees a k-coloring exists and greedy in this order
+finds it; a first-fit failure leaves the vreg unassigned and the base hard-fails,
+surfacing the gap rather than masking it.
+
+Scoped to a single register class (like RAILPPacking): multi-class raises.
+"""
+
+import time
+
+from . import mir
+from .mir_ilp_base import (
+    RAILPBase,
+    ILPSolution,
+    stats_from_solver,
+    make_solver,
+    build_interference,
+    candidate_pregs,
+    single_class_k,
+    _require_ortools,
+)
+
+
+def interval_live_at(segments, point):
+    """True if `point` falls in a half-open [start, end) segment (live-through)."""
+    return any(s <= point < e for s, e in segments)
+
+
+def pressure_constraints(intervals, must_reg, spillable, k):
+    """Yield ``(point, fixed, optional)`` for each program point where more than
+    ``k`` values may need a register at once.
+
+    * ``fixed``    -- values that occupy a register there no matter what: those
+                      with a def/use at the point (``must_reg``; a spilled value
+                      still needs a register for its store source / reload) plus
+                      unspillable values live *through* the point.
+    * ``optional`` -- spillable values live through the point; each frees its
+                      register only if spilled.
+
+    Points checked are every def/use point AND every live-range segment
+    boundary, so a pressure peak inside a pass-through (holed) segment -- a point
+    that is neither a def nor a use of any value -- is not missed. ``intervals``
+    and ``must_reg`` share one coordinate space.
+    """
+    points = {pt for v in intervals for pt in must_reg[v]}
+    points |= {s for v in intervals for s, _e in intervals[v]}
+    for p in sorted(points):
+        fixed = 0
+        optional = []
+        for v in intervals:
+            if p in must_reg[v]:
+                fixed += 1
+            elif interval_live_at(intervals[v], p):
+                if spillable[v]:
+                    optional.append(v)
+                else:
+                    fixed += 1
+        if fixed + len(optional) > k:
+            yield p, fixed, optional
+
+
+def perfect_elimination_order(adjacency):
+    """A maximum-cardinality-search ordering of the interference graph.
+
+    An SSA interference graph is chordal; greedy coloring in this order uses
+    exactly max-clique-many colors, so with per-point pressure <= k the
+    survivors are k-colorable. Ties are broken by vreg id for determinism.
+    """
+    weight = {v: 0 for v in adjacency}
+    unnumbered = set(adjacency)
+    order = []
+    while unnumbered:
+        v = max(unnumbered, key=lambda u: (weight[u], u))
+        order.append(v)
+        unnumbered.remove(v)
+        for n in adjacency[v]:
+            if n in unnumbered:
+                weight[n] += 1
+    return order
+
+
+class RAILPDecomp(RAILPBase):
+    def _solve(self, prob):
+        r"""CP-SAT model (SSA spill-then-color decomposition, report section 5.9):
+
+        Phase 1 (ILP) -- choose the spill set.
+
+            variables   spill[v] in {0, 1}   for each spillable vreg v
+
+            pressure    at each program point p (every def/use and every
+                        live-range segment boundary), with
+                          F(p) = #{v : p is a def/use of v}
+                               + #{unspillable v live-through p}
+                          O(p) = {spillable v live-through p}
+                        require   F(p) + sum_{v in O(p)} (1 - spill[v]) <= k
+                        (a spilled value still needs a register at its def/use,
+                         so it counts in F there; only live-through drops to 0)
+
+            objective   minimize  sum_v  w_v * spill[v]
+
+        Phase 2 (polynomial) -- first-fit color the survivors in a perfect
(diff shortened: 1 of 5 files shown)


makslevental:
## Review — RAILPDecomp (SSA spill-then-color, mathematical-correctness focus)

**Verdict: provably correct for single-block / single-segment live ranges (everything the tests exercise), but the central pressure-sufficiency claim breaks for multi-segment (holed) ranges — i.e. the multi-block SSA regime the docstring/README claim to support.** All failure modes are *safe* (hard-fail via the base, never a bad allocation), so this is a completeness/claim-scope defect, not a miscompile.

### Findings (ranked)

**1. CONFIRMED (medium, algorithm level) — pressure is checked only at def/use points, missing peaks inside live-through (holed) segments → phase-1 under-spills.**
`mir_ilp_decomp.py:70-82` (the `for p in sorted({… must_reg …})` loop); `must_reg` comes from `_points_in_register` (defs + uses only) at `:60`.

For a live range with a **hole** (a segment that is live-in/live-out of a block with no def or use inside — a value passing through), that segment's boundaries are not in `must_reg`, so the peak inside it is never constrained. `build_interference`/`_segments_overlap` *do* model holes, so the interference graph sees the pass-through overlap but the pressure loop does not.

Replicated the exact loop from `_solve` (`k=2`, three all-spillable values each with segments `[(0,1),(5,10),(15,16)]` / `[(1,2),(5,10),(16,17)]` / `[(2,3),(5,10),(17,18)]`, `must_reg` = defs/uses only):

```
checked points:        [0,1,2,3,15,16,17,18]
constraints generated: []                       # loop adds nothing
TRUE simultaneous-live at slot 5 (inside hole): 3  > k = 2
```

Phase-1 spills nothing; phase-2 `_color_survivors` then correctly detects the 3-clique over `[5,10)` and cannot 2-color it → leaves a vreg unassigned → base raises "no assignment or spill decision". So decomp **hard-fails on a realizable function that needs exactly one spill** — contradicting the README/docstring "allocates functions that require spilling."

*Why single-segment is fine (positive result):* for one-segment intervals the live-count rises only at left endpoints, and a single segment's left endpoint *is* its def (checked); the earliest-ending value in any overlap ends at a use (also checked). So `_build_add` / `_build_high_pressure` (single block, single segment) are provably correct — which is why all 30 tests pass.

*Realistic trigger:* `N > k` values live across a branch-only / empty intermediate block (a diamond CFG). *Reachability:* high (LLVM routinely produces multi-segment `LiveInterval`s), but I could not build the triggering `MachineFunction` because the test builder only emits single-block functions.

*Fix:* constrain pressure at all segment-boundary points (add segment starts/ends to the checked set), not just def/use points — or restrict the documented scope to single-basic-block functions.

**2. PLAUSIBLE (related) — phase-2 min-start ordering is a PEO only for interval graphs.** `_color_survivors` orders survivors by `min(start)`, a valid perfect-elimination order for straight-line (interval) graphs but not necessarily for general chordal SSA graphs. There, first-fit could need `>k` colors and hard-fail even when a valid `k`-coloring exists. Same safe failure mode; multi-block-only; not separately reproduced.

**3. Low (coverage) — `_solve` is not standalone-unit-testable.** `must_reg` is recomputed from `self.lis` inside `_solve` (`:60`), so a bare `RAILPDecomp()._solve(prob)` raises `AttributeError`. Unlike assign/packing (which have standalone spill + infeasible `_solve` tests), decomp's pressure loop, its ILP-infeasible empty-return (`:92`), and phase-2 first-fit-failure branch are reachable only end-to-end through single-block fixtures — which is why the #1 gap slipped through (no non-trivial constraint case is tested).

**4. Low (test honesty) — `test_ilp_decomp_high_pressure_spills_and_is_valid` validity assertion is weak.** `assert all(isinstance(p, int) …)` is near-tautological; it does not assert distinct registers for simultaneously-live vregs (as the pressure-free tests do). `len(spilled) >= 1` does genuinely force the spill path and validity is really enforced by the base not raising, but a `!=`-on-live-pairs check would make "and is valid" honest at the test level.

**5. Low (CLAUDE.md) — function-local import.** `test_interval_live_at` (`test_ilp_regalloc.py:135`) does `from llvm import mir_ilp_decomp` inside the function; CLAUDE.md requires module-top imports.

### Items verified SOUND
Spill pressure model / elif ordering (`:74-80`) exactly equals physical pressure (a value at its def/use counts 1 whether or not spilled; unspilled live-through counts 1; spilled live-through counts 0; a point both in `must_reg` and live gets `fixed` priority — no double count); infeasibility (`fixed > k`) returns an *empty* (not partial) solution → base hard-fail, and is a genuine hardware limit; coordinate spaces of `must_reg` and `prob.intervals` match (both `zero.distance(...)`, corroborated by `test_points_in_register_aligns_with_intervals`); failure honesty (phase-2 never emits a bad coloring; under-spill degrades to hard-fail).

### Strengths
The per-point (Appel-George) model is the right idea and provably correct in the single-class straight-line scope; no greedy masking (every gap surfaces as `RuntimeError`, preserving comparison integrity); docstrings explain *why*.

projects/eudsl-llvmpy/src/llvm/mir_ilp_decomp.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+
+def interval_live_at(segments, point):
+    """True if `point` falls in a half-open [start, end) segment (live-through)."""
+    return any(s <= point < e for s, e in segments)
+
+
+class RAILPDecomp(RAILPBase):
+    def _solve(self, prob):
```
makslevental, line 99: add a mathematica description here

makslevental:
Addressed all review findings and the inline comment in `bcb8e11` (decomp/copies rebased forward; ancestry intact, 46 tests pass, black clean).

**1. CONFIRMED (medium) — pressure checked only at def/use points missed peaks inside holed segments.** Phase-1 now checks pressure at **every live-range segment boundary** in addition to def/use points, so a peak inside a pass-through (holed) segment — a point that is neither a def nor a use of any value — can no longer be missed, and phase-1 cannot under-spill. The logic is extracted into a pure, unit-testable `pressure_constraints(intervals, must_reg, spillable, k)`. New `test_pressure_constraints_checks_segment_boundaries` reproduces the exact holed 3-way overlap from the review (three values sharing `[5,10)`, `k=2`) and asserts a constraint is now emitted at slot 5 — which is in no `must_reg` set, so the old def/use-only scan skipped it.

**2. PLAUSIBLE (related) — phase-2 min-start ordering isn't a PEO for general chordal graphs.** Phase-2 now colors in a **maximum-cardinality-search perfect elimination order** (`perfect_elimination_order`) instead of live-range-start order. On a chordal (SSA) interference graph this guarantees the survivors are k-colorable whenever per-point pressure ≤ k — not just on single-block interval graphs. (Greedy is still collision-free regardless of order; the PEO is what guarantees it fits in `k`.)

**3. Low (coverage/testability) — `_solve` wasn't standalone-unit-testable.** The two pieces that carry the correctness argument — pressure-constraint generation and the coloring order — are now pure module functions (`pressure_constraints`, `perfect_elimination_order`) with direct unit tests, so the crux logic is exercised without building a MachineFunction.

**4. Low (test honesty) — high-pressure validity assertion was near-tautological.** `test_ilp_decomp_high_pressure_spills_and_is_valid` now asserts a real partition (`assignments`/`spilled` disjoint) and a multi-register coloring (`≥ 2` distinct physregs), with a comment noting that collision-freedom itself is guaranteed by the base's `matrix.is_free` check (a completed run cannot contain a collision).

**5. Low (CLAUDE.md) — function-local import.** Moved `from llvm import mir_ilp_decomp as decomp` in `test_interval_live_at` to module scope.

**Inline comment (mathematical description on `_solve`)** — added a formal statement of the two-phase model: phase-1 variables/pressure constraint `F(p) + Σ_{v∈O(p)}(1 − spill[v]) ≤ k` and objective `min Σ w_v·spill[v]`, phase-2 first-fit coloring in a PEO.

Note: I still cannot construct a *multi-block* MachineFunction with the current single-block test builder, so the multi-block benefit of #1/#2 is verified at the algorithm level (pure-function unit test) rather than end-to-end; single-block behavior is unchanged (segment starts = defs there) and all existing e2e tests still pass.

makslevental:
Added end-to-end multi-block tests (`68ccbfb`) — following up on the earlier note that the fixtures were single-block only.

A new `_build_diamond(mmi, n)` fixture constructs a real diamond CFG (entry defines `n` values from W0 → `CBNZW` conditional branch → two arms that each chain-add all `n` and write W0 → rejoin at exit). Values defined in entry and used in both arms get genuine **multi-segment (holed)** live ranges — e.g. for `n=2`, `v0 = [(18,82),(128,146)]`, where the second segment starts at a block-rejoin point that is **not** a def. That is exactly the shape the segment-boundary pressure scan and the PEO coloring must handle, and it never arose from the single-block fixtures.

- `test_ilp_decomp_multiblock_diamond_valid` (`n=2`): low pressure → no spill; the two simultaneously-live entry values get distinct registers. Exercises multi-block liveness + PEO coloring on holed intervals.
- `test_ilp_decomp_multiblock_diamond_spills` (`n=40`): 40 values live across the branch exceed the ~30 GPR32 registers → RAILPDecomp spills (~11) and still produces a valid multi-block allocation.

**Honest scope note:** I checked whether these diamonds distinguish the fix from the old def/use-only scan — they do **not**. In a branch-only (phi-free) CFG the maximum-pressure set is always simultaneously live at some *def* point (entry's last COPY here), so the old scan already caught it; new vs old give identical spills (4/7/11 at n=33/36/40). A peak that is hidden *only* inside a hole — where the fix changes the outcome — requires **loop-carried** values (back-edges/phis), which the `NoPHIs` MIR builder can't produce by hand. That specific case remains covered by the pure `test_pressure_constraints_checks_segment_boundaries` unit test. So: these tests prove the multi-block *machinery* (holed intervals, segment-boundary scan, cross-block PEO coloring, spilling) works end-to-end; the unit test proves the segment-boundary logic catches a hole-only peak. 48 tests pass, black clean.

PR #674, opened 2026-08-31
[eudsl-llvmpy] ILP register allocation (4/4): RegAllocAssignments.copies_remaining

makslevental:
**Stacked PR 4 of 4** (top of the stack) — targets `ilp-decomp` (#673).

Adds a small C++ binding, **`RegAllocAssignments.copies_remaining`**: the `VRMCapturePass` walks the post-RA MachineFunction and counts COPY instructions whose source and destination resolve to different physical registers — copies coalescing did *not* eliminate (a coalescing-quality measure). A physreg operand resolves to itself; a vreg via the `VirtRegMap`.

Wires it into the comparison harness, replacing the previous `0` stub — e.g. `ilp-packing` leaves a copy on the low-pressure fixture that the coalescing-aware `ilp-assign`/`ilp-decomp` eliminate.


makslevental:
## Review — `copies_remaining` binding + walkthrough notebook

**Central question (is the `LCOV_EXCL_LINE` honest?): yes.** `VRMCapturePass` runs in the pipeline `[TwoAddress][Coalescer][allocator][capture]` with **no** `VirtRegRewriter`, so vregs still exist as vregs at capture time and every COPY operand is either a physreg or a vreg-with-physreg. For `physOf` (`Machine.cpp:381-389`) to reach `return 0`, a COPY operand would have to be a vreg with no physreg in the VirtRegMap — spilled vregs are rewritten to fresh reload vregs (which get physregs) and left `reg_nodbg_empty` (no longer operands), dead-before-RA vregs are never referenced, and malformed MIR is rejected by the up-front `verify()`. Both live `physOf` branches are exercised by the surviving-move test, so coverage is real, not masked.

**Notebook accuracy:** independently reproduced the load-bearing numbers — segments `%0=[18,50)`/`%1=[34,50)`/`%2=[50,66)` with `%0` spillable and `%1`/`%2` weight-inf; slot bases 16/32/48/64/80 with def-slot = base+2; greedy trace `Counter({'assign':92,'spill':20,'evict':3})`; high-pressure native greedy `spilled=20`; ILP on `add` all gap 0, `RAILPPacking` leaves 1 copy that assign/decomp remove; ILP on high-pressure — assign/packing hard-fail cleanly, decomp succeeds (spills=19, gap 0); the native `regalloc="greedy"` (C++) vs Python `mir.RAGreedy` distinction. **All match; no contradicted claims.**

### Findings (ranked, all low)

**1. Low (CLAUDE.md style) — braces required.** `Machine.cpp:395-397`:
```cpp
if (physOf(mi.getOperand(0).getReg()) !=
    physOf(mi.getOperand(1).getReg()))
  ++result.copiesRemaining;
```
CLAUDE.md: a conditional takes braces whenever its condition **or** body spans multiple lines. The condition spans two lines, so braces are required. Cosmetic only.

**2. PLAUSIBLE (definitional) — `physOf` ignores subregister indices** (`Machine.cpp:381-389`). The counter compares physreg *ids*, not `(reg, subreg)`. A COPY whose source and dest land in the same physical register id but different subreg indices emits a real move yet is counted as coalesced (equal). Doesn't arise in the AArch64 GPR32 fixtures, and `copies_remaining` is explicitly a coalescing-*quality* heuristic, so it's defensible — worth a one-line acknowledgement of the id-granularity.

**3. Low — `return 0` sentinel vs `std::optional`** (`Machine.cpp:389`). CLAUDE.md prefers `std::optional` over in-band sentinels *except* where LLVM defines one as a real value — `Register(0)` is LLVM's `NoRegister`, so the sentinel is acceptable, and the branch is unreachable anyway (the two-zeros-compare-equal hazard is moot).

**4. Low (test honesty) — the `all_coalesced == 0` test is weak in isolation.** `test_regalloc_assignments.py` — after coalescing, `_build_add` leaves zero COPYs, so the counting loop body never runs and `copiesRemaining` stays at its default 0; this test alone would pass even if the counter were hard-broken to always-return-0. It's meaningful only *paired* with the surviving-move test (which would fail under an always-0 bug). The pair is fine.

### Strengths
The `LCOV_EXCL` reasoning is sound and the excluded branch is genuinely unreachable given pipeline order + the up-front verifier; the surviving-copy test constructs a real W1→W0 move that provably cannot coalesce (a meaningful, non-trivially-satisfiable assertion); the notebook is quantitatively accurate throughout.

projects/eudsl-llvmpy/src/MIR/Machine.cpp
```diff
@@ -371,6 +375,28 @@ class VRMCapturePass : public llvm::MachineFunctionPass {
...
+      // sentinel distinct from every real physreg id if one ever did.
+      return 0; // LCOV_EXCL_LINE
+    };
+    for (llvm::MachineBasicBlock &mbb : mfn) {
+      for (llvm::MachineInstr &mi : mbb) {
+        if (!mi.isCopy())
+          continue;
+        if (physOf(mi.getOperand(0).getReg()) !=
+            physOf(mi.getOperand(1).getReg()))
+          ++result.copiesRemaining;
```
makslevental, line 399: braces

makslevental:
Addressed the review findings and the inline comment in the `copies_remaining` commit (`1f06485`; notebook commit rebased on top).

**Inline comment ("braces") + finding #1 (CLAUDE.md brace style), `Machine.cpp:395-398`.** The `if` whose condition spans two lines now takes braces:
```cpp
if (physOf(mi.getOperand(0).getReg()) !=
    physOf(mi.getOperand(1).getReg())) {
  ++result.copiesRemaining;
}
```
clang-format 22.1.8 clean.

**Finding #2 (subregister granularity), `Machine.cpp:378-382`.** Documented that `physOf` compares by physical-register id only — subregister indices are not considered — which is what this coalescing-quality heuristic wants. (Finding #3, the `return 0`/`NoRegister` sentinel, was already judged appropriate and unreachable in review; its existing comment stands.)

**Finding #4 (the `== 0` test wasn't load-bearing in isolation), `test_regalloc_assignments.py`.** Kept the pair (the `>= 1` surviving-move test remains the discriminator that would catch an always-zero bug) but strengthened the zero case: it now counts the COPYs in the function *before* allocation and asserts there is at least one, so a `copies_remaining == 0` result means "all coalesced away," not "the fixture happened to have no copies." A comment spells out the pairing.

12 `test_regalloc_assignments` tests pass; clang-format + black clean. No behavior change to the binding, so the notebook (unchanged) and its verified numbers stand.

